PR #184, opened 2014-01-21
Add a public C API to the library by mimicking the C++ ABI with a C struct

mstorsjo:
Alternative to #138, review at https://rbcommons.com/s/OpenH264/r/57/.

### DIFF
--- a/codec/api/svc/codec_api.h
+++ b/codec/api/svc/codec_api.h
@@ -36,36 +36,43 @@
 #include "codec_app_def.h"
 #include "codec_def.h"
 
+#if defined(_WIN32) || defined(__cdecl)
+#define EXTAPI __cdecl
+#else
+#define EXTAPI
+#endif
+
+#ifdef __cplusplus
 class ISVCEncoder {
  public:
   /*
    * return: CM_RETURN: 0 - success; otherwise - failed;
    */
-  virtual int Initialize (SVCEncodingParam* pParam, const INIT_TYPE kiInitType = INIT_TYPE_PARAMETER_BASED) = 0;
-  virtual int Initialize2 (void* pParam, const INIT_TYPE kiInitType = INIT_TYPE_CONFIG_BASED) = 0;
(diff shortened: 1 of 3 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/57/

----
**testing_done**:  
**description**:  Add a public C API to the libraryAdd a struct that matches the C++ interface vtable.This requires that the C++ interface methods are declared to usethe same calling convention as normal C functions, and that theC struct exactly matches the layout and ordering of the C++virtual table - MSVC seemed to reorder methods if there wereoverloaded methods.Rename public interface methods to avoid two overloaded methods with the same nameThis is required to make the order in the C++ virtual tableconsistent in MSVC - previously the overloaded methods wereordered differently in the vtable compared to the interfacedeclaration.This is a second alternative to providing a C interface to the library, with the first being in review request #33.

----
 time 2014-01-21T20:07:11+00:00 
2014-01-21T20:07:11+00:00 Ethan Hugg (ethanhugg):
>This looks like it will work and satisfies the comments in review #33.  I will merge it tomorrow unless there are any objections before then.


----

----
 reviewer JW Wang (jwwang) 
 time 2014-02-07T08:10:55+00:00 
2014-02-07T08:10:55+00:00 JW Wang (jwwang):
>I got this error: codec/api/svc/codec_app_def.h:207:3: error: unknown type name &#39;bool&#39;when including #include &quot;codec_api.h&quot; in a C file.It looks like bool type is not available in C.

2014-02-07T08:14:18+00:00 Martin Storsj (mstorsjo):
>Yes, the bool type itself isn&#39;t available in C89 in general. In C99 it is available if you include stdbool.h (and everywhere I&#39;ve tested it&#39;s been ABI compatible with the C++ bool).The places where I&#39;ve tested this I&#39;ve included stdbool.h explicitly before including codec_api.h, but I guess we should include that automatically in codec_api.h instead.


----

Finish archiving from https://rbcommons.com/s/OpenH264/r/57/

PR #227, opened 2014-01-26
Fix handling of snprintf return values and buffer size calculations

mstorsjo:
This makes the code more secure in the case the allocated string buffers aren't large enough.

### DIFF
--- a/codec/decoder/plus/src/welsDecoderExt.cpp
+++ b/codec/decoder/plus/src/welsDecoderExt.cpp
@@ -90,10 +90,12 @@ CWelsDecoder::CWelsDecoder (void_t)
   str_t chFileName[1024] = { 0 };  //for .264
   int iBufUsed = 0;
   int iBufLeft = 1023;
+  int iCurUsed;
 
   str_t chFileNameSize[1024] = { 0 }; //for .len
   int iBufUsedSize = 0;
   int iBufLeftSize = 1023;
+  int iCurUsedSize;
 #endif//OUTPUT_BIT_STREAM
 
   m_pTrace = CreateWelsTrace (Wels_Trace_Type);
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/70/

----
**testing_done**:  
**description**:  Please see https://github.com/cisco/openh264/pull/227 for the three individual changes - when they are bundled together as one diff here they might be harder to understand.Fix calculating buffer positions when appending with SNPRINTF/WelsSnprintfThis fixes two separate issues.First, with the MSVC _snprintf implementations, the return valueis negative if the buffer wasn&#39;t large enough - this would inthe worst case lead to making iBufferUsed negative, writing beforethe start of the buffer.Secondly, when both iBufferUsed and iBufferLeft are accumulated,one can&#39;t do &quot;iBufferLeft -= iBufferUsed;&quot;. As an example,say the buffer is 100 bytes in total and iBufferLeft is 40 andiBufferUsed is 60. If SNPRINTF then writes 5 more bytes to thebuffer, iBufferUsed would be 65, but if we now do&quot;iBufferLeft -= iBufferUsed;&quot; then iBufferLeft would end up as-25 even though there&#39;s 35 bytes left in the buffer to use.Therefore, we use a separate variable to store the return valuefrom the latest SNPRINTF call. This is checked to make sure itwasn&#39;t negative, and only this amount is added to iBufferUsedand subtracted from iBufferLeft.This is the same pattern used in codec/encoder/core/src/utils.cpp.strftime never returns negative numbers, so those calls don&#39;tneed as much checking.Fix buffer size checks when appending using SNPRINTF/WelsSnprintfChecking iBufferLeft &gt; iBufferUsed does not make sense, sincethis would stop writing into the buffer alredy after the bufferis half full, when there is less space left than has been used.The right check is iBufferLeft &gt; 0.Don&#39;t manually null terminate after calling SNPRINTFThe following pattern is unsafe on all platforms: n = SNPRINTF(buf, ...); buf[n] = &#39;\0&#39;;On windows, the _snprintf variants return a negative numberif the buffer was too small, thus buf[n] would be outsideof (before the start of) the buffer.On other platforms, the C99 snprintf function returns thetotal number of characters which would have been written ifthe buffer had been large enough, which can be larger thanthe buffer size itself, and thus buf[n] would be beyond theend of the buffer.The C99 snprintf function always null terminate the buffer.These invocations of SNPRINTF are within !WIN32, so we canbe sure that the SNPRINTF call itself already null terminatedthe buffer.

----
 time 2014-01-27T05:14:36+00:00 
2014-01-27T05:14:36+00:00 Volvet Zhang (Volvet):
>agree the fix and that the origin code is not good.   
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/70/

PR #258, opened 2014-01-30
Avoid endian-specific code

mstorsjo:
The few places that still used code that was endian dependent can easily be converted not to require setting a define to indicate the endianness.

This doesn't incur any significant measurable slowdown in my tests.

### DIFF
--- a/codec/common/deblocking_common.h
+++ b/codec/common/deblocking_common.h
@@ -15,12 +15,6 @@ void_t DeblockChromaLt4H_c (uint8_t* pPixCb, uint8_t* pPixCr, int32_t iStride, i
                             int8_t* pTc);
 void_t DeblockChromaEq4H_c (uint8_t* pPixCb, uint8_t* pPixCr, int32_t iStride, int32_t iAlpha, int32_t iBeta);
 
-#ifdef WORDS_BIGENDIAN
-#define DEBLOCK_BS_SHIFTED(x) ((x) | ((x) << 8))
-#else
-#define DEBLOCK_BS_SHIFTED(x) ((x) | ((x) >> 8))
-#endif
-
 #if defined(__cplusplus)
 extern "C" {
 #endif//__cplusplus
--- a/codec/decoder/core/src/deblocking.cpp
+++ b/codec/decoder/core/src/deblocking.cpp
@@ -145,34 +145,29 @@ static const uint8_t g_kuiTableBIdx[2][8] = {
 
 void_t inline DeblockingBSInsideMBAvsbase (int8_t* pNnzTab, uint8_t nBS[2][4][4], int32_t iLShiftFactor) {
   uint32_t uiNnz32b0, uiNnz32b1, uiNnz32b2, uiNnz32b3;
-  ENFORCE_STACK_ALIGN_1D (uint8_t, uiBsx3, 4, 4);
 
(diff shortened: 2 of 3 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/71/

----
**testing_done**:  
**description**:  Make the C++ code endian independentThis avoids having to set a define indicating the endianness.This doesn&#39;t incur any significant measurable slowdown.Instead of byteswapping a 32 bit word and writing it out as a whole (which could even possibly lead to crashes due to incorrect alignment on some platforms), write it out explicitly in the intended byte order.There&#39;s some amount of duplication in the deblocking code between the encoder and decoder - some of these changes are almost identical in both of them. Unfortunately, even though the functions look quite similar, there&#39;s enough small differences that make merging them not quite trivial.

----
 time 2014-02-04T20:41:00+00:00 
2014-02-04T20:41:00+00:00 Martin Storsj (mstorsjo):
>If there are no objections to these changes, can they be approved and merged?

2014-02-07T19:24:14+00:00 Martin Storsj (mstorsjo):
>Ping, any objections to merging this?


----
 time 2014-02-08T02:05:02+00:00 
2014-02-08T02:05:02+00:00 Volvet Zhang (Volvet):
 on **codec/encoder/core/inc/svc_enc_golomb.h line49**: 
 id="comment92818">1.  the macro WRITE_BE_32 is little endian case,  it is not correct at bit endian case, Though mose CPUs are little endian,  but we still big endian, right?  2.  I doubt that the WRITE_BE_32 performance is not good as origin one.  

2014-02-08T12:03:31+00:00 Volvet Zhang (Volvet):
>The performance is not changed significantly. I measured the timing, and this becomes at most around 1 cycle slower per write - this is very much insignificant in practice.Note that we don&#39;t need to do the byteswap any longer which can be more or less optimized depending on compiler.Also, note that doing a write using *(uint32_t*) can lead to failures on platforms that require strict alignment, since I don&#39;t think we can assume that the bitstream writer is 4-byte aligned here. Doing this properly requires using e.g. the ST32() macro, which tells the compiler that this is unaligned, and if you do that, on platforms that require strict alignment, the compiler will generate code similar to what I suggest changing it to, storing it one byte at a time. Even on platforms where unaligned writes are supported (such as x86), the speed benefit is much smaller when the 32 bit write isn&#39;t aligned, since it might have to do multiple reads/writes on the underlying levels in the CPU anyway.Thus, there is no significant performance benefit to doing this, and it only introduces unnecessary complication and possible portability issues, for very questionable benefit.


----
 time 2014-02-10T01:56:49+00:00 
2014-02-10T01:56:49+00:00 Volvet Zhang (Volvet):
>I also test it,  agree the performance is not changed significantly and the code is more clean.   


----
 time 2014-02-10T01:57:36+00:00 
2014-02-10T01:57:36+00:00 Volvet Zhang (Volvet):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/71/

Marktamm:
They are still on my computer putting more files on it. Is there anyone that you Know of that has enough authority to have these people arrested. Because i will press charges. Enough is enough, They are all a bunch of sneaks and thieves changing the conversation records and  reporting the wrong results.

________________________________

PR #283, opened 2014-02-12
Test more aspects of the C interface, fix using the C interface on MSVC

mstorsjo:
Under review at https://rbcommons.com/s/OpenH264/r/76/.

### DIFF
--- a/test/c_interface_test.c
+++ b/test/c_interface_test.c
@@ -1,4 +1,5 @@
 #include "codec_api.h"
+#include <stddef.h>
 
 // Cast to this function type to ignore other parameters.
 typedef int (*Func)(void*);
@@ -29,4 +30,21 @@ void CheckDecoderInterface(ISVCDecoder* p, CheckFunc check) {
   CHECK(5, p, DecodeFrameEx);
   CHECK(6, p, SetOption);
   CHECK(7, p, GetOption);
+}
+
+struct bool_test_struct {
+  char c;
+  bool b;
+};
+
+size_t GetBoolSize(void) {
+  return sizeof(bool);
+}
+
+size_t GetBoolOffset(void) {
+  return offsetof(struct bool_test_struct, b);
+}
+
+size_t GetBoolStructSize(void) {
+  return sizeof(struct bool_test_struct);
 }
--- a/test/cpp_interface_test.cpp
+++ b/test/cpp_interface_test.cpp
@@ -1,5 +1,6 @@
 #include <gtest/gtest.h>
 #include "codec_api.h"
+#include <stddef.h>
 
 static void CheckFunctionOrder(int expect, int actual, const char* name) {
   EXPECT_EQ(expect, actual) << "Wrong function order: " << name;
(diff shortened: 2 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/76/

----
**testing_done**:  
**description**:  Add a test that does basic verification that the interface ABI matches when used from CIdeally we would also want to check e.g. offsetof certain structmembers and alignof(bool) (but the latter is a C++11 feature only).Don&#39;t try to include stdbool.h on MSVCMSVC only got stdbool.h in MSVC 2013; provide a manual typedeffor this compiler instead.The C interface test makes sure that this doesn&#39;t break the ABI.

----
 time 2014-02-12T22:14:33+00:00 
2014-02-12T22:14:33+00:00 Ethan Hugg (ethanhugg):
>JWWang: Could you review this?   Thanks.


----
 reviewer JW Wang (jwwang) 
 time 2014-02-13T03:09:36+00:00 
2014-02-13T03:09:36+00:00 JW Wang (jwwang):
>It looks like the test checks if the alignment of bool is consistent in C++ and C (where bool type is emulated instead of a built-in type supported by the compiler).Isn&#39;t EXPECT_EQ(sizeof(bool), GetBoolSize()) enough to ensure all structure containing bool members will have the same memory layout?

2014-02-13T08:50:47+00:00 Martin Storsj (mstorsjo):
>That does sound like an easier test indeed, I&#39;ll update the test accordingly. The explicit test for sizeof(bool) might also be worthwhile to keep (at least since it says more explicitly what&#39;s wrong in case it fails) even though it probably is covered by testing the size of that struct as well.


----

----
 reviewer JW Wang (jwwang) 
 time 2014-02-13T09:32:16+00:00 
2014-02-13T09:32:16+00:00 JW Wang (jwwang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/76/

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/76/

----
**testing_done**:  
**description**:  Add a test that does basic verification that the interface ABI matches when used from CIdeally we would also want to check e.g. offsetof certain structmembers and alignof(bool) (but the latter is a C++11 feature only).Don&#39;t try to include stdbool.h on MSVCMSVC only got stdbool.h in MSVC 2013; provide a manual typedeffor this compiler instead.The C interface test makes sure that this doesn&#39;t break the ABI.

----
 time 2014-02-12T22:14:33+00:00 
2014-02-12T22:14:33+00:00 Ethan Hugg (ethanhugg):
>JWWang: Could you review this?   Thanks.


----
 reviewer JW Wang (jwwang) 
 time 2014-02-13T03:09:36+00:00 
2014-02-13T03:09:36+00:00 JW Wang (jwwang):
>It looks like the test checks if the alignment of bool is consistent in C++ and C (where bool type is emulated instead of a built-in type supported by the compiler).Isn&#39;t EXPECT_EQ(sizeof(bool), GetBoolSize()) enough to ensure all structure containing bool members will have the same memory layout?

2014-02-13T08:50:47+00:00 Martin Storsj (mstorsjo):
>That does sound like an easier test indeed, I&#39;ll update the test accordingly. The explicit test for sizeof(bool) might also be worthwhile to keep (at least since it says more explicitly what&#39;s wrong in case it fails) even though it probably is covered by testing the size of that struct as well.


----

----
 reviewer JW Wang (jwwang) 
 time 2014-02-13T09:32:16+00:00 
2014-02-13T09:32:16+00:00 JW Wang (jwwang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/76/

PR #284, opened 2014-02-13
Clean spatial picture alloc code

volvet:
Remove useless code.

in the function AllocSpatialPictures: which need not to call FreeMemorySvc on memory alloc fail, it is handled well at the code where AllocSpatialPicture is called. pls see the line 867 @ encoder_ext.cpp. So we can use pCtx to replace ppCtx.

Review: 
https://rbcommons.com/s/OpenH264/r/78/

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -590,8 +590,8 @@ static inline int32_t AcquireLayersNals (sWelsEncCtx** ppCtx, SWelsSvcCodingPara
 /*!
  * \brief	alloc spatial layers pictures (I420 based source pictures)
  */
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/78/

----
**testing_done**:  
**description**:  Remove useless code.  in the function AllocSpatialPictures:  which need not to call FreeMemorySvc on memory alloc fail,  it is handled well at the code where AllocSpatialPicture is called.   pls see the line 867 @ encoder_ext.cpp.  So we can use pCtx to replace ppCtx.    

----
 time 2014-02-13T23:17:45+00:00 
2014-02-13T23:17:45+00:00 Ethan Hugg (ethanhugg):
>This seems correct, merging it in.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/78/

PR #305, opened 2014-02-18
Don't duplicate the SEncParamExt/SEncParamBase structs for C compatibility

mstorsjo:
Also clean up the encoder parameter interface.

Posted for review at https://rbcommons.com/s/OpenH264/r/95/, although the individual steps are much more readable (with their individual explanations) here on github.

### DIFF
--- a/codec/api/svc/codec_app_def.h
+++ b/codec/api/svc/codec_app_def.h
@@ -116,12 +116,6 @@ typedef enum {
   VIDEO_CODING_LAYER = 1
 } LAYER_TYPE;
 
-/* SVC Encoder/Decoder Initializing Parameter Types */
-typedef enum {
-  INIT_TYPE_PARAMETER_BASED = 0,	// For SVC DEMO Application
-  INIT_TYPE_PARAMETER_EXT,			// For SVC CONSOLE Application
-} INIT_TYPE;
-
 //enumerate the type of video bitstream which is provided to decoder
 typedef enum {
   VIDEO_BITSTREAM_AVC               = 0,


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/95/

----
**testing_done**:  
**description**:  This is a pretty large set of patches, so if you want to see the individual steps, have a look at https://github.com/cisco/openh264/pull/305.Currently the SEncParamExt struct is duplicated for C compatibility - only because it inherits the SEncParamBase struct, for little actual benefit. Testing that the structs actually match ABI-wise is very tedious and very prone to break.This patchset avoids requiring to duplicate those structs, by restructuring the encoder parameter interface slightly. There&#39;s a number of smaller, slightly unrelated, fixes in the set as well, but they would conflict with the rest of the cleanup so I wouldn&#39;t be easily able to submit them separately.Split the encoder Initialize method into twoThis avoids the struct type casting and having to pass a parameterthat says which kind of struct it actually is.Use SEncParamExt in ProcessEncodingSvcWithParamBoth FillSpecificParameters and ParseCommandLine use parametersthat don&#39;t exist in SEncParamBase.Don&#39;t have SEncParamExt inherit SEncParamBaseInstead just duplicate the common fields. These fields had tobe duplicated for the C interface compatibility anyway - butthis way there is no risk to accidentally introduce an ABIbreak since there is no need for the layout of SEncParamBase toactually match the start of SEncParamExt.Don&#39;t copy out the SEncParamBase, manually read the individual fields insteadRespect the bEnableRc field in SEncParamExtRemove a field from SWelsSvcCodingParam that already exists in SEncParamExtMark the Initialize pointer parameter as constThis makes it clear to the callers that the struct willnot be modified by this call.Don&#39;t needlessly copy the input parameter structMake Param*Transcode take a const referenceThere is no need to modify the caller&#39;s struct.Move REC_FRAME_COUNT logging into the Initialize2 methodFix building with REC_FRAME_COUNT definedRemove unnecessary checks/initialization from the Initialize methodInitialize2 does the same checks and initializes the same fields.Simplify the CWelsH264SVCEncoder::Initialize2 methodRemove the useless iInitType parameter, make the methodprivate within CWelsH264SVCEncoder class, give the pointerparameter the correct type, avoiding needless casting.Remove the Initialize2 function from the external APIThis function no longer needs to be public since one can passSEncParamExt to the normal Initialize function.

----
 time 2014-02-18T19:33:02+00:00 
2014-02-18T19:33:02+00:00 Ethan Hugg (ethanhugg):
>This looks OK to me, but I would like someone on the original team to do a review tomorrow.

2014-02-19T01:11:46+00:00 Martin Storsj (mstorsjo):
>This modification is fine. I will merge this pull request.


----
 reviewer Karina Li (ruil2) 
 time 2014-02-19T01:11:52+00:00 
2014-02-19T01:11:52+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/95/

PR #322, opened 2014-02-20
init the spatial last spatial picture if excption

volvet:
this patch is try to fix the incorrect exception handle logic which is discussed before in preprocess module. 
the origin code: 
if the iCurTid or m_uiSpatialLayersInTemporal is invalid,  it will return false and need to force IDR frame.  But it is lack of re-order the spatial picture list,  then it cause the memory conflict when exchange the last spatial pictures(encoding next frame).  

The modification here means to init the last spatial pictures in the exception handle,  it is assume that the spatial picture list is in initialize status(so need not to reorder it).  Its OK because next frame should be IDR.  

Review:
https://rbcommons.com/s/OpenH264/r/99/

### DIFF
--- a/codec/encoder/core/src/wels_preprocess.cpp
+++ b/codec/encoder/core/src/wels_preprocess.cpp
@@ -405,6 +405,7 @@ int32_t CWelsPreProcess::UpdateSpatialPictures(sWelsEncCtx* pCtx, SWelsSvcCoding
     const int8_t iCurTid, const int32_t d_idx) {
   if (iCurTid < m_uiSpatialLayersInTemporal[d_idx] - 1 || pParam->iDecompStages == 0){
     if ((iCurTid >= MAX_TEMPORAL_LEVEL) || (m_uiSpatialLayersInTemporal[d_idx] - 1 > MAX_TEMPORAL_LEVEL)) {
+      InitLastSpatialPictures(pCtx);
       return 1;
     }
     if (pParam->bEnableLongTermReference && pCtx->bLongTermRefFlag[d_idx][iCurTid]) {


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/99/

----
**testing_done**:  
**description**:  this patch is try to fix the incorrect exception handle logic which is discussed before in preprocess module.   the origin code: if the iCurTid or m_uiSpatialLayersInTemporal is invalid,  it will return false and need to force IDR frame.  But it is lack of re-order the spatial picture list,  then it cause the memory conflict when exchange the last spatial pictures(encoding next frame).  The modification here means to init the last spatial pictures in the exception handle,  it is assume that the spatial picture list is in initialize status(so need not to reorder it).  Its OK because next frame should be IDR.  

----
 reviewer Sijia Chen (sijchen) 
 time 2014-02-21T06:54:14+00:00 
2014-02-21T06:54:14+00:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/wels_preprocess.cpp line408**: 
 id="comment96856">it seems that, in InitLastSpatialPictures() the evaluation of last 0 is needed but not needed to be -2, the evaluation of last 1 seemed not needed. this is not an issue but  may be removed when refactor the preprocess structure. 

2014-02-21T07:08:11+00:00 Volvet Zhang (Volvet):
>Yes,  the picture list manage logic is not good.    The refactor iterater need be continue.  


----
 reviewer Sijia Chen (sijchen) 
 time 2014-02-21T06:54:20+00:00 
2014-02-21T06:54:20+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/99/

PR #372, opened 2014-02-28
Prepare build system and code for adding ARM assembly optimizations

mstorsjo:


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/130/

----
**testing_done**:  
**description**:  Hook up building arm assembly in the generic arch and android makefilesCall WelsCPUFeatureDetect within ARM_ASM as wellAdd cpu feature detection for arm/linuxFor other platforms where we don&#39;t have arm cpu feature detection,this relies on using the HAVE_NEON define at compile time.For iOS, this isn&#39;t an issue because all iOS armv7 devices haveneon, but for android/linux there are armv7 devices that do nothave NEON, and thus need to have it enabled only at runtime.Update the makefile generator to support .S arm assembly filesThese are built if ASM_ARCH is set to arm.

----
 time 2014-02-28T23:07:52+00:00 
2014-02-28T23:07:52+00:00 Martin Storsj (mstorsjo):
>I saw some of the ARM assembly work in progress in a branch on github; I guess most of my work here is kinda redundant/conflicting with that - I&#39;ll rebase it on top of that after it&#39;s been merged. The only thing directly useful from this patchset might be the makefile generation part.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/130/

PR #387, opened 2014-03-03
Arm asm

zhilwang:
Add arm 32bit asm code for OpenH264 decoder. The review request is https://rbcommons.com/s/OpenH264/r/136/.

### DIFF
--- a/codec/decoder/core/inc/decoder_context.h
+++ b/codec/decoder/core/inc/decoder_context.h
@@ -134,7 +134,7 @@ typedef struct TagDeblockingFunc {
   PChromaDeblockingLT4Func  pfChromaDeblockingLT4Ver;
   PChromaDeblockingEQ4Func  pfChromaDeblockingEQ4Ver;
   PChromaDeblockingLT4Func  pfChromaDeblockingLT4Hor;
-  PChromaDeblockingEQ4Func  pfChromaDeblockinEQ4Hor;
+  PChromaDeblockingEQ4Func  pfChromaDeblockingEQ4Hor;
 } SDeblockingFunc, *PDeblockingFunc;
 
 typedef void (*PWelsBlockAddStrideFunc) (uint8_t* pDest, uint8_t* pPred, int16_t* pRes, int32_t iPredStride,
--- a/codec/decoder/core/src/deblocking.cpp
+++ b/codec/decoder/core/src/deblocking.cpp
@@ -384,7 +384,7 @@ void FilteringEdgeChromaIntraV (SDeblockingFilter* pFilter, uint8_t* pPixCb, uin
                           iBeta);
 
   if (iAlpha | iBeta) {
-    pFilter->pLoopf->pfChromaDeblockinEQ4Hor (pPixCb, pPixCr, iStride, iAlpha, iBeta);
+    pFilter->pLoopf->pfChromaDeblockingEQ4Hor (pPixCb, pPixCr, iStride, iAlpha, iBeta);
   }
   return;
 }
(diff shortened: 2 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/136/

----
**testing_done**:  
**description**:  Add arm asm code for OpenH264 decoder.1. Add arm_arch_common_macro.S and deblocking_neon.S to common folder which will be used by encoder also.2. Add block_add_neon.S, intra_pred_neon.S and mc_neon.S to decoder/core/arm folder which are used by decoder only.3. Modify welsdec and common ios project and related files.

----
 time 2014-03-03T00:11:50-08:00 
2014-03-03T00:11:50-08:00 Volvet Zhang (Volvet):
>The code is fine except some style issues. it is acceptable for me.  


----

----
 time 2014-03-03T01:06:19-08:00 
2014-03-03T01:06:19-08:00 Martin Storsj (mstorsjo):
>Looks good enough to me, any further fixes can be done on top of this later.
Ship it!


----
 time 2014-03-03T01:32:33-08:00 
2014-03-03T01:32:33-08:00 Martin Storsj (mstorsjo):
 on **codec/decoder/core/src/deblocking.cpp line733**: 
 id="comment99604">This function has been renamed in the latest git master version, so this will fail to build once you merge it. So I think you should rebase this on top of master to make sure it works when it is merged.

2014-03-03T17:27:17-08:00 Licai Guo (licaiguo):
>I have modified the function name


----

----
 time 2014-03-03T18:59:51-08:00 
2014-03-03T18:59:51-08:00 Ethan Hugg (ethanhugg):
>If this merges with the latest master, we should bring it in today.


----
 time 2014-03-03T19:08:06-08:00 
2014-03-03T19:08:06-08:00 Volvet Zhang (Volvet):
>I will merge it.  
Ship it!


----

----
 time 2014-03-04T01:07:49-08:00 
2014-03-04T01:07:49-08:00 Martin Storsj (mstorsjo):
 on **codec/decoder/core/src/deblocking.cpp line707**: 
 id="comment99947">Just for the record, this is NOT the way to handle issues when master has changed while your own branch was in development. Nobody ever does it like this, anywhere. It is not the correct way to resolve issues in git.In this case, the changes only were a couple of lines so it was no big thing, but if it would have been hundreds of lines, would you still have duplicated that change? (History shows that yes, you would because someone already did that. But it is NOT the right way.)In this situation, either rebase your changes on top of the lastest master (please read up on &quot;git rebase&quot;), possibly renaming your branch if you don&#39;t like rewriting public branches, or merge the master branch into your branch and fix the issue in the merge commit. Or even create a new branch and cherrypick your changes into that one and fix them up.Duplicating code changes like this only ruins the git history, making it harder to understand what was changed and why (making it harder for tools like &quot;git blame&quot;). Good, clean git history can be very valuable in debugging, allowing developers to read the history to figure out why something is the way it is, but if it is overwritten with shoddy changes with bad commit messages, nobody can learn anything from it.

2014-03-04T09:48:55-08:00 Ethan Hugg (ethanhugg):
>I&#39;d like to add one more Git detail here.  Most Windows installations of Git give you a choice of how to handle line endings.  Something like 1. always LF, 2. Convert to CRLF on checkout and back to LF on push, or 3. preserve whichever line ending is there.  The first two are good because then you cannot checkin with Windows-style line endings (CRLF).  This merge evidently had a bunch of those.  We need to keep the lines endings consistent in the repo.  Thanks.


----

Finish archiving from https://rbcommons.com/s/OpenH264/r/136/

PR #401, opened 2014-03-04
Enable the arm assembly in android builds

mstorsjo:


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/145/

----
**testing_done**:  
**description**:  Also enable arm assembly in iOS make builds of the plain library.Individual commits:---Add the APPLE_IOS define when building for iOS with makePreviously the define was only used when building from within thexcode project files.This enables the arm neon assembly.---Don&#39;t set -mfpu=neon while building the normal C++ filesThis will make the normal C++ code not run on non-NEON devices atall, making the runtime CPU feature detection pointless.Adding -mfpu=neon to CFLAGS is not necessary, it&#39;s enough toadd it while building those individual .S files (via ASMFLAGS).---Enable the android runtime CPU feature detectionBuild and include the android cpufeatures library in libcommon.---Hook up building arm assembly in the generic linux and android makefiles---Update the makefile generator to support .S arm assembly filesThese are built if ASM_ARCH is set to arm.

----
 reviewer Licai Guo (licaiguo) 
 time 2014-03-04T13:47:38+00:00 
2014-03-04T13:47:38+00:00 Licai Guo (licaiguo):
>It looks good to me.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/145/

PR #402, opened 2014-03-04
Unify the interface for the different variants of WelsCPUFeatureDetect

mstorsjo:
The caller of the function should not need to know exactly which
implementation of it is being used.

For the variants that don't support detecting the number of cores,
the pNumberOfLogicProcessors parameter can be left untouched
and the caller will use a higher level API for finding it out.

This simplifies all the calling code, and simplifies adding
more implementations of cpu feature detection.

Also add neon feature detection for generic arm/linux.

### DIFF
--- a/codec/common/cpu.cpp
+++ b/codec/common/cpu.cpp
@@ -38,6 +38,7 @@
  *************************************************************************************
  */
 #include <string.h>
+#include <stdio.h>
 #ifdef ANDROID_NDK
 #include <cpu-features.h>
 #endif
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/144/

----
**testing_done**:  
**description**:  Add cpu feature detection for generic arm/linuxFor platforms without runtime detection, assume whoever built itwith HAVE_NEON actually wanted it.Unify the interface for the different variants of WelsCPUFeatureDetectThe caller of the function should not need to know exactly whichimplementation of it is being used.For the variants that don&#39;t support detecting the number of cores,the pNumberOfLogicProcessors parameter can be left untouchedand the caller will use a higher level API for finding it out.This simplifies all the calling code, and simplifies addingmore implementations of cpu feature detection.

----
 time 2014-03-04T08:55:53+00:00 
2014-03-04T08:55:53+00:00 Volvet Zhang (Volvet):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/144/

PR #403, opened 2014-03-04
get cpu cores for android

volvet:
let cpu feature detect function be same with X86 and can get cpu cores. 
  if cpu feature functio in iOS can be  same with the X86 and android, it can bring simplification when calling  

Review: 
https://rbcommons.com/s/OpenH264/r/143/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/143/

----
**testing_done**:  
**description**:  let cpu feature detect function be same with X86 and can get cpu cores.  

----
 time 2014-03-04T08:05:36+00:00 
2014-03-04T08:05:36+00:00 Martin Storsj (mstorsjo):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/143/

PR #410, opened 2014-03-04
Allow building plain armeabi binaries for android

mstorsjo:
Building with "make OS=android APP_ABI=armeabi" will
produce arm binaries that will conform to the armeabi
ABI - not using any features outside of armv5te by
default (but still optionally using the NEON functions at
runtime if detected - even though such devices should rather
use the default armeabi-v7a build).

### DIFF
--- a/build/platform-android.mk
+++ b/build/platform-android.mk
@@ -4,10 +4,12 @@ SHAREDLIBSUFFIX = so
 GCCVERSION = 4.8
 NDKLEVEL = 12
 ifeq ($(ARCH), arm)
+  ifneq ($(APP_ABI), armeabi)
     CFLAGS += -march=armv7-a -mfloat-abi=softfp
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/158/

----
**testing_done**:  
**description**:  Allow building plain armeabi binaries for androidBuilding with &quot;make OS=android APP_ABI=armeabi&quot; willproduce arm binaries that will conform to the armeabiABI - not using any features outside of armv5te bydefault (but still optionally using the NEON functions atruntime if detected - even though such devices should ratheruse the default armeabi-v7a build).

----
 time 2014-03-10T09:44:09-07:00 
2014-03-10T09:44:09-07:00 Ethan Hugg (ethanhugg):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/158/

Issue #373, opened 2014-02-28
Updating encode size corrupts output


mmulhaupt:
I use InitializeExt to setup the encoder with a particular output resolution:
    _encodingParam.sSpatialLayers[0].iVideoWidth  = 640;
    _encodingParam.sSpatialLayers[0].iVideoHeight = 480;

I then update the encoding parameters after running for a bit:
    _encodingParam.sSpatialLayers[0].iVideoWidth  = 320;
    _encodingParam.sSpatialLayers[0].iVideoHeight = 240;
    _encoder->SetOption(ENCODER_OPTION_SVC_ENCODE_PARAM_EXT, &_encodingParam);

The display now seems to be covered in lots of green - even though the primary shape is visible.  I'm assuming that the Y component is being correctly scaled, but the Cr/Cb components aren't.  Even if I do a ForceIntraFrame(true) it doesn't fix.

Also the SetOption(ENCODER_OPTION_SVC_ENCODE_PARAM_EXT) is broken:
  welsEncodingExt.cpp - line 749.
sizeof (SEncParamBase) should be sizeof(SEncParamExt)

PR #425, opened 2014-03-06
fix bug of re-allocate memory for small-size SPS

huili2:
remove the error-remaining logic and variables to make sure active SPS is used.
It should fix the issue #373 
review request can be seen via:
https://rbcommons.com/s/OpenH264/r/153/

### DIFF
--- a/codec/decoder/core/inc/decoder_context.h
+++ b/codec/decoder/core/inc/decoder_context.h
@@ -252,9 +252,6 @@ typedef struct TagWelsDecoderContext {
   int32_t				iCsStride[3];		// strides for Cs
   int32_t				iRsStride[3];		// strides for Rs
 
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/153/

----
**testing_done**:  
**description**:  Old logic is not completely deleted after active SPS is used. Delete them.This guarantees the active SPS usage.

----
 reviewer Licai Guo (licaiguo) 
 time 2014-03-06T02:35:34+00:00 
2014-03-06T02:35:34+00:00 Licai Guo (licaiguo):
>It seems OK to me
Ship it!

2014-03-06T03:02:10+00:00 wayne liu (huili2):
>seems this case will keep re-allocating for same SPS. need consider this case.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/153/

PR #437, opened 2014-03-06
Fix building arm encoder assembly for android

mstorsjo:
https://rbcommons.com/s/OpenH264/r/166/

### DIFF
--- a/codec/common/arm_arch_common_macro.S
+++ b/codec/common/arm_arch_common_macro.S
@@ -44,6 +44,8 @@ mov pc, lr
 .endm
 #else
 
+.syntax unified
+
 .macro WELS_ASM_FUNC_BEGIN funcName
 .align 2
 .arm


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/166/

----
**testing_done**:  
**description**:  Explicitly enable the UAL syntax when using gnu toolsArm assembly has got two variants of the syntax, the old legacysyntax, and the new modern UAL (unified assembly language) syntax.Most arm assembly is the same in the both syntaxes, but someuncommon cases change the order of suffixes - the &quot;subscs&quot;instruction would be written &quot;subcss&quot; in the old syntax.The apple tools default to UAL, while the GNU tools (e.g. inandroid) require you to specify &quot;.syntax unified&quot; to enable thenew syntax. When enabling the new syntax with the GNU tools, somecases of &quot;sub r0, r1, lsl #1&quot; needs to be written explicitly as&quot;sub r0, r0, r1, lsl #1&quot;, handled in the previous commit.This allows using the same, modern syntax for things like subscs,without needing to have two alternate forms of writing it.Use the three-operand form of add/sub with shiftWhen using unified syntax, the two operand form with a shiftisn&#39;t allowed.Don&#39;t use WELS_ASM_FUNC_END in the middle of a functionWELS_ASM_FUNC_END declares the end of the function, and needsto be paired with WELS_ASM_FUNC_BEGIN.Rename a function to avoid conflicts between almost duplicate neon functionsThere&#39;s a different version of the same function in the encoder,but they&#39;re not identical - the encoder version has got stricteralignment requirements.If someone can confirm that it is ok to use the function from theencoder, pixel_sad_neon.S in processing could be deleted, and theencoder version moved to codec/common instead.Regenerate makefiles to include the encoder arm assemblyRename expand_picture.S to expand_picture_neon.SThis avoids ambiguity in the make based build system aboutwhether expand_picture.o should be built from expand_picture.Sor expand_picture.asm.

----
 time 2014-03-07T02:33:31+00:00 
2014-03-07T02:33:31+00:00 zhiliang wang (zhilwang):
 on **codec/processing/src/arm/pixel_sad_neon.S line38**: 
 id="comment101207">We will confirm this soon. If it is OK to use the function from encoder, We will merge it to codec/common


----
 time 2014-03-07T02:39:34+00:00 
2014-03-07T02:39:34+00:00 zhiliang wang (zhilwang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/166/

PR #445, opened 2014-03-07
Use the iMultipleThreadIdc field from SEncParamExt

mstorsjo:


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/174/

----
**testing_done**:  
**description**:  Use the iMultipleThreadIdc field from SEncParamExtThis makes sure that the iMultipleThreadIdc parameter in SEncParamExt that the user specified actually gets copied into the parameter struct within the encoder. Currently, the user cannot specify the actual number of threads (since the parameter goes uncopied and thus gets ignored), the only options are no threading at all, or automatic number of threads.

----
 reviewer Sijia Chen (sijchen) 
 time 2014-03-10T07:07:45+00:00 
2014-03-10T07:07:45+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/174/

PR #491, opened 2014-03-14
prevent from function re-do in MACRO WELS_CLIP1

huili2:
Suggest MACRO parameter not use function as input for performance consideration.
review request see:
https://rbcommons.com/s/OpenH264/r/194/

### DIFF
--- a/codec/decoder/core/src/mc.cpp
+++ b/codec/decoder/core/src/mc.cpp
@@ -181,7 +181,8 @@ static inline void McHorVer20_c (const uint8_t* pSrc, int32_t iSrcStride, uint8_
   int32_t i, j;
   for (i = 0; i < iHeight; i++) {
     for (j = 0; j < iWidth; j++) {
-      pDst[j] = WELS_CLIP1 ((FilterInput8bitWithStride_c (pSrc + j, 1) + 16) >> 5);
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/194/

----
**testing_done**:  
**description**:  Original code:#define WELS_CLIP1(iX) (((iX) &amp; ~255) ? (-(iX) &gt;&gt; 31) : (iX)) //iX not only a value but also can be an expressionWhen iX is a function, it will be called too many times, thus decreasing the performance.Use temporal value to store the result for WELS_CLIP1.

----
 reviewer JW Wang (jwwang) 
 time 2014-03-14T07:35:24+00:00 
2014-03-14T07:35:24+00:00 JW Wang (jwwang):
>Why not making WELS_CLIP1 a function?

2014-03-14T08:31:21+00:00 wayne liu (huili2):
>Hi JW, from my point, if everything is OK and no bug exists, I will use a best one, rather than a good one.In this code, if MACRO is surely guaranteed without error use, I still prefer MACRO. After all, the codes are finished, and whether a tool is good or not depends on how people use it. :)


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/194/

PR #512, opened 2014-03-17
  modify EncoderFrame return type.

ruil2:
review at: https://rbcommons.com/s/OpenH264/r/198/  
   error type is more meaning for EncoderFrame.
1. use error type instead of encoder type  in EncoderFrame

### DIFF
--- a/codec/api/svc/codec_api.h
+++ b/codec/api/svc/codec_api.h
@@ -64,7 +64,7 @@ class ISVCEncoder {
   virtual int EXTAPI Uninitialize() = 0;
 
   /*
-   * return: EVideoFrameType [IDR: videoFrameTypeIDR; P: videoFrameTypeP; ERROR: videoFrameTypeInvalid]
+   * return: 0 - success; otherwise -failed;
    */
   virtual int EXTAPI EncodeFrame (const SSourcePicture* kpSrcPic, SFrameBSInfo* pBsInfo) = 0;
   /*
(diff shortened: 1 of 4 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/198/

----
**testing_done**:  
**description**:     error type is more meaning for EncoderFrame.1. use error type instead of encoder type  in EncoderFrame   

----

----

----
 reviewer Sijia Chen (sijchen) 
 time 2014-03-17T09:40:50+00:00 
2014-03-17T09:40:50+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----
 time 2014-03-18T07:59:01+00:00 
2014-03-18T07:59:01+00:00 Martin Storsj (mstorsjo):
 on **codec/encoder/plus/src/welsEncoderExt.cpp line647**: 
 id="comment104857">This made the iFrameType variable almost useless. The frame type expressed in the public API enum isn&#39;t stored anywhere or exposed to the caller, so it feels a little unnecessary to even try to remap it.

2014-03-18T07:59:01+00:00 Martin Storsj (mstorsjo):
 on **test/BaseEncoderTest.cpp line85**: 
 id="comment104855">This check doesn&#39;t work as it should. info.eOutputFrameType uses the encoder internal WELS_FRAME_TYPE_SKIP = 0x0008, while videoFrameTypeSkip has a different value.Would it be simplest to make the encoder internally use the same videoFrameTypeSkip constants from the public header instead of having two sets of almost similar enums that have different values?


----

Finish archiving from https://rbcommons.com/s/OpenH264/r/198/

PR #534, opened 2014-03-18
    fix bug that there is no output in encoder console

ruil2:
review at : https://rbcommons.com/s/OpenH264/r/217/

### DIFF
--- a/codec/console/enc/src/welsenc.cpp
+++ b/codec/console/enc/src/welsenc.cpp
@@ -761,13 +761,13 @@ int ProcessEncodingSvcWithParam (ISVCEncoder* pPtrEnc, int argc, char** argv) {
     iStart	= WelsTime();
     long iEncode = pPtrEnc->EncodeFrame (pSrcPic, &sFbi);
     iTotal += WelsTime() - iStart;
-    if (videoFrameTypeInvalid == iEncode) {
+    if (cmResultSuccess != iEncode) {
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/217/

----
**testing_done**:  
**description**:  no update encoder console when encoder return value is changed  

----
 time 2014-03-18T08:01:00+00:00 
2014-03-18T08:01:00+00:00 Martin Storsj (mstorsjo):
 on **codec/console/enc/src/welsenc.cpp line770**: 
 id="comment104859">This looks very wrong - you&#39;re changing a check from &quot;skip != encode&quot; to &quot;skip == frametype&quot;.These checks don&#39;t work as they should at all, since sFbi.eOutputFrameType never gets set to videoFrameTypeSkip, it&#39;s only ever set to WELS_FRAME_TYPE_SKIP which has a different value.

2014-03-18T08:36:31+00:00 Karina Li (ruil2):
>different frame type for internal and external. which is unreasonable will use the same type for internal and external in the next update.


----

----
 time 2014-03-18T09:16:28+00:00 
2014-03-18T09:16:28+00:00 Martin Storsj (mstorsjo):
>I guess this looks ok as a first step forward.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/217/

PR #548, opened 2014-03-20
add set method for scene change detection

volvet:
the set method here is target to give the memory pointer to the scene change detection object,  by the compiler default copy (weak copy),   as the requirement from the algorithm, it need multiple static block idc memories for multi-reference, and the memory is maintained out of the processing module.   

Review: 
https://rbcommons.com/s/OpenH264/r/238/

### DIFF
--- a/codec/processing/src/scenechangedetection/SceneChangeDetection.h
+++ b/codec/processing/src/scenechangedetection/SceneChangeDetection.h
@@ -172,6 +172,13 @@ class CSceneChangeDetection : public IStrategy {
     return RET_SUCCESS;
   }
 
+  EResult Set(int32_t iType, void * pParam) {
+    if( pParam == NULL ){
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/238/

----
**testing_done**:  
**description**:  the set method here is target to give the memory pointer to the scene change detection object,  by the compiler default copy (weak copy),   as the requirement from the algorithm, it need multiple static block idc memories for multi-reference, and the memory is maintained out of the processing module.   

----
 reviewer james zhu (jamezhu) 
 time 2014-03-20T01:33:09+00:00 
2014-03-20T01:33:09+00:00 james zhu (jamezhu):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/238/

PR #561, opened 2014-03-21
Make the RC_MODES enum public, use it for the iRCMode field

mstorsjo:
This makes it clearer what values the iRCMode field should be set to.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/248/

----
**testing_done**:  
**description**:  Make the iRCMode field use the RC_MODES type instead of plain intThis makes it even clearer for users about how to set this field.Move the RC_MODES enum to the public headerThis allows users to know what values to set for the iRCModeparameter.

----
 reviewer Li Yao (lyao2) 
 time 2014-03-21T07:29:01+00:00 
2014-03-21T07:29:01+00:00 Li Yao (lyao2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/248/

PR #605, opened 2014-04-01
clean multi-threading macro

volvet:
Review:
https://rbcommons.com/s/OpenH264/r/270/

### DIFF
--- a/build/platform-android.mk
+++ b/build/platform-android.mk
@@ -35,7 +35,7 @@ SYSROOT = $(NDKROOT)/platforms/android-$(NDKLEVEL)/arch-$(ARCH)
 CXX = $(TOOLCHAINPREFIX)g++
 CC = $(TOOLCHAINPREFIX)gcc
 AR = $(TOOLCHAINPREFIX)ar
-CFLAGS += -DLINUX -DANDROID_NDK -DMT_ENABLED -fpic --sysroot=$(SYSROOT)
+CFLAGS += -DLINUX -DANDROID_NDK -fpic --sysroot=$(SYSROOT)
 CXXFLAGS += -fno-rtti -fno-exceptions
 LDFLAGS += --sysroot=$(SYSROOT)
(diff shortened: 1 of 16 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/270/

----
**testing_done**:  
**description**:  with the MT_EANBLE on,  the encoder can configured to multi-threading or single threading. So the code which is active when MT_ENABLE off can be simply removed.  This change is intend to simplify the multi-threading code.     

----

----
 reviewer james zhu (jamezhu) 
 time 2014-04-01T02:26:29+00:00 
2014-04-01T02:26:29+00:00 james zhu (jamezhu):
 on **codec/encoder/core/inc/param_svc.h line153**: 
 id="comment109706">Should param.iMultipleThreadIdc be default to 0 ? (0 is auto detect cpu cores, but 1 is disable)

2014-04-01T05:00:56+00:00 Volvet Zhang (Volvet):
>No, I think set the default for the iMultipleThreadIdc to 1(disable multithreading) is better.   

2014-04-01T05:00:56+00:00 Volvet Zhang (Volvet):
 on **codec/encoder/core/src/encoder_ext.cpp line784**: 
 id="comment109707">when multi-thread is disabled, I think this logic should be also active? because SM_DYN_SLICE is to limit slice size

2014-04-01T05:00:56+00:00 Volvet Zhang (Volvet):
>I do not see any special code here,   it just allocate memory for slice and mbs.   And the code only correct in single thread.   So I think it should be removed.   And the code will not be compiled at current build(MT_ENABLE is always be on).   


----
 reviewer james zhu (jamezhu) 
 time 2014-04-01T05:28:19+00:00 
2014-04-01T05:28:19+00:00 james zhu (jamezhu):
>OK, I misunderstand the old logic. Single thread can work when MT_ENABLE, the deleted code is redundant.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/270/

PR #646, opened 2014-04-08
Clean up the nasm commands in the MSVC project files

mstorsjo:
Remove the -O3 parameter which doesn't really exist according to "nasm -h" (and the corresponding actual option is on by default), and remove extra CRLF from the end of the commands.

### DIFF
--- a/codec/build/win32/dec/WelsDecCore.vcproj
+++ b/codec/build/win32/dec/WelsDecCore.vcproj
@@ -356,7 +356,7 @@
 						>
 						<Tool
 							Name="VCCustomBuildTool"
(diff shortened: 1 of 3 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/305/

----
**testing_done**:  
**description**:  Remove the literal CRLF (&amp;#x0D;&amp;#x0A;) within the nasm commands in msvc project filesRemove the explicit -O3 parameter to nasm in visual studio project filesAccording to &quot;nasm -h&quot;, there is no -O3 parameter at all, andthe highest optimization level (-Ox) is already the default.The corresponding parameter never was set when building with themake build system.

----
 reviewer Karina Li (ruil2) 
 time 2014-04-08T09:08:30+00:00 
2014-04-08T09:08:30+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/305/

PR #675, opened 2014-04-11
Silence MSVC warnings about "unsafe" standard C functions in console apps

mstorsjo:
The functions are used correctly, and we don't want to use large workarounds for the functions in the console apps. Within the library itself, a proper compatibility layer exists that takes care of this in a more elaborate way.

### DIFF
--- a/codec/console/enc/src/read_config.cpp
+++ b/codec/console/enc/src/read_config.cpp
@@ -38,14 +38,10 @@
  *
  *****************************************************************************/
 
+#define _CRT_SECURE_NO_WARNINGS
 #include <stdio.h>
 #include <string.h>
 #include "read_config.h"
-
-#if defined(_MSC_VER)
-#pragma warning(push)
-#pragma warning(disable:4996)
-#endif
 
 CReadConfig::CReadConfig()
   : m_pCfgFile (NULL)
@@ -141,7 +137,3 @@ const bool CReadConfig::ExistFile() {
 const string& CReadConfig::GetFileName() {
   return m_strCfgFileName;
 }
-
-#if defined(_MSC_VER)
-#pragma warning(pop)
-#endif


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/326/

----
**testing_done**:  
**description**:  Use _CRT_SECURE_NO_WARNINGS to avoid warnings in the encoder version of read_config.cpp as wellThis is clearer and requires less code than using the pragmas fordisabling warnings.Silence MSVC warnings about &quot;unsafe&quot; standard C functions in console appsWithin the libraries themselves, we use other wrapper functions tomake sure the functions are used safely, but this wrapper layer isn&#39;tavailable outside of the library in the console apps. The usage of thesefunctions is safe here, so silence the warnings.This gets rid of all warnings when building with MSVC in 32 bit mode,making it much easier to spot new warnings.

----
 reviewer Karina Li (ruil2) 
 time 2014-04-11T09:29:54+00:00 
2014-04-11T09:29:54+00:00 Karina Li (ruil2):
>my concern is : if some of warnings are the potential issues that we need to fix. But after using this macro, we can&#39;t find this issue.                whether are all warnings invisible when using this macro? 

2014-04-11T10:18:55+00:00 Martin Storsj (mstorsjo):
>That&#39;s a valid point, but I don&#39;t think it is necessary.One should be careful when using these functions regardless of if the compiler warns about them or not, and the warnings will not go away even if the use is correct.And this is only for some files in the console apps, not for the library itself which is more important and critical. The same warnings are already disabled in the encoder version of read_config.cpp anyway.


----
 time 2014-04-12T09:44:56+00:00 
2014-04-12T09:44:56+00:00 Volvet Zhang (Volvet):
>I can agree the change.   The macro only disable the warning on use of some old CRT functions which have newer, more security versions,  it does not means the old function is unsafe.  Unfortunately,   this new CRT functions are platform dependent,   this is the reason why we use wrap for those functions inside the library.  Since the warp is invisible to the console app, and the console app already use pragma to disable the security warning,  I think the change is a better way to suppress those warnings.  
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/326/

PR #680, opened 2014-04-14
update the default value for bPrefixNalAddingCtrl and iLoopFilterDisableIdc

ruil2:
Approved at: https://rbcommons.com/s/OpenH264/r/330/
1. update the default value of iLoopFilterDisableIdc as 0. In order to get the better quality, it is better to turn on deblocking filter 
2. update the default value of bPrefixNalAddingCtrl as false.  For NAL 14 is SVC NAL type, not AVC. but when the user only sets base parameter, we should remove this NAL type.
1. update hash code because the default values change.

### DIFF
--- a/codec/encoder/core/inc/param_svc.h
+++ b/codec/encoder/core/inc/param_svc.h
@@ -160,7 +160,7 @@ static void FillDefault (SEncParamExt& param) {
   // false: Streaming Video Sharing; true: Video Conferencing Meeting;
 
   /* Deblocking loop filter */
-  param.iLoopFilterDisableIdc		= 1;	// 0: on, 1: off, 2: on except for slice boundaries
+  param.iLoopFilterDisableIdc		= 0;	// 0: on, 1: off, 2: on except for slice boundaries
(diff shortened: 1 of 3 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/330/

----
**testing_done**:  
**description**:  1. update the default value of iLoopFilterDisableIdc as 0. In order to get the better quality, it is better to turn on deblocking filter 2. update the default value of bPrefixNalAddingCtrl as false.  For NAL 14 is SVC NAL type, not AVC. but when the user only sets base parameter, we should remove this NAL type.3. update hash code because the default values change.

----
2014-04-14T05:56:45+00:00 status-change:
update the default value of bPrefixNalAddingCtrl as false.  For NAL 14 is SVC NAL type, not AVC. but when the user only sets base parameter, we should remove this NAL type.
update hash code because the default values change.


----
 reviewer Sijia Chen (sijchen) 
 time 2014-04-14T06:01:47+00:00 
2014-04-14T06:01:47+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/330/

PR #698, opened 2014-04-16
update encDemo ios makefile for arm64 support

volvet:
Review: 
https://rbcommons.com/s/OpenH264/r/341/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/341/

----
**testing_done**:  
**description**:    Add target for arm64 architecture support for common/process/welsenc/encDemo.    BTW:  will raise another request for decoder.   

----
 time 2014-04-16T08:57:40+00:00 
2014-04-16T08:57:40+00:00 Martin Storsj (mstorsjo):
>This looks way too complicated. We should not add a separate version of the library just for the 64 bit build, it should be handled as any other architecture variant in the apple &quot;fat binaries&quot;.The change in https://github.com/cisco/openh264/pull/688 looks correct to me. I will send a rebased version of that change for review (with the original author&#39;s name intact).


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/341/

PR #738, opened 2014-04-23
Fix building the aarch64 assembly using gnu binutils

mstorsjo:


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/372/

----
**testing_done**:  
**description**:  Add macros for the non-standard mov.16b/mov.8b/ext.16b/ext.8bThis fixes building with gnu binutils, which don&#39;t support thisnonstandard form of the instructions.Once Apple&#39;s tools support the proper standard form of theinstructions, the code should be updated to use that everywhereinstead, and these macros should be removed.Use the correct ext syntax in the gnu version of macrosUse a plain &quot;ret&quot; instead of &quot;ret lr&quot;This fixes an issue with assembling with gnu binutils.Add correct arguments to the gnu version of UNPACK_FILTER_SINGLE_TAG_16BITSAdd \() to macro arguments to separate the argument from the following .8h or similarRemove .syntax unified from the aarch64 common headerThis directive isn&#39;t available in aarch64 code, only in arm code.

----
 reviewer Licai Guo (licaiguo) 
 time 2014-04-24T09:16:31+00:00 
2014-04-24T09:16:31+00:00 Licai Guo (licaiguo):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/372/

PR #750, opened 2014-04-25
Check for WELS_CPU_NEON before calling DeblockingBSCalcEnc_neon

mstorsjo:
Checking HAVE_NEON is not enough; e.g. android devices with
armeabi-v7a are not required to have NEON, so every use of such
functions should be check WELS_CPU_NEON in the cpu features
as well.

### DIFF
--- a/codec/encoder/core/inc/wels_func_ptr_def.h
+++ b/codec/encoder/core/inc/wels_func_ptr_def.h
@@ -94,6 +94,7 @@ typedef void (*PChromaDeblockingLT4Func) (uint8_t* iSampleCb, uint8_t* iSampleCr
     int32_t iBeta, int8_t* iTc);
 typedef void (*PChromaDeblockingEQ4Func) (uint8_t* iSampleCb, uint8_t* iSampleCr, int32_t iStride, int32_t iAlpha,
     int32_t iBeta);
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/381/

----
**testing_done**:  
**description**:  Check for WELS_CPU_NEON before calling DeblockingBSCalcEnc_neonChecking HAVE_NEON is not enough; e.g. android devices witharmeabi-v7a are not required to have NEON, so every use of suchfunctions should be check WELS_CPU_NEON in the cpu featuresas well.

----
 time 2014-04-25T09:38:41+00:00 
2014-04-25T09:38:41+00:00 Volvet Zhang (Volvet):
 on **codec/encoder/core/inc/wels_func_ptr_def.h line109**: 
 id="comment117518">I got confused.  Why NEON call code is quite different with the x86.   I do not think add iCpu is good solution.   the C code should be same, only assembly code be different,  right?

2014-04-25T10:01:45+00:00 Volvet Zhang (Volvet):
>Yes, ideally the assembly function should have a C version. Perhaps the reason is that splitting out the same code as a C function would have negative effect on x86 builds (if some tricks used in the neon code isn&#39;t available there).Anyway, I&#39;ve done a slightly better way of splitting this which at least avoids checking the cpu flags when calling functions, I&#39;ll update the review request.


----

----
 time 2014-04-25T10:13:03+00:00 
2014-04-25T10:13:03+00:00 Volvet Zhang (Volvet):
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/381/

PR #755, opened 2014-04-27
refine pNzc set access

licaiguo:
review ist at: https://rbcommons.com/s/OpenH264/r/382/

### DIFF
--- a/codec/decoder/core/src/decode_slice.cpp
+++ b/codec/decoder/core/src/decode_slice.cpp
@@ -440,7 +440,8 @@ int32_t WelsActualDecodeMbCavlcISlice (PWelsDecoderContext pCtx) {
 
   int32_t iMbX = pCurLayer->iMbX;
   int32_t iMbY = pCurLayer->iMbY;
-  int32_t iMbXy = pCurLayer->iMbXyIndex;
+  const int32_t iMbXy = pCurLayer->iMbXyIndex;
+  int8_t *pNzc = pCurLayer->pNzc[iMbXy];
   int32_t i;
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/382/

----
**testing_done**:  
**description**:  Use a temp variable pNzc to cache pCurLayer-&gt;pNzc[iMbXy]. This can improve the performance of the expression like ST32(&amp;pCurLayer-&gt;pNzc[iMbXy][0], 0). The reason is :pCurLayer-&gt;pNzc[iMbXy] points to a buffer, whenever we do write operation to it, the following access to pCurLayer-&gt;pNzc[iMbXy] must be load from memory sincethe compiler assumes that the memory pCurLayer-&gt;pNzc[iMbXy] points to may alias to the memory pCurLayer points to, which is not true in the code.Use a temp variable to cache the pointer can eliminate the redundant load.This can also improve the performance for generated javascript.

----
 time 2014-04-28T05:52:58+00:00 
2014-04-28T05:52:58+00:00 Juanny Wang (JuannyWang):
 on **codec/decoder/core/src/decode_slice.cpp line509**: 
 id="comment118200">Is it better to use sizeof(pNzc) instead of sizeof(pCurLayer-&gt;pNzc[iMbXy])?

2014-04-28T07:18:39+00:00 Licai Guo (licaiguo):
>Since I declared pNzc as int8_t*, using sizeof(pNzc) is the length of pointer, so it is not right.


----
 time 2014-04-28T07:45:18+00:00 
2014-04-28T07:45:18+00:00 Juanny Wang (JuannyWang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/382/

PR #768, opened 2014-04-29
use the parameters set by app instead of the default values

ruil2:
approved at: https://rbcommons.com/s/OpenH264/r/389/
1. set -1 as the default value of iNumRefFrame.  -1 means that encoder will select the the number of reference frames automatically.
2. use the parameters set by app instead of the default values


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/389/

----
**testing_done**:  
**description**:  1. set -1 as the default value of iNumRefFrame.  -1 means that encoder will select the the number of reference frames automatically.2. use the parameters set by app instead of the default valises

----
 reviewer Sijia Chen (sijchen) 
 time 2014-04-29T09:02:12+00:00 
2014-04-29T09:02:12+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/389/

PR #853, opened 2014-05-16
Fix encoder dequant Unit Test bug for master

dongzha:
Fix encoder dequant Unit Test bug
Review:
https://rbcommons.com/s/OpenH264/r/442/

### DIFF
--- a/test/encoder/EncUT_DecodeMbAux.cpp
+++ b/test/encoder/EncUT_DecodeMbAux.cpp
@@ -63,17 +63,18 @@ TEST(DecodeMbAuxTest, TestIhdm_4x4_dc) {
 TEST(DecodeMbAuxTest, TestDequant_4x4_luma_dc) {
   short T[16],W[16];
   srand((uint32_t)time(NULL));
-  for(int i=0;i<16;i++) {
-    T[i]=rand()%256+1;
-    W[i]=T[i];
-  }
 
-  //TODO: QP<18 will cause case fail, need fix and enable the test afterwards
-  for (int qp=18;qp<52;qp++) {
+  for (int qp=0; qp<12; qp++) {
+    for(int i=0; i<16; i++) {
+        T[i]=rand()%256+1;
+        W[i]=T[i];
+    }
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/442/

----
**testing_done**:  
**description**:  Fix QP&lt;18 dequant UT bug at encoder sideThe bug is caused by the root cause: T and W result will be use for next QP, so T need to be wrote back to T after dequantzation.Also, I move the generation of rand value inside the QP loop.However, In Original Code when QP&gt;=18, the result is always 0 because shift left a negtive value, so the bug is hidden.There is still some issue cofused me in the new Code,if I use        EXPECT_EQ((((T[i]*g_kuiDequantCoeff[qp%6][0]+(1 &lt;&lt; (1 -  qp / 6))))&gt;&gt;(2- qp / 6)),W[i]);instead of         T[i]= (((T[i]*g_kuiDequantCoeff[qp%6][0]+(1 &lt;&lt; (1 -  qp / 6))))&gt;&gt;(2- qp / 6));        EXPECT_EQ(T[i],W[i]);the result is still error.But when I use        short tmp = (((T[i]*g_kuiDequantCoeff[qp%6][0]+(1 &lt;&lt; (1 -  qp / 6))))&gt;&gt;(2- qp / 6));        EXPECT_EQ(tmp,W[i]);instead of         T[i]= (((T[i]*g_kuiDequantCoeff[qp%6][0]+(1 &lt;&lt; (1 -  qp / 6))))&gt;&gt;(2- qp / 6));        EXPECT_EQ(T[i],W[i]);the result is correct.Any comments?

----

----
 reviewer wayne liu (huili2) 
 time 2014-05-16T02:53:24+00:00 
2014-05-16T02:53:24+00:00 wayne liu (huili2):
>use QP&lt;12 is OK for that. For me, if QP&lt;12 is used, the UT can pass. I guess it is caused by left shift of negative ones when QP&gt;12. We can ignore that.Ship it.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/442/

PR #866, opened 2014-05-20
Add new seq check v1.0

huili2:
add a single function for new seq detection.
review request see:
https://rbcommons.com/s/OpenH264/r/448/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/448/

----
**discarded**

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/448/

PR #1001, opened 2014-06-23
remove unused param to prevent access violation

huili2:
see:
https://rbcommons.com/s/OpenH264/r/520/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/520/

----
**testing_done**:  
**description**:  1. kiSwapIdx may be less than 0, and it will cause array access violation.2. Seems pRefList-&gt;pRef[kiSwapIdx] is not used any more, but just here for return? Why is it needed? Suggest remove it.

----
 reviewer Karina Li (ruil2) 
 time 2014-06-23T06:02:46+00:00 
2014-06-23T06:02:46+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/520/

PR #1043, opened 2014-06-29
Get rid of global non-readonly variables within the library

mstorsjo:
This makes sure separate instances of the codec within a process don't interfere with each other in any way.

### DIFF
--- a/codec/encoder/core/src/mc.cpp
+++ b/codec/encoder/core/src/mc.cpp
@@ -81,18 +81,6 @@ static const uint8_t g_kuiABCD[8][8][4] = { ////g_kuiA[dy][dx], g_kuiB[dy][dx], 
     {4, 4, 28, 28}, {3, 5, 21, 35}, {2, 6, 14, 42}, {1, 7, 7, 49}
   }
 };
-typedef void (*WelsMcFunc0) (const uint8_t* pSrc, int32_t iSrcStride, uint8_t* pDst, int32_t iDstStride,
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/571/

----
**testing_done**:  
**description**:  This makes sure separate instances of the codec within a process don&#39;t interfere with each other in any way.Remove global function pointers in the encoder MC codeThis is similar to how this is done in the decoder part of MC.Avoid intermediate function pointers for some function pointers that are never overriddenThese are only used in the leaf C versions of functions anyway.Reduce code duplicationAvoid using static arrays for keeping track of frame dumping stateMove pfCavlcParamCal into the function pointer listThis avoids using a global variable for this function.Hardwire functions set up in WelsInitFillingPredFuncsThese are never overridden by any assembler optimized versions.(The assembler optimized intra predictor functions may use suchoptimized code internally anyway.)This avoids global writable data.Don&#39;t use and modify function pointers in global static tablesMake the tables const and keep a copy of the table in theSWelsFuncPtrList struct instead, where any modifications tothe table only touch that individual instance of the codec.

----
 reviewer Sijia Chen (sijchen) 
 time 2014-07-02T02:07:40+00:00 
2014-07-02T02:07:40+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/571/

PR #1049, opened 2014-07-01
Fix for memory leak of input frames in encoder.

ethanhugg:
This should fix this bug - https://bugzilla.mozilla.org/show_bug.cgi?id=1032492
Review here - https://rbcommons.com/s/OpenH264/r/576/

We were often returning without freeing the memory of the input frame.

### DIFF
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -397,6 +397,11 @@ class OpenH264VideoEncoder : public GMPVideoEncoder {
     }
 
     if (!has_frame) {
+      // This frame must be destroyed on the main thread.
+      g_platform_api->syncrunonmainthread (WrapTask (
+                                             this,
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/576/

----
**testing_done**:  
**description**:  This should solve the memory leak reported in Bugzilla - https://bugzilla.mozilla.org/show_bug.cgi?id=1032492This leak was happening often.  There are still error-condition leaks that could happen in this code that we can look into later.

----
 reviewer Randell Jesup (jesup) 
 time 2014-07-01T00:46:27+00:00 
2014-07-01T00:46:27+00:00 Randell Jesup (jesup):
>lgtm
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/576/

PR #1059, opened 2014-07-02
add max payload setting in gmp wrapper

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/584/

### DIFF
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -205,39 +205,93 @@ class OpenH264VideoEncoder : public GMPVideoEncoder {
       return GMPVideoGenericErr;
     }
 
-    SEncParamBase param;
-    memset (&param, 0, sizeof (param));
+    if (maxPayloadSize <= 0) {
+      SEncParamBase param;
+      memset (&param, 0, sizeof (param));
 
-    GMPLOG (GL_INFO, "Initializing encoder at "
-            << codecSettings.mWidth
-            << "x"
-            << codecSettings.mHeight
-            << "@"
-            << static_cast<int> (codecSettings.mMaxFramerate)
-            << "max payload size="
-            << maxPayloadSize);
+      GMPLOG (GL_INFO, "Initializing encoder at "
+              << codecSettings.mWidth
+              << "x"
+              << codecSettings.mHeight
+              << "@"
+              << static_cast<int> (codecSettings.mMaxFramerate)
+              << "max payload size="
+              << maxPayloadSize);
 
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/584/

----
**testing_done**:  
**description**:  <p>same with commit# a20c503.</p><p>1. we should support this setting though there is bad video quality in firefox (green block)<br />2. after merge this change, we should work with firefox to check why has bad quality.</p>

----
 time 2014-11-10T23:17:14+00:00 
2014-11-10T23:17:14+00:00 Ethan Hugg (ethanhugg):
 on **module/gmp-openh264.cpp line208**: 
 id="comment189784"><p>maxPayloadSize is a uint32_t so it cannot be less than zero.  It is also never passed in currently as zero so the mode 1 case takes the else with maxPayloadSize = 4M (see WebrtcGmpVideoCodec.cpp:200</p><p>So given that this works for Mode 1 with Firefox this way, should this be replaced by<br />if (maxPayloadSize == 0) maxPayloadSize = UINT32_MAX <br />or something similar in order to avoid all of the duplicate code in the two blocks below.</p>
maxPayloadSize is a uint32_t so it cannot be less than zero.  It is also never passed in currently as zero so the mode 1 case takes the else with maxPayloadSize = 4M (see WebrtcGmpVideoCodec.cpp:200
So given that this works for Mode 1 with Firefox this way, should this be replaced by<br />if (maxPayloadSize == 0) maxPayloadSize = UINT32_MAX <br />or something similar in order to avoid all of the duplicate code in the two blocks below.


----
2014-11-11T03:29:27+00:00 status-change:
remove repeat blocks by using SEncParamExt for all cases
same with commit# a20c503.
<span class="hl">1. </span>we should support this setting though there is bad video quality in firefox (green block)<br/></td></tr><tr class="replace-new"> <td class="marker">&nbsp;</td> <td class="marker">~</td> <td class="line rich-text"><span class="hl">2. </span>after merge this change, we should work with firefox to check why has bad quality.


----
 reviewer Sijia Chen (sijchen) 
 time 2014-11-11T22:57:26+00:00 
2014-11-11T22:57:26+00:00 Sijia Chen (sijchen):
><p>it will be better if we can remove unrelevant changes next time. I will wait for Ethan's ship before merging.</p>
it will be better if we can remove unrelevant changes next time. I will wait for Ethan's ship before merging.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/584/

PR #1060, opened 2014-07-02
Only force iLoopFilterDisableIdc to 2 if multithreading has been enabled

mstorsjo:
Previously these lines were within an #ifdef MT_ENABLED block,
but now that threading is enabled by default, we should probably
only do it if threading has been requested.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/586/

----
**testing_done**:  
**description**:  Only force iLoopFilterDisableIdc to 2 if multithreading has been enabledPreviously these lines were within an #ifdef MT_ENABLED block,but now that threading is enabled by default, we should probablyonly do it if threading has been requested.

----

----
 reviewer Karina Li (ruil2) 
 time 2014-07-04T02:43:16+00:00 
2014-07-04T02:43:16+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/586/

PR #1061, opened 2014-07-02
Use GetDefaultParams instead of memset to initialize SEncParamExt in the console test app

mstorsjo:
Most of the default parameters still are overwritten with hardcoded
values in FillSpecificParameters though, but this makes sure that
defaults are applied for any new fields at least.

### DIFF
--- a/codec/console/enc/src/welsenc.cpp
+++ b/codec/console/enc/src/welsenc.cpp
@@ -204,8 +204,6 @@ int ParseConfig (CReadConfig& cRdCfg, SSourcePicture* pSrcPic, SEncParamExt& pSv
   int32_t iRet = 0;
   int8_t iLayerCount = 0;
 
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/587/

----
**testing_done**:  
**description**:  Use GetDefaultParams instead of memset to initialize SEncParamExt in the console test appMost of the default parameters still are overwritten with hardcodedvalues in FillSpecificParameters though, but this makes sure thatdefaults are applied for any new fields at least.

----
 reviewer Karina Li (ruil2) 
 time 2014-07-04T02:21:11+00:00 
2014-07-04T02:21:11+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/587/

PR #1073, opened 2014-07-03
Hook up the x86_32-only assembly as well

mstorsjo:
This makes sure the code is at least somewhat tested. Ideally those
assembly functions should be upgraded to work in 64 bit mode as well,
but until then it's better to actually use them in the configurations
where they can be built.

Review at https://rbcommons.com/s/OpenH264/r/597/.

### DIFF
--- a/build/platform-x86-common.mk
+++ b/build/platform-x86-common.mk
@@ -17,6 +17,9 @@ ASMFLAGS_PLATFORM = -DX86_32
 endif
 ifeq ($(USE_ASM),Yes)
 CFLAGS += -DX86_ASM
+ifneq ($(ENABLE64BIT), Yes)
+CFLAGS += -DX86_32_ASM
+endif
 ASM_ARCH = x86
 endif
 ASM = nasm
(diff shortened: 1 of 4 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/597/

----
**testing_done**:  
**description**:  Hook up the x86_32-only assembly as wellThis makes sure the code is at least somewhat tested. Ideally thoseassembly functions should be upgraded to work in 64 bit mode as well,but until then it&#39;s better to actually use them in the configurationswhere they can be built.

----
 time 2014-07-03T06:48:03+00:00 
2014-07-03T06:48:03+00:00 zhiliang wang (zhilwang):
>Is working on x86 and arm 64bit asm code for downsample now, will give the code next week.

2014-07-03T06:53:10+00:00 Martin Storsj (mstorsjo):
>Ok, then there&#39;s probably not much need for this. Other than downsample, the only x86_32-only function left is CavlcParamCal_sse2.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/597/

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/597/

----
**testing_done**:  
**description**:  Hook up the x86_32-only assembly as wellThis makes sure the code is at least somewhat tested. Ideally thoseassembly functions should be upgraded to work in 64 bit mode as well,but until then it&#39;s better to actually use them in the configurationswhere they can be built.

----
 time 2014-07-03T06:48:03+00:00 
2014-07-03T06:48:03+00:00 zhiliang wang (zhilwang):
>Is working on x86 and arm 64bit asm code for downsample now, will give the code next week.

2014-07-03T06:53:10+00:00 Martin Storsj (mstorsjo):
>Ok, then there&#39;s probably not much need for this. Other than downsample, the only x86_32-only function left is CavlcParamCal_sse2.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/597/

PR #1082, opened 2014-07-04
Enable whole program optimization for both decoder and encoder MSVC projects

mstorsjo:
Enable it on the project level, instead of having to set separate options
for both compiler and linker.

The processing project actually had the options set in this way originally
as well.

Review at https://rbcommons.com/s/OpenH264/r/602/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/602/

----
**testing_done**:  
**description**:  Enable whole program optimization for both decoder and encoder MSVC projectsEnable it on the project level, instead of having to set separate optionsfor both compiler and linker.The processing project actually had the options set in this way originallyas well.

----

----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/602/

PR #1086, opened 2014-07-04
Unify oddly set options within MSVC projects

mstorsjo:
This set tries to make the options set in the MSVC projects consistent (with consistent optimization flags within the encoder and decoder parts), and removes a lot of leftover manual options which aren't necessary at all.

Review at https://rbcommons.com/s/OpenH264/r/606/.

### DIFF
--- a/codec/build/win32/dec/WelsDecPlus.vcproj
+++ b/codec/build/win32/dec/WelsDecPlus.vcproj
@@ -43,11 +43,6 @@
 			/>
 			<Tool
 				Name="VCMIDLTool"
-				PreprocessorDefinitions="NDEBUG"
-				MkTypLibCompatible="true"
-				TargetEnvironment="1"
-				TypeLibraryName=".\..\..\..\..\..\bin\win32\Release/WelsDecPlus.tlb"
-				HeaderFileName=""
 			/>
 			<Tool
 				Name="VCCLCompilerTool"
@@ -131,11 +126,6 @@
 			/>
 			<Tool
 				Name="VCMIDLTool"
-				PreprocessorDefinitions="NDEBUG"
-				MkTypLibCompatible="true"
-				TargetEnvironment="3"
-				TypeLibraryName=".\..\..\..\..\..\bin\win32\Release/WelsDecPlus.tlb"
-				HeaderFileName=""
 			/>
 			<Tool
 				Name="VCCLCompilerTool"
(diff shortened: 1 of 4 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/606/

----
**testing_done**:  
**description**:  This set tries to make the options set in the MSVC projects consistent (with consistent optimization flags within the encoder and decoder parts), and removes a lot of leftover manual options which aren&#39;t necessary at all.Don&#39;t set custom build options for VCMIDLToolThis tool isn&#39;t even used in the build (and thus, these settingsaren&#39;t even visible in the IDE).Don&#39;t set a custom path to browse information files, when no browse information is generatedRemove stray optimization options from the WelsEncPlus projectThere&#39;s not much performance critical code within this lib, allthe performance critical code in the encoder is in WelsEncCore andWelsVP, so there&#39;s little point in having an extra odd option enabledhere.Set optimizer options in WelsVP similarly to the other encoder projectsUse macros for setting the output directoryThis makes sure the same string shows up in the IDE gui in allthe configurations, simplifying editing multiple configurationsat the same time.This changes the output directory for 64 bit binaries frombin/win64 to bin/x64, but this is the common pattern used byMSVC in new projects.Use $(OutDir) instead of hardcoding the directory nameThis is how it is done in the other projects.Use default paths for and object files and other intermediate filesThis avoids accidentally reusing intermediates for one architecturewith the files for another, if e.g. switching from building forWin32/Debug to x64/Debug without doing a clean inbetween.Set linker debug options consistentlySet DebugInformationFormat consistently in all projectsUse the &quot;Program Database (/Zi)&quot; in release mode and in debugmode for x64, use &quot;Program Database for Edit &amp; Continue (/ZI)&quot;in debug mode for Win32.This is how new visual studio projects are set by default.Don&#39;t set a path to a precompiled header when precompiled headers are disabledRemove unnecessarily set AdditionalLibraryDirectoriesSome of these even pointed to directories that don&#39;t exist (anddon&#39;t get created during the build process).Remove some manually set options, that are set to the default valueThese are the only places where these options are overridden.Remove a stray option set on only one MSVC project

----
 time 2014-07-07T05:58:10+00:00 
2014-07-07T05:58:10+00:00 Martin Storsj (mstorsjo):
>Since it might be a bit hard to see what this actually does, all the individual steps are viewable at https://github.com/cisco/openh264/pull/1086, where this is done one isolated cleanup at a time.

2014-07-08T08:47:40+00:00 Martin Storsj (mstorsjo):
>Is there any interest in reviewing this? It is much easier to review if you look at the individual steps at https://github.com/cisco/openh264/pull/1086 instead. Or do you want me to send the cleanup one single commit at a time for review instead?


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/606/

PR #1102, opened 2014-07-08
add skip frame processing for screen content

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/620/

### DIFF
--- a/codec/encoder/core/src/encoder.cpp
+++ b/codec/encoder/core/src/encoder.cpp
@@ -325,10 +325,14 @@ EVideoFrameType DecideFrameType (sWelsEncCtx* pEncCtx, const int8_t kiSpatialNum
     } else {
       iFrameType = videoFrameTypeP;
     }
-    if (videoFrameTypeIDR == iFrameType) {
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/620/

----
**testing_done**:  
**description**:  add skip frame processing for screen content

----
 reviewer wayne liu (huili2) 
 time 2014-07-08T02:56:23-04:00 
2014-07-08T02:56:23-04:00 wayne liu (huili2):
 on **codec/encoder/core/src/encoder.cpp line328**: 
 id="comment144952">flag compared with 0 is not a good operation. suggest directly use flag.


----
 reviewer wayne liu (huili2) 
 time 2014-07-08T03:31:00-04:00 
2014-07-08T03:31:00-04:00 wayne liu (huili2):
>seems it is not a real flag, but an integer. OK for that, but suggest modify it in future.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/620/

PR #1135, opened 2014-07-09
Set -Zi instead of -ZI when building with MSVC in debug mode

mstorsjo:
The -ZI option is only supported when building for 32 bit x86,
when building for x64 or for arm, it gives the warning "/ZI is
not supported on this platform; enabling /Zi instead".

-ZI is for "edit & continue" debugging, which one probably wouldn't
be doing with a build outside of the IDE anyway, so keep it simple
instead of trying to set -ZI specifically for x86 32 bit builds.

Review at https://rbcommons.com/s/OpenH264/r/649/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/649/

----
**testing_done**:  
**description**:  Set -Zi instead of -ZI when building with MSVC in debug modeThe -ZI option is only supported when building for 32 bit x86,when building for x64 or for arm, it gives the warning &quot;/ZI isnot supported on this platform; enabling /Zi instead&quot;.-ZI is for &quot;edit &amp; continue&quot; debugging, which one probably wouldn&#39;tbe doing with a build outside of the IDE anyway, so keep it simpleinstead of trying to set -ZI specifically for x86 32 bit builds.

----
 time 2014-07-10T01:24:42+00:00 
2014-07-10T01:24:42+00:00 zhiliang wang (zhilwang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/649/

PR #1146, opened 2014-07-11
reuse dsNoParamSets in dec

huili2:
when dsNoParamSets, should not clear SPS/PPS buffer.
see:
https://rbcommons.com/s/OpenH264/r/660/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/660/

----
**testing_done**:  
**description**:  dsNoParamSets was originally used to clear the whole SPS/PPS buffer, as the info in API.Actually the buffer should only be cleared when memory has issues.So re-use this info only when real SPS/PPS errors for current data utility, when EC is enabled.This modification need not modify the API.It should be added for both v1.1 and master.

----
 time 2014-07-11T05:31:34+00:00 
2014-07-11T05:31:34+00:00 dong zhang (dongzhang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/660/

PR #1187, opened 2014-07-17
add interface UT for RC when changing temporal layer

lyao2:
review at:
https://rbcommons.com/s/OpenH264/r/682/

### DIFF
--- a/test/encoder/EncUT_EncoderExt.cpp
+++ b/test/encoder/EncUT_EncoderExt.cpp
@@ -265,6 +265,7 @@ TEST_F (EncoderInterfaceTest, RCTest) {
   #define MEM_VARY_SIZE 512
   #define TEST_FRAMES 500
 
+  pParamExt->iUsageType = (rand() % 2) ? SCREEN_CONTENT_REAL_TIME : CAMERA_VIDEO_REAL_TIME;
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/682/

----
**testing_done**:  
**description**:  dd interface UT for RC and code refactor

----
2014-07-23T07:14:58+00:00 status-change:
add interface UT for RC<span class="hl"> when</span> ch<span class="hl">anging temporal layer</span>
add interface UT for RC<span class="hl">,BS</span> ch<span class="hl">eck and code refactor</span>


----

----
 reviewer Karina Li (ruil2) 
 time 2014-07-24T02:28:57+00:00 
2014-07-24T02:28:57+00:00 Karina Li (ruil2):
 on **test/encoder/EncUT_EncoderExt.cpp line621**: 
 id="comment151644">what target is this case? there is no difference when only encoding one frame on iTemporalId,iSubSeqId,iLayerNum and so on.

2014-07-24T02:44:27+00:00 Li Yao (lyao2):
>the target is to see whether Encoder result is same when using Initialize() and SetOption() with same params.


----
2014-07-24T09:36:21+00:00 status-change:
<span class="hl">a</span>dd interface UT for RC<span class="hl">,BS check</span> and code refactor
dd interface UT for RC and code refactor


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/682/

PR #1194, opened 2014-07-18
Use macros for setting the output directory in MSVC project files

mstorsjo:
This makes sure this is set to the exact same string in all
the configurations, simplifying editing multiple configurations
at the same time.

This changes the output directory for 64 bit binaries from
bin/win64 to bin/x64, but this is the common pattern used by
MSVC in new projects.

Review at https://rbcommons.com/s/OpenH264/r/687/.

### DIFF
--- a/codec/build/win32/dec/WelsDecCore.vcproj
+++ b/codec/build/win32/dec/WelsDecCore.vcproj
@@ -23,7 +23,7 @@
 	<Configurations>
 		<Configuration
 			Name="Release|Win32"
-			OutputDirectory=".\..\..\..\..\bin\win32\Release"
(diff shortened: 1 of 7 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/687/

----
**testing_done**:  
**description**:  Use macros for setting the output directoryThis makes sure this is set to the exact same string in allthe configurations, simplifying editing multiple configurationsat the same time.This changes the output directory for 64 bit binaries frombin/win64 to bin/x64, but this is the common pattern used byMSVC in new projects.

----
 time 2014-07-18T08:59:11+00:00 
2014-07-18T08:59:11+00:00 zhiliang wang (zhilwang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/687/

PR #1207, opened 2014-07-23
Check that a working android toolchain was found

mstorsjo:
If the user selects a nonexistent ARCH, previously the call that
sets TOOLCHAINPREFIX would set it to contain a long error message,
which makes it quite hard to figure out what actually went wrong.

Instead check that CXX=$(TOLCHAINPREFIX)g++ actually points to a
working executable, and abort immediately if this isn't true.

Review at https://rbcommons.com/s/OpenH264/r/698/.

### DIFF
--- a/build/platform-android.mk
+++ b/build/platform-android.mk
@@ -46,6 +46,10 @@ CXXFLAGS += -fno-rtti -fno-exceptions
 LDFLAGS += --sysroot=$(SYSROOT)
 SHLDFLAGS = -Wl,--no-undefined -Wl,-z,noexecstack -Wl,-z,relro -Wl,-z,now -Wl,-soname,lib$(PROJECT_NAME).so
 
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/698/

----
**testing_done**:  
**description**:  Check that a working android toolchain was foundIf the user selects a nonexistent ARCH, previously the call thatsets TOOLCHAINPREFIX would set it to contain a long error message,which makes it quite hard to figure out what actually went wrong.Instead check that CXX=$(TOLCHAINPREFIX)g++ actually points to aworking executable, and abort immediately if this isn&#39;t true.

----

----
 time 2014-07-24T15:20:04+00:00 
2014-07-24T15:20:04+00:00 Ethan Hugg (ethanhugg):
>Ship It!
Ship It!
Ship it!


----

Finish archiving from https://rbcommons.com/s/OpenH264/r/698/

PR #1215, opened 2014-07-24
Mark the x86 assembly object files as not requiring an executable stack

mstorsjo:
This avoids having to add extra linker flags in order to specify this.

This is similar to how this already is handled for the arm assembly.

Review at https://rbcommons.com/s/OpenH264/r/704/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/704/

----
**testing_done**:  
**description**:  Mark the x86 assembly object files as not requiring an executable stackThis avoids having to add extra linker flags in order to specify this.This is similar to how this already is handled for the arm assembly.Consistently use -f elf instead of -f elf32The freebsd platform file already did things like this.

----
 time 2014-07-25T06:14:58+00:00 
2014-07-25T06:14:58+00:00 zhiliang wang (zhilwang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/704/

PR #1254, opened 2014-08-08
Add attributes incidating intentional data type aliasing, fix GetIntraPredictorTest on android

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/733/.

### DIFF
--- a/test/encoder/EncUT_GetIntraPredictor.cpp
+++ b/test/encoder/EncUT_GetIntraPredictor.cpp
@@ -6,7 +6,6 @@
 #include "macros.h"
 
 using namespace WelsSVCEnc;
-/* disable two UT to avoide fail on Android, root cause not found
 TEST (GetIntraPredictorTest, TestGetI4x4LumaPredV) {
   uint8_t* pPred = new uint8_t[64];
   uint8_t*	pRef  = new uint8_t[64];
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/733/

----
**testing_done**:  
**description**:  Add attributes incidating intentional data type aliasingInterpreting data of one type via a pointer of a different type is analiasing violating. This means that a compiler optimizer&#39;s analyzercan assume that data loaded into an array as uint32_t isn&#39;t relatedto data read out from the same array as uint64_t, and e.g. reorderloads/stores.Since these structs are intentionally used to load data via pointersof a wrong size, tell the compiler that these accesses may aliasother reads.This fixes the GetIntraPredictorTest tests of WelsI4x4LumaPredV_cand WelsI4x4LumaPredH_c. (The compiler optimizer did the wrong thingas long as WelsFillingPred8to16_c or WelsFillingPred8x2to16_c wereinlined into the calling function.)Reenable GetIntraPredictorTest tests that were broken on androidThese tests work as intended now.

----
 time 2014-08-08T09:08:09+00:00 
2014-08-08T09:08:09+00:00 dong zhang (dongzhang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/733/

PR #1256, opened 2014-08-08
Speed up test cases in EncoderInterfaceTest

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/735/.

### DIFF
--- a/test/encoder/EncUT_EncoderExt.cpp
+++ b/test/encoder/EncUT_EncoderExt.cpp
@@ -138,8 +138,10 @@ void EncoderInterfaceTest::InitializeParamExt() {
 
 TEST_F (EncoderInterfaceTest, TemporalLayerSettingTest) {
 
-  pParamExt->iPicWidth = m_iWidth;
-  pParamExt->iPicHeight = m_iHeight;
+  pParamExt->iPicWidth = 1280;
+  pParamExt->iPicHeight = 720;
+  m_iWidth = pParamExt->iPicWidth;
+  m_iHeight = pParamExt->iPicHeight;
   pParamExt->iTargetBitrate = 60000;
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/735/

----
**testing_done**:  
**description**:  When running the unit tests in valgrind, the few test cases in EncoderInterfaceTest took over 60% of the runtime of the whole test suite, even though the tests mostly test the same thing over and over very many times.Instead of repeating some parts in one test a huge number of times, the whole test suite can be run multiple times if one wants more test coverage. Previously this gave excessive attention to these few tests.Set a specific frame size in TemporalLayerSettingTestPreviously, this test used whatever size was set in m_iWidthand m_iHeight before, which depended on the order that the testswere executed. When this test was the first one executed in theEncoderInterfaceTest, the width and height were set to the max.Instead of having the test behaviour depend on the test order,set a specific size, just as InitializeParamExt and MemoryCheckTestdo.This reduces the runtime of TemporalLayerSettingTest from 86 secondsto 26 seconds, when run in valgrind.Reduce the number of frames tested in EncoderInterfaceTestThere&#39;s little point in running the same test over and over fora huge number of frames if it doesn&#39;t test much different things.This reduces the runtime of EncoderInterfaceTest.* from 322 secondsto 140 seconds, when running in valgrind.

----
 time 2014-08-08T10:25:15+00:00 
2014-08-08T10:25:15+00:00 Martin Storsj (mstorsjo):
>This seems to reduce the runtime of the unittest travis runs by about 3-5 minutes each.


----
 time 2014-08-11T02:47:28+00:00 
2014-08-11T02:47:28+00:00 dong zhang (dongzhang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/735/

PR #1257, opened 2014-08-08
Explicitly add .syntax unified when building for iOS

mstorsjo:
This is the default when building with the clang built-in assembler,
but not if using the external assembler - thus always specify it,
for clarity.

Also use the three-operand for of a sub instruction in BS_NZC_CHECK.
The same is already done in the gnu version of the macro.

This fixes building most of the arm assembly with Apple's external
assembler. While this isn't a necessary goal in itself, there's no
harm in doing this either.

Review at https://rbcommons.com/s/OpenH264/r/736/.

### DIFF
--- a/codec/common/arm/arm_arch_common_macro.S
+++ b/codec/common/arm/arm_arch_common_macro.S
@@ -30,6 +30,8 @@
  *
  */
 
+.syntax unified
+
 #ifdef __APPLE__
 
 .macro WELS_ASM_FUNC_BEGIN
@@ -44,7 +46,6 @@ mov pc, lr
 .endm
 #else
 
-.syntax unified
 .section .note.GNU-stack,"",%progbits // Mark stack as non-executable
 .text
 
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/736/

----
**testing_done**:  
**description**:  Explicitly add .syntax unified when building for iOSThis is the default when building with the clang built-in assembler,but not if using the external assembler - thus always specify it,for clarity.Also use the three-operand for of a sub instruction in BS_NZC_CHECK.The same is already done in the gnu version of the macro.This fixes building most of the arm assembly with Apple&#39;s externalassembler. While this isn&#39;t a necessary goal in itself, there&#39;s noharm in doing this either.

----
 time 2014-08-11T00:39:56+00:00 
2014-08-11T00:39:56+00:00 zhiliang wang (zhilwang):
>Ship It!
Ship It!
Ship it!


----

----
 time 2014-08-15T05:27:45+00:00 
2014-08-15T05:27:45+00:00 dong zhang (dongzhang):
 on **codec/common/arm/arm_arch_common_macro.S line33**: 
 id="comment160362">Seems Xcodec5.1 build fail when &quot;.syntax unified&quot; is added, can anyone comform this?

2014-08-15T05:34:09+00:00 zhiliang wang (zhilwang):
>I also meet the same problem. It will build fail in Xcodec5.1


----

Finish archiving from https://rbcommons.com/s/OpenH264/r/736/

PR #1259, opened 2014-08-08
Remove a needless, no-op change of the number of threads

mstorsjo:
If kiCpuCores < 2, then iCountThreadsNum (and iMultipleThreadIdc)
can't be >= 2, because they're initialized with
WELS_MIN (kiCpuCores, ...) just a few lines above.

If iMultipleThreadIdc is initially set to 0 by the caller, this
removed piece of code would change it to 1, if kiCpuCores < 2.

When iMultipleThreadIdc is changed from the originally set value,
a call to WelsEncoderParamAdjust with the original parameters
would reset the whole codec since iMultipleThreadIdc differs.

This fixes running EncoderInterfaceTest.TemporalLayerSettingTest
on machines where the detected number of cores is 1.

Review at https://rbcommons.com/s/OpenH264/r/738/.

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -1862,14 +1862,6 @@ int32_t InitSliceSettings (SLogContext* pLogCtx, SWelsSvcCodingParam* pCodingPar
   pCodingParam->iCountThreadsNum				= WELS_MIN (kiCpuCores, iMaxSliceCount);
   pCodingParam->iMultipleThreadIdc	= pCodingParam->iCountThreadsNum;
 
-#ifndef WELS_TESTBED	// for product release and non-SGE testing
-
-  if (kiCpuCores < 2) {	// single CPU core, make no sense for MT parallelization
-    pCodingParam->iMultipleThreadIdc	= 1;
-    pCodingParam->iCountThreadsNum				= 1;
-  }
-#endif
-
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/738/

----
**testing_done**:  
**description**:  Remove a needless, no-op change of the number of threadsIf kiCpuCores &lt; 2, then iCountThreadsNum (and iMultipleThreadIdc)can&#39;t be &gt;= 2, because they&#39;re initialized withWELS_MIN (kiCpuCores, ...) just a few lines above.If iMultipleThreadIdc is initially set to 0 by the caller, thisremoved piece of code would change it to 1, if kiCpuCores &lt; 2.When iMultipleThreadIdc is changed from the originally set value,a call to WelsEncoderParamAdjust with the original parameterswould reset the whole codec since iMultipleThreadIdc differs.This fixes running EncoderInterfaceTest.TemporalLayerSettingTeston machines where the detected number of cores is 1.This is one out of two different ways of fixing this issue, but the both solutions don&#39;t conflict so both can be applied if both are deemed ok.

----
 reviewer Karina Li (ruil2) 
 time 2014-08-11T05:31:38+00:00 
2014-08-11T05:31:38+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/738/

PR #1280, opened 2014-08-15
Add ifdef HAVE_NEON around the contents of arm_arch_common_macro.S

mstorsjo:
This file is built on its own from within the xcode projects,
even though it isn't necessary. Previously its contents was just
empty, but now a .syntax unified was added, which failed the build
when building for arm64.

Make this file a no-op, just like the other arm assembly source files,
unless HAVE_NEON is defined.

Review at https://rbcommons.com/s/OpenH264/r/757/.

### DIFF
--- a/codec/common/arm/arm_arch_common_macro.S
+++ b/codec/common/arm/arm_arch_common_macro.S
@@ -30,6 +30,8 @@
  *
  */
 
+#ifdef HAVE_NEON
+
 .syntax unified
 
 #ifdef __APPLE__
@@ -62,4 +64,6 @@ mov pc, lr
 mov pc, lr
 .endfunc
 .endm
+#endif
+
 #endif


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/757/

----
**testing_done**:  
**description**:  Add ifdef HAVE_NEON around the contents of arm_arch_common_macro.SThis file is built on its own from within the xcode projects,even though it isn&#39;t necessary. Previously its contents was justempty, but now a .syntax unified was added, which failed the buildwhen building for arm64.Make this file a no-op, just like the other arm assembly source files,unless HAVE_NEON is defined.

----
 time 2014-08-15T06:11:22+00:00 
2014-08-15T06:11:22+00:00 dong zhang (dongzhang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/757/

PR #1281, opened 2014-08-15
Fix building svc_motion_estimation.S for android

mstorsjo:
The .align directive takes an argument in number of bits, i.e. the
actual alignment is 2^n. Previously building with binutils failed,
since 16 isn't a valid parameter to .align, the maximum is 15.

Thus, this makes the code try to align to 16 bytes, instead of aligning
to 65536 bytes.

This fixes building for android.

This also clears up the same mistake in the aarch64 code, even though
that one built just fine.

Review at https://rbcommons.com/s/OpenH264/r/758/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/758/

----
**testing_done**:  
**description**:  Fix building svc_motion_estimation.S for androidThe .align directive takes an argument in number of bits, i.e. theactual alignment is 2^n. Previously building with binutils failed,since 16 isn&#39;t a valid parameter to .align, the maximum is 15.Thus, this makes the code try to align to 16 bytes, instead of aligningto 65536 bytes.This fixes building for android.This also clears up the same mistake in the aarch64 code, even thoughthat one built just fine.

----
 time 2014-08-15T10:16:21+03:00 
2014-08-15T10:16:21+03:00 dong zhang (dongzhang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/758/

PR #1285, opened 2014-08-19
avoid encoder crash with random pixel input

dongzha:
review at:
https://rbcommons.com/s/OpenH264/r/761/

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -2965,6 +2965,9 @@ bool CheckFrameSkipBasedMaxbr (sWelsEncCtx* pCtx, int32_t iSpatialNum) {
  * \return	EFrameType (videoFrameTypeIDR/videoFrameTypeI/videoFrameTypeP)
  */
 int32_t WelsEncoderEncodeExt (sWelsEncCtx* pCtx, SFrameBSInfo* pFbi, const SSourcePicture* pSrcPic) {
+  if( pCtx == NULL ) {
+    return ENC_RETURN_MEMALLOCERR;
+  }
   SLayerBSInfo* pLayerBsInfo					= &pFbi->sLayerInfo[0];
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/761/

----
**testing_done**:  
**description**:  avoid encoder crash with random pixel inputUse rand pixel value to generate  a picture and encode,  the encoder will crash.Flowing error return will lead encoder crash.  //since for each 0x000 need a 0x03, so the needed length will not exceed (iAssumeNeedLenth + iAssumeNeedLength/3), here adjust to &gt;&gt;1 to omit division  if (kiDstBufferLen &lt; (iAssumedNeededLength + (iAssumedNeededLength &gt;&gt; 1))) {    return ENC_RETURN_MEMALLOCERR;    //TODO: call the realloc&amp;copy instead  }When it return ENC_RETURN_MEMALLOCERR, EncodeFrameInternal will call WelsUninitEncoderExt and release pCtx.But the console code does not call InitEncoderExt again, so pCtx is still NULL when call WelsEncoderEncodeExt next frame,This will lead crash.The patch fixes the crash by exam pCtx inside WelsEncoderEncodeExt, if NULL, it will return ENC_RETURN_MEMALLOCERR.Note: we may need to call  InitEncoderExt again. But it is not the scope of this patch. 

----
 reviewer wayne liu (huili2) 
 time 2014-08-19T07:50:39+00:00 
2014-08-19T07:50:39+00:00 wayne liu (huili2):
>suggest add corresponding UT for that.
Ship it!


----
 reviewer Karina Li (ruil2) 
 time 2014-08-19T08:03:04+00:00 
2014-08-19T08:03:04+00:00 Karina Li (ruil2):
>when the size of bitstream is larger than the size of buffer, we should re-allocate the buffer,not return.this patch only can be a temporary version , we need to think it about later.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/761/

PR #1318, opened 2014-08-29
fix auto threadIdc issue

ruil2:
approved at: https://rbcommons.com/s/OpenH264/r/783/

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -1946,35 +1946,13 @@ void OutputCpuFeaturesLog (SLogContext* pLogCtx, uint32_t uiCpuFeatureFlags, uin
            iCacheLineSize);
 }
 
-/*!
- * \brief	initialize Wels avc encoder core library
- * \pParam	ppCtx		sWelsEncCtx**
- * \pParam	pParam		SWelsSvcCodingParam*
- * \return	successful - 0; otherwise none 0 for failed
- */
-int32_t WelsInitEncoderExt (sWelsEncCtx** ppCtx, SWelsSvcCodingParam* pCodingParam, SLogContext* pLogCtx) {
-  sWelsEncCtx* pCtx		= NULL;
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/783/

----
**testing_done**:  
**description**:  <p>when user sets iMultipleThreadIdc as 0,the encoder will change iMultipleThreadIdc according to the number of cores and the number of slices, then iMultipleThreadIdc will be set non-zero value.</p><p>if the user updates the other parameters with keeping iMultipleThreadIdc as 0. the encoder will be reset because the difference between old iMultipleThreadIdc and new iMultipleThreadIdc.</p><p>if old iMultipleThreadIdc is different from new iMultipleThreadIdc,the encoder will be initilized again. in order to avoid the repeat initilization, update iMultipleThreadIdc before checking</p>

----
 time 2014-08-29T05:53:11+00:00 
2014-08-29T05:53:11+00:00 dong zhang (dongzhang):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----
2014-08-29T05:53:24+00:00 status-change:
when user sets iMultipleThreadIdc as 0,the encoder will change iMultipleThreadIdc according to the number of cores and the number of slices, then iMultipleThreadIdc will be set non-zero value.
if the user updates the other parameters with keeping iMultipleThreadIdc as 0. the encoder will be reset because the difference between old iMultipleThreadIdc and new iMultipleThreadIdc.
if old iMultipleThreadIdc is different from new iMultipleThreadIdc,the encoder will be initilized again. in order to avoid the repeat initilization, update iMultipleThreadIdc before checking


----
 reviewer wayne liu (huili2) 
 time 2014-08-29T05:55:13+00:00 
2014-08-29T05:55:13+00:00 wayne liu (huili2):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/783/

PR #1351, opened 2014-09-12
enlarge QP range when skipframeflag off on BitRate mode

lyao2:
review at:
https://rbcommons.com/s/OpenH264/r/797/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/797/

----
**testing_done**:  
**description**:  <p>enlarge QP range when skipframeflag off on BitRate mode</p>

----
 reviewer Karina Li (ruil2) 
 time 2014-09-10T02:04:45+00:00 
2014-09-10T02:04:45+00:00 Karina Li (ruil2):
 on **codec/encoder/core/src/ratectl.cpp line515**: 
 id="comment170052"><p>it is better to remove the processing for RC_LOW_BW_MODE. this mode is same meaning with RC_BITRATE_MODE without skip.</p>
it is better to remove the processing for RC_LOW_BW_MODE. this mode is same meaning with RC_BITRATE_MODE without skip.


----
 time 2014-09-10T08:05:34+00:00 
2014-09-10T08:05:34+00:00 Martin Storsj (mstorsjo):
><p>I'm not so fond of this change. As Karina said, this makes RC_BITRATE_MODE with skip==false identical to RC_LOW_BW_MODE. When encoding with RC_LOW_BW_MODE (or with RC_BITRATE_MODE with skip==false with this change), the QP varies very much, and when frames have QP &gt; 35 it looks really really bad. At least for my use cases, I rather allow the bitrate to exceed the limit a little (so the rate control honors the iMaxQp parameter) than making it look really bad. Right now I have the choice between RC_BITRATE_MODE and RC_LOW_BW_MODE, but after this change it would always look bad if I set skip==false.</p>
I'm not so fond of this change. As Karina said, this makes RC_BITRATE_MODE with skip==false identical to RC_LOW_BW_MODE. When encoding with RC_LOW_BW_MODE (or with RC_BITRATE_MODE with skip==false with this change), the QP varies very much, and when frames have QP &gt; 35 it looks really really bad. At least for my use cases, I rather allow the bitrate to exceed the limit a little (so the rate control honors the iMaxQp parameter) than making it look really bad. Right now I have the choice between RC_BITRATE_MODE and RC_LOW_BW_MODE, but after this change it would always look bad if I set skip==false.

2014-09-10T08:19:34+00:00 Martin Storsj (mstorsjo):
><p>This change aims to merge RC_BITRATE_MODE and RC_LOW_BW_MODE when skip==false. And later I will eliminate RC_LOW_BW_MODE. RC_BITRATE_MODE should works on target bitrate and normallu skip is true. If we want to maintain video quality, we'd better use Quality MODE.</p>
Or perhaps I just misunderstand the RC modes? Perhaps I just should use RC_QUALITY_MODE instead? There really are no explanations in the comments/headers to what the difference is between the modes. Apparently RC_QUALITY_MODE also tries to follow iTargetBitrate (more or less).
This change aims to merge RC_BITRATE_MODE and RC_LOW_BW_MODE when skip==false. And later I will eliminate RC_LOW_BW_MODE. RC_BITRATE_MODE should works on target bitrate and normallu skip is true. If we want to maintain video quality, we'd better use Quality MODE.


----
2014-09-10T09:43:52+00:00 status-change:
Remove RC_LOW_BW_MODE and merge with Bitrate Mode


----
 reviewer Karina Li (ruil2) 
 time 2014-09-11T04:51:13+00:00 
2014-09-11T04:51:13+00:00 Karina Li (ruil2):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/797/

PR #1359, opened 2014-09-16
Simplify code by getting rid of unnecesary SafeDelete/SafeFree macros

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/813/.

### DIFF
--- a/codec/processing/src/common/util.h
+++ b/codec/processing/src/common/util.h
@@ -102,9 +102,6 @@ inline EMethods WelsVpGetValidMethod (int32_t a) {
 }
 
 
-#define _SafeFree(p)		if (p) { WelsFree(p); (p) = NULL; }
-#define _SafeDelete(p)		if (p) { delete (p); (p) = NULL; }
-
 
 WELSVP_NAMESPACE_END
 


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/813/

----
**testing_done**:  
**description**:  <p>Remove some now unused macros</p><p></p><p></p><p>Simplify code by calling WelsFree directly instead of using a helper function</p><p>This makes the code simpler and clearer. It's not necessary to use the<br />_SafeFree helper macro in neither of the case - in one case the<br />pointer is overwritten immediately, in the other case it's witihin<br />the destructor where the pointer won't ever be read again.</p><p></p><p>Use delete instead of the _SafeDelete macro</p><p>Since this is the destructor of the class, the member variables<br />won't be read afterwards, and setting the pointers to NULL afterwards<br />is redundant.</p><p></p><p>Remove unnecesary use of _SafeDelete</p><p>It is not necessary to check whether a pointer is NULL before<br />deleting it, "delete NULL" is explicitly ok.</p><p>The deleted pointer is a local variable here, so setting it to<br />NULL after deleting isn't necessary in these cases either.</p>

----
 time 2014-09-17T00:49:14+00:00 
2014-09-17T00:49:14+00:00 dong zhang (dongzhang):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/813/

PR #1418, opened 2014-10-17
ec disable bug fix

huili2:
fix the case when EC is disabled, to prevent following construction.
see:
https://rbcommons.com/s/OpenH264/r/850/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/850/

----
**testing_done**:  
**description**:  <p>When EC is disabled, original logic would go on decoding with error info return.<br />Should prevent following decoding.</p>

----
 reviewer Haibo Zhu (haibozhu) 
 time 2014-10-16T09:27:35+00:00 
2014-10-16T09:27:35+00:00 Haibo Zhu (haibozhu):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/850/

PR #1430, opened 2014-10-22
Set LDFLAGS after the object files in shared library linking commands

mstorsjo:
This makes sure that the built libopenh264.so actually links to
libpthread.so.

Review at https://rbcommons.com/s/OpenH264/r/861/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/861/

----
**testing_done**:  
**description**:  Set LDFLAGS after the object files in shared library linking commandsThis makes sure that the built libopenh264.so actually links tolibpthread.so.

----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/861/

PR #1450, opened 2014-10-24
Add casts to avoid warnings about comparison between signed and unsigned

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/872/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/872/

----
**testing_done**:  
**description**:  Add casts to avoid warnings about comparison between signed and unsigned

----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/872/

PR #1472, opened 2014-10-29
Make sure the width/height is large enough in SetOptionEncParamExt

mstorsjo:
When using 4 downsampled layers, make sure the random width/height
is large enough that the downsampled layers don't have a zero size.

Review at https://rbcommons.com/s/OpenH264/r/881/.

### DIFF
--- a/test/api/encode_decode_api_test.cpp
+++ b/test/api/encode_decode_api_test.cpp
@@ -1985,8 +1985,8 @@ TEST_F (EncodeDecodeTestAPI, Engine_SVC_Switch_P) {
 
 
 TEST_F (EncodeDecodeTestAPI, SetOptionEncParamExt) {
-  int iWidth       = WELS_CLIP3 ((((rand() % MAX_WIDTH) >> 1) + 1) << 1,  2, MAX_WIDTH);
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/881/

----
**testing_done**:  
**description**:  <p>Make sure the width/height is large enough in SetOptionEncParamExt</p><p>When using 4 downsampled layers, make sure the random width/height<br />is large enough that the downsampled layers don't have a zero size.</p>

----
 time 2014-10-30T00:31:14+00:00 
2014-10-30T00:31:14+00:00 dong zhang (dongzhang):
><p>sees good to me except the following comments, but it is OK.</p>
sees good to me except the following comments, but it is OK.
Ship it!

2014-10-30T00:31:14+00:00 dong zhang (dongzhang):
 on **test/api/encode_decode_api_test.cpp line1994**: 
 id="comment186049"><p>after the modification, the first value will always larger than 2, no need to use WELS_CLIP3.</p>
after the modification, the first value will always larger than 2, no need to use WELS_CLIP3.


----
2014-10-30T06:59:51+00:00 status-change:
Removed the WELS_CLIP3 as suggested


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/881/

PR #1486, opened 2014-11-03
Don't try to use pow(int,int)

mstorsjo:
When building for android or older MSVC versions, this makes the build
fail, when it is ambiguous which pow() overload to use, when there
is no function matching pow(int, int).

Review at https://rbcommons.com/s/OpenH264/r/894/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/894/

----
**testing_done**:  
**description**:  Don&#39;t try to use pow(int,int)When building for android or older MSVC versions, this makes the buildfail, when it is ambiguous which pow() overload to use, when thereis no function matching pow(int, int).

----
 reviewer wayne liu (huili2) 
 time 2014-11-03T09:34:24+00:00 
2014-11-03T09:34:24+00:00 wayne liu (huili2):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/894/

PR #1501, opened 2014-11-06
remove repeat initialization

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/902/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/902/

----
**testing_done**:  
**description**:  <p></p><p>sDependencyLayers and sSpatialLayers  are initialized in FillDefault (SEncParamExt&amp; param).<br /> then sDependencyLayers and sSpatialLayers are set 0 in FillDefault() again. which is an uncorrect operation</p>

----
2014-11-06T01:59:03-05:00 status-change:
fix a error UT case-- it is possible to have continuous IDR frames 


----
2014-11-06T03:04:08-05:00 status-change:
don't skip frame for the first IDR frame


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/902/

PR #1517, opened 2014-11-10
fix cabac encoding bug -- don't get the correct previous mb qp

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/913/

### DIFF
--- a/codec/encoder/core/src/svc_set_mb_syn_cabac.cpp
+++ b/codec/encoder/core/src/svc_set_mb_syn_cabac.cpp
@@ -224,12 +224,11 @@ void WelsCabacMbCbp (SMB* pCurMb, int32_t iMbWidth, SCabacCtx* pCabacCtx) {
   }
 }
 
-void WelsCabacMbDeltaQp (SMB* pCurMb, SCabacCtx* pCabacCtx, bool bPredMb) {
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/913/

----
**testing_done**:  
**description**:  <p>fix cabac encoding bug -- don't get the correct previous mb qp</p>

----
 reviewer Sijia Chen (sijchen) 
 time 2014-11-10T22:58:44+00:00 
2014-11-10T22:58:44+00:00 Sijia Chen (sijchen):
><p>fine with me after the suggested fix</p>
fine with me after the suggested fix
Ship it!

2014-11-10T22:58:44+00:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/svc_set_mb_syn_cabac.cpp line227**: 
 id="comment189773"><p>suggest a better naming of "bPredMb", this one is not self-explaining and a bit misleading, can we just use something like "bFirstMbInSlice"?</p>
suggest a better naming of "bPredMb", this one is not self-explaining and a bit misleading, can we just use something like "bFirstMbInSlice"?


----
2014-11-11T04:23:21+00:00 status-change:
rename


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/913/

PR #1547, opened 2014-11-24
parse only add and UT

huili2:
add ParseOnly for input bitstream without reconstruction.
This modifies API with "bParseOnly", which can not be modified during decoding.
It also enable some re-writing functions of single layer SVC header to AVC (under no inter-layer prediction) to some degree.

see:
https://rbcommons.com/s/OpenH264/r/910/

### DIFF
--- a/codec/api/svc/codec_app_def.h
+++ b/codec/api/svc/codec_app_def.h
@@ -457,6 +457,7 @@ typedef struct TagSVCDecodingParam {
   unsigned char	uiTargetDqLayer;       ///< setting target dq layer id
 
   ERROR_CON_IDC eEcActiveIdc;          ///< whether active error concealment feature in decoder
+  bool bParseOnly;                     ///< decoder for parse only, no reconstruction
 
   SVideoProperty   sVideoProperty;    ///< video stream property
 } SDecodingParam, *PDecodingParam;
@@ -574,7 +575,7 @@ typedef struct TagParserBsInfo {
   unsigned char* pDstBuff;                     ///< outputted dst buffer for parsed bitstream
   int iSpsWidthInPixel;                        ///< required SPS width info
   int iSpsHeightInPixel;                       ///< required SPS height info
-} SParserBsInfo, PParserBsInfo;
+} SParserBsInfo, *PParserBsInfo;
 
 /**
 * @brief Structure for encoder statistics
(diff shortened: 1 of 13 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/910/

----
**testing_done**:  
**description**:  <p>add parseonly code and UT.<br />add a safety guarantee in welsDecoderExt.cpp in Line 240.<br />This will modify current API.</p>

----
 time 2014-11-13T03:00:30+00:00 
2014-11-13T03:00:30+00:00 Juanny Wang (JuannyWang):
 on **codec/decoder/core/inc/decoder_context.h line115**: 
 id="comment190730"><p>Should sps_id of current pps be added? Which can detect different sps interval effectively for consecutive sps/pps loss.</p>
Should sps_id of current pps be added? Which can detect different sps interval effectively for consecutive sps/pps loss.


----
 time 2014-11-13T03:16:40+00:00 
2014-11-13T03:16:40+00:00 Juanny Wang (JuannyWang):
 on **codec/decoder/core/src/au_parser.cpp line1030**: 
 id="comment190732"><p>Just direct memcpy may be inappropriate in here I think. For subset sps, profile_id is 83 or 86, which is different from profile_id of normal sps (7). Rewriting may be necessary if want to avoid the mismatch between profile_id and sps type (7 or 15). Of course, based on DX80 project experience, HW decoder still can work normally if given the profile_id (83 or 86) but sps type(7).</p>
Just direct memcpy may be inappropriate in here I think. For subset sps, profile_id is 83 or 86, which is different from profile_id of normal sps (7). Rewriting may be necessary if want to avoid the mismatch between profile_id and sps type (7 or 15). Of course, based on DX80 project experience, HW decoder still can work normally if given the profile_id (83 or 86) but sps type(7).


----
2014-11-19T06:12:38+00:00 status-change:
Add re-write for subsetSPS into SPS, with default MAIN profile.<br />And add related UVLC writing functions as in encoder, with Decxxx as prefix function name.


----
 time 2014-11-21T05:57:12+00:00 
2014-11-21T05:57:12+00:00 Juanny Wang (JuannyWang):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/910/

PR #1550, opened 2014-11-25
add maxbitrate verification

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/928/

### DIFF
--- a/test/encoder/EncUT_EncoderExt.cpp
+++ b/test/encoder/EncUT_EncoderExt.cpp
@@ -208,45 +208,50 @@ TEST_F (EncoderInterfaceTest, EncoderAdditionalOptionSetTest) {
   iResult = pPtrEnc->SetOption (eOptionId, &sInfo);
   if (sInfo.iBitrate <= 0)
     EXPECT_EQ (iResult, static_cast<int> (cmInitParaError));
-  else {
-    EXPECT_EQ (iResult, static_cast<int> (cmResultSuccess));
-    sReturn.iLayer = static_cast<LAYER_NUM> (pParamExt->iSpatialLayerNum);
-    iResult = pPtrEnc->GetOption (eOptionId, &sReturn);
-    EXPECT_EQ (iResult, static_cast<int> (cmResultSuccess));
-    EXPECT_EQ (WELS_CLIP3 (sInfo.iBitrate, 1, 2147483647), sReturn.iBitrate);
-  }
-  PrepareOneSrcFrame();
-  iResult = pPtrEnc->EncodeFrame (pSrcPic, &sFbi);
-  EXPECT_EQ (iResult, static_cast<int> (cmResultSuccess));
-  pSrcPic->uiTimeStamp += 30;
-
-  eOptionId = ENCODER_OPTION_MAX_BITRATE;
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/928/

----
**testing_done**:  
**description**:  <p>update maxbitrate checking<br />1. use 0 as the default value for maxbitrate<br />2. if the app doesn't set this value, get this from the level limiation table<br />3. if the maxbitrate set by app is larger than the level limiation, it is considered as a error value.</p>

----
2014-11-20T01:48:04+00:00 status-change:
use UNSPECIFIED_BIT_RATE instead of 0


----
 reviewer Sijia Chen (sijchen) 
 time 2014-11-21T21:59:47+00:00 
2014-11-21T22:00:25+00:00 Sijia Chen (sijchen):
><p>I am okay with 1/2 in the list but not on <br />3:"if the maxbitrate set by app is larger than the level limiation, it is considered as a error value."</p>
I am okay with 1/2 in the list but not on <br />3:"if the maxbitrate set by app is larger than the level limiation, it is considered as a error value."

2014-11-21T22:00:25+00:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/encoder_ext.cpp line169**: 
 id="comment193377"><p>i think we had agreement on this, similar to SDP logic, the max-br can be expanded and encoder will adjust level_idx accordingly under the REAL_TIME usage type.</p>
i think we had agreement on this, similar to SDP logic, the max-br can be expanded and encoder will adjust level_idx accordingly under the REAL_TIME usage type.


----
2014-11-24T06:58:03+00:00 status-change:
change #3: if the maxbitrate set by app is larger than the level limiation, the level is changed


----
 reviewer Sijia Chen (sijchen) 
 time 2014-11-24T19:27:08+00:00 
2014-11-24T19:27:08+00:00 Sijia Chen (sijchen):
Ship it!

2014-11-24T19:27:08+00:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/encoder_ext.cpp line527**: 
 id="comment193840"><p>should we return error at this setoption if the bit rate is larger as your logic here? so that we can add a UT to check it.</p>
should we return error at this setoption if the bit rate is larger as your logic here? so that we can add a UT to check it.

2014-11-24T19:27:08+00:00 Sijia Chen (sijchen):
 on **test/encoder/EncUT_EncoderExt.cpp line441**: 
 id="comment193844"><p>could you add another tests or put it into our to-do list:<br />1, a setting which BR is higher than LEVEL_IDC's setting so that it will hit the logic of increasing LEVEL_IDC<br />2, a setting which BR will meet the condition of above "ENC_RETURN_UNSUPPORTED_PARA", and see if it can returns failure?</p>
could you add another tests or put it into our to-do list:<br />1, a setting which BR is higher than LEVEL_IDC's setting so that it will hit the logic of increasing LEVEL_IDC<br />2, a setting which BR will meet the condition of above "ENC_RETURN_UNSUPPORTED_PARA", and see if it can returns failure?

2014-11-25T03:09:06+00:00 Karina Li (ruil2):
><p>will do in the next modification</p>
will do in the next modification


----

----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/928/

PR #1557, opened 2014-11-26
add timestamp in decoder

dongzha:
review at:
https://rbcommons.com/s/OpenH264/r/936/

### DIFF
--- a/codec/api/svc/codec_app_def.h
+++ b/codec/api/svc/codec_app_def.h
@@ -577,6 +577,8 @@ typedef struct TagParserBsInfo {
   unsigned char* pDstBuff;                     ///< outputted dst buffer for parsed bitstream
   int iSpsWidthInPixel;                        ///< required SPS width info
   int iSpsHeightInPixel;                       ///< required SPS height info
+  unsigned long long uiInBsTimeStamp;               ///< input BS timestamp
+  unsigned long long uiOutBsTimeStamp;             ///< output BS timestamp
 } SParserBsInfo, *PParserBsInfo;
 
 /**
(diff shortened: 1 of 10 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/936/

----
**testing_done**:  
**description**:  <p>add timestamp in decoder</p>

----
 time 2014-11-25T07:52:14+00:00 
2014-11-25T07:52:14+00:00 Martin Storsj (mstorsjo):
 on **codec/api/svc/codec_app_def.h line579**: 
 id="comment194091"><p>Should these perhaps be unsigned long long instead, to make sure they are 64 bit even on 32 bit platforms? Same for a number of cases further down as well.</p>
Should these perhaps be unsigned long long instead, to make sure they are 64 bit even on 32 bit platforms? Same for a number of cases further down as well.

2014-11-26T01:52:31+00:00 dong zhang (dongzhang):
><p>Yes, its better to use unsigned long long, although all the modules among the transmission path uses 32 bits to represent timestamp.<br />Will fix it.</p>
Yes, its better to use unsigned long long, although all the modules among the transmission path uses 32 bits to represent timestamp.<br />Will fix it.


----
2014-11-26T02:06:18+00:00 status-change:
use unsigned long long


----
2014-11-26T02:59:27+00:00 status-change:
fix bug


----
 reviewer wayne liu (huili2) 
 time 2014-11-26T03:13:13+00:00 
2014-11-26T03:13:13+00:00 wayne liu (huili2):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/936/

PR #1572, opened 2014-11-28
Fixes relating to the loopfilter parameter

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/949/.

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -261,6 +261,12 @@ int32_t ParamValidationExt (SLogContext* pLogCtx, SWelsSvcCodingParam* pCodingPa
              pCodingParam->uiIntraPeriod, pCodingParam->uiGopSize);
     return ENC_RETURN_UNSUPPORTED_PARA;
   }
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/949/

----
**testing_done**:  
**description**:  <p>Validate that iLoopFilterDisableIdc is set to a sensible value in ParamValidationExt</p><p>In WelsEncoderParamAdjust it is already clipped in the<br />same way, but here we should probably rather error out<br />than to silently clip it to the right range.</p><p></p><p>Log what uiDisableDeblockingFilterIdc was set to when printing an error message</p><p></p><p>Don't set invalid iLoopFilterDisableIdc in RandomParamExtCombination</p>

----
 reviewer Sijia Chen (sijchen) 
 time 2014-11-28T18:27:05+00:00 
2014-11-28T18:27:05+00:00 Sijia Chen (sijchen):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/949/

PR #1582, opened 2014-12-01
Fix/clean up building and installing shared libraries

mstorsjo:
Please see each individual commit for an explanation on what it does and what issue it fixes.

Review at https://rbcommons.com/s/OpenH264/r/955/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/955/

----
**testing_done**:  
**description**:  Set a default SHAREDLIB_DIR in the main MakefileOnly override it in the ones where the default isn&#39;t desired.Simplify creating output directories for the install-shared ruleActually install the shared library into $(SHAREDLIB_DIR) as intendedSet SHAREDLIB_DIR for msvc as wellHandle generating the versionless library via make rulesThis makes sure both &quot;make libopenh264.so&quot; and &quot;make libopenh264.so.0&quot;work as intended.Use the versionless name as dependency for the install rule,to make sure both (if possible) are created.Check the suffix instead of checking whether a file is a linkThis feels more straightforward - some file systems/environments(such as MSys) implement symlinks as plain file copies, wheredistinguishing whether a file is a link might be hard.Actually install the versioned shared library file as wellSimplify setting the shlib flags for msvcFix the dependencies of the install-shared targetPreviously calling &quot;make install&quot; failed unless the targetactually already was built, for platforms where the sharedlibraries are versioned.Honor the silent build flag when adding the shared library symlinkSet the right SHAREDLIBSUFFIXVER for msvcInstall the import library via the EXTRA_LIB mechanism on mingw as wellPrior to 3f69873c99, there already was a mechanism for installingthe import library, used for the MSVC builds. Use that for mingwas well, instead of adding more hardcoded rules.

----
 time 2014-12-02T19:46:05+00:00 
2014-12-02T19:46:05+00:00 Ethan Hugg (ethanhugg):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/955/

PR #1590, opened 2014-12-04
add output trace for parameter setting

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/963/

### DIFF
--- a/codec/encoder/plus/src/welsEncoderExt.cpp
+++ b/codec/encoder/plus/src/welsEncoderExt.cpp
@@ -166,8 +166,6 @@ void CWelsH264SVCEncoder::InitEncoder (void) {
   if (m_pWelsTrace == NULL) {
     return;
   }
-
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/963/

----
**testing_done**:  
**description**:  <p>we also need to output parameter trace to verify whether they are what we want to be,though these parmaters passsed the verification.</p>

----
2014-12-04T06:09:49+00:00 status-change:
add output trace for parameter setting
we also need to output parameter trace to verify whether they are what we want to be,though these parmaters passsed the verification.


----
2014-12-04T09:13:22+00:00 status-change:
we should not set WELS_LOG_ERROR level in encoder initilization. <br />1. after setting WELS_LOG_ERROR level, some important info can't be outputed.such as parameters info, version info <br />2. the level should be changed by app


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/963/

PR #1595, opened 2014-12-05
should not set WELS_LOG_ERROR level in encoder initilization.

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/965/

### DIFF
--- a/codec/api/svc/codec_app_def.h
+++ b/codec/api/svc/codec_app_def.h
@@ -328,7 +328,7 @@ enum {
   WELS_LOG_DETAIL      = 1 << 4,        ///< per packet/frame log
   WELS_LOG_RESV        = 1 << 5,        ///< resversed log iLevel
   WELS_LOG_LEVEL_COUNT = 6,
-  WELS_LOG_DEFAULT     = WELS_LOG_DEBUG	///< default log iLevel in Wels codec
+  WELS_LOG_DEFAULT     = WELS_LOG_INFO	///< default log iLevel in Wels codec
 };
 
 /**


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/965/

----
**testing_done**:  
**description**:  <p>we should not set WELS_LOG_ERROR level in encoder initilization. </p><p>after setting WELS_LOG_ERROR level, some important info can't be outputed.such as parameters info, version info<br />the level should be changed by app</p>

----
2014-12-05T09:04:51+00:00 status-change:
fix UT bug<br />1. TestOutPutTrace checks the level to judge whether the output is a correct output. when the default level isn't same with external setting level, will return error. so should set the level firstly before trace output.


----
2014-12-05T09:20:28+00:00 status-change:
change default level as info level


----
 reviewer Sijia Chen (sijchen) 
 time 2014-12-08T03:00:53+00:00 
2014-12-08T03:00:53+00:00 Sijia Chen (sijchen):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----
 time 2014-12-10T12:45:09+00:00 
2014-12-10T12:45:09+00:00 Martin Storsj (mstorsjo):
><p>I don't understand why the default level should be changed - I preferred having ERROR as the default level, to keep things silent normally as long as nothing goes wrong. Previously there was a bug that made it impossible to show the encoder version though, but I've fixed that in https://rbcommons.com/s/OpenH264/r/978/.</p><p>But as long as the default is something else than ERROR, the unit test suite prints a lot of unnecessary text, making it much harder to see the important things. And even if we fix the testsuite to set the log level back to ERROR, it's quite annoying to have to change this in every single application that uses the code.</p>
I don't understand why the default level should be changed - I preferred having ERROR as the default level, to keep things silent normally as long as nothing goes wrong. Previously there was a bug that made it impossible to show the encoder version though, but I've fixed that in https://rbcommons.com/s/OpenH264/r/978/.
But as long as the default is something else than ERROR, the unit test suite prints a lot of unnecessary text, making it much harder to see the important things. And even if we fix the testsuite to set the log level back to ERROR, it's quite annoying to have to change this in every single application that uses the code.

2014-12-11T01:52:38+00:00 Karina Li (ruil2):
><ol><li>for encoder, error info is the most important,but the parameter info is also important. In the most of time, though there is no error info, it doesn't mean the parameters is correct. we need to verify these parameters.</li><li>for UT, we can set the error level.</li></ol>


----

Finish archiving from https://rbcommons.com/s/OpenH264/r/965/

Issue #1529, opened 2014-11-13
use openh264 in ffmpeg


Arthur-wang:
how to use open264 in ffmpeg?

mstorsjo:
libav and ffmpeg (in the latest git versions) now support linking to OpenH264.

sijchen:
thank you very much!

BarkyTheDog:
I'd like to try using OpenH264 with ffmpeg...is there any specific info on how this can be done (a few lines of example code, what steps are necessary, etc), before I have to resort to digging through source code?  :-) Thanks! Apologies if I'm just missing something obvious here...

mstorsjo:
Build and install openh264 (`make install PREFIX=/some/path`). Configure ffmpeg (`configure --enable-libopenh264`), make sure that `pkg-config` is available and looks in the right directory (if `/some/path` is a default path like `/usr` or `/usr/local` it may be picked up by default, otherwise set `export PKG_CONFIG_PATH=/some/path/lib/pkgconfig`). For other details with `pkg-config` in e.g. cross-building setups, refer to ffmpeg.

BarkyTheDog:
Thanks much!

mstorsjo:
If you've only built/installed a static library, it will compile it in, but if you built a .so/.dll, it will only link to it, not compile it in. (For MSVC builds, replace <prefix>/lib/openh264.lib with <prefix>/lib/openh264_dll.lib, to have it link to it as a DLL instead.) Then if you leave out the openh264 shared library, you can replace it with one downloaded at runtime (just make sure you've built it from the same release branch as the binary you're using.)

A libavcodec library built this way won't load or run at all until you actually provide the right openh264.so (or provide a fake dummy one which doesn't contain any implementations) though. Support for loading it at runtime via dlopen isn't available (yet at least).

BarkyTheDog:
OK. That makes sense. So, the openh264 shared library would simply need to be located where the runtime linker can find it. Support for dlopen, etc. would be a wee bit more convenient, but not a big deal at this point :-) Thanks!

BarkyTheDog:
I'm attempting to make use of openh264 with FFmpeg, but for now am not dealing with the "downloaded shared library" issue -- I'm just compiling both from their latest git repositories.

[...]

PR #1603, opened 2014-12-10
add generating script and the first version of codec_ver.h

sijchen:
being reviewed at: https://rbcommons.com/s/OpenH264/r/956/
for https://github.com/cisco/openh264/issues/1529

### DIFF
--- a/codec/api/svc/codec_ver.h
+++ b/codec/api/svc/codec_ver.h
@@ -4,13 +4,12 @@
 
 #include "codec_app_def.h"
 
-static const Openh264Version g_stCodecVersion  = {1, 2, 1, 0};
-static const char* g_strCodecVer  = "Openh264 version:1.2.1.0";
-static const char* g_strCodecBuildNum = "Openh264 revision:a2c47f2";
+static const Openh264Version g_stCodecVersion  = {1,3,0,0};
+static const char* g_strCodecVer  = "Openh264 version:1.3.0.0";
 
 #define OPENH264_MAJOR (1)
-#define OPENH264_MINOR (2)
-#define OPENH264_REVISION (1)
+#define OPENH264_MINOR (3)
+#define OPENH264_REVISION (0)
 #define OPENH264_RESERVED (0)
 
 #endif  // CODEC_VER_H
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/956/

----
**testing_done**:  
**description**:  <p>add generating script and the first version of codec_ver.h</p><p>can this fit the requirement raised in https://github.com/cisco/openh264/issues/1529 ?</p>

----

----
 time 2014-12-01T20:07:53+00:00 
2014-12-01T20:07:53+00:00 Martin Storsj (mstorsjo):
 on **codec/api/svc/codec_ver.h line4**: 
 id="comment195301"><p>Please have it generate the version as a #define macro as well.</p><p>Either something like "#define OPENH264_VERSION (1 &lt;&lt; 24) | (2 &lt;&lt; 16) | (1 &lt;&lt; 8) | (0 &lt;&lt; 0)" with the full version as one number, or "#define OPENH264_MAJOR 1, #define OPENH264_MINOR 2 #define OPENH264_RELEASE 1 #define OPENH264_PATCHLEVEL 0" as split up into separate values. This allows doing "#if OPENH264_VERSION &lt; (SOME_NUMBER) ... #else ... #endif".</p>
Please have it generate the version as a #define macro as well.
Either something like "#define OPENH264_VERSION (1 &lt;&lt; 24) | (2 &lt;&lt; 16) | (1 &lt;&lt; 8) | (0 &lt;&lt; 0)" with the full version as one number, or "#define OPENH264_MAJOR 1, #define OPENH264_MINOR 2 #define OPENH264_RELEASE 1 #define OPENH264_PATCHLEVEL 0" as split up into separate values. This allows doing "#if OPENH264_VERSION &lt; (SOME_NUMBER) ... #else ... #endif".

2014-12-01T20:07:53+00:00 Martin Storsj (mstorsjo):
 on **codec/build/generate_codec_ver.sh line25**: 
 id="comment195302"><p>I agree here that it's better to not make this use C++ namespaces, to allow the same to be used from both C and C++.</p>
I agree here that it's better to not make this use C++ namespaces, to allow the same to be used from both C and C++.


----
2014-12-02T19:05:22+00:00 status-change:
support sectioning the version


----
 time 2014-12-16T10:06:06+00:00 
2014-12-16T10:06:06+00:00 Martin Storsj (mstorsjo):
><p>This looks mostly good, but I think we should skip the git version hash in the header. Since the header isn't updated automatically all the time, the hash version will not be correct in most cases anyway.</p>
This looks mostly good, but I think we should skip the git version hash in the header. Since the header isn't updated automatically all the time, the hash version will not be correct in most cases anyway.


----

----
 time 2014-12-18T09:21:42+00:00 
2014-12-18T09:21:42+00:00 Martin Storsj (mstorsjo):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/956/

PR #1632, opened 2014-12-16
use the same complexity computation branch for screen and camera

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/993/

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -3263,11 +3263,10 @@ int32_t WelsEncoderEncodeExt (sWelsEncCtx* pCtx, SFrameBSInfo* pFbi, const SSour
 #ifdef LONG_TERM_REF_DUMP
     DumpRef (pCtx);
 #endif
-    if ((pSvcParam->iUsageType == SCREEN_CONTENT_REAL_TIME) && (pSvcParam->iRCMode != RC_OFF_MODE))
+    if ((pSvcParam->iRCMode != RC_OFF_MODE))
       pCtx->pVpp->AnalyzePictureComplexity (pCtx, pCtx->pEncPic, ((pCtx->eSliceType == P_SLICE)
                                             && (pCtx->iNumRef0 > 0)) ? pCtx->pRefList0[0] : NULL,
-                                            iCurDid, pSvcParam->bEnableBackgroundDetection);
-
+                                            iCurDid,(pCtx->eSliceType == P_SLICE)&&(pSvcParam->bEnableBackgroundDetection));
     WelsUpdateRefSyntax (pCtx,  pCtx->iPOC,
                          eFrameType);	//get reordering syntax used for writing slice header and transmit to encoder.
     PrefetchReferencePicture (pCtx, eFrameType);	// update reference picture for current pDq layer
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/993/

----
**testing_done**:  
**description**:  <ol><li>use the same complexity computation branch for screen and camera,put this computation after build reference list because this computation needs the correct reference frame.</li><li>clean up some redundent judgement</li><li>add some protection</li></ol>

----
 reviewer Sijia Chen (sijchen) 
 time 2014-12-16T03:57:49+00:00 
2014-12-16T03:57:49+00:00 Sijia Chen (sijchen):
><p>I am not sure the following two are the same which can be combined, do we have experimental results on this switch of algo?</p>
I am not sure the following two are the same which can be combined, do we have experimental results on this switch of algo?

2014-12-16T04:31:54+00:00 Karina Li (ruil2):
><p>camera path should be error operation from my understanding.</p>
camera path should be error operation from my understanding.

2014-12-16T04:31:54+00:00 Karina Li (ruil2):
 on **codec/encoder/core/src/encoder_ext.cpp line3267**: 
 id="comment199692"><p>here it is using the <em>reconstructed</em> ref to compare with current original picture</p>
here it is using the <em>reconstructed</em> ref to compare with current original picture

2014-12-16T05:32:34+00:00 Karina Li (ruil2):
><p>to which line does this comment goes to?</p>
the correct operation is that the ref should be from the reference pictures. the complexity is used to estimate the bits, so the complexity calcualtion should be same with the encoding processing.
to which line does this comment goes to?

2014-12-16T05:32:34+00:00 Karina Li (ruil2):
 on **codec/encoder/core/src/wels_preprocess.cpp line253**: 
 id="comment199693"><p>the picture in this memory is the <em>original</em> of the last frame, which may probably not in the ref list</p>
the picture in this memory is the <em>original</em> of the last frame, which may probably not in the ref list

2014-12-16T07:44:26+00:00 Karina Li (ruil2):
><p>yes, before modification, the ref is from original picture list. but it is not reasonable desigin. I think this is error logic. compelixty should use the ref from the reference pictures, which will get the more precise result.</p>
yes, before modification, the ref is from original picture list. but it is not reasonable desigin. I think this is error logic. compelixty should use the ref from the reference pictures, which will get the more precise result.


----
 reviewer Sijia Chen (sijchen) 
 time 2014-12-18T02:27:23+00:00 
2014-12-18T02:27:23+00:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/encoder_ext.cpp line3269**: 
 id="comment200272"><p>why delete this PSLICE condition for BGD?</p>
why delete this PSLICE condition for BGD?

2014-12-18T02:30:55+00:00 Karina Li (ruil2):
><p>this means that only P frame needs to do BGD</p>
this means that only P frame needs to do BGD


----
 reviewer Sijia Chen (sijchen) 
 time 2014-12-18T03:07:38+00:00 
2014-12-18T03:07:38+00:00 Sijia Chen (sijchen):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/993/

PR #1634, opened 2014-12-16
use WELS_LOG_ERROR as default trace level in console app

ruil2:
reviewed at:https://rbcommons.com/s/OpenH264/r/995/

### DIFF
--- a/codec/console/enc/src/welsenc.cpp
+++ b/codec/console/enc/src/welsenc.cpp
@@ -109,7 +109,7 @@ static int     g_iCtrlC = 0;
 static void    SigIntHandler (int a) {
   g_iCtrlC = 1;
 }
-static int     g_LevelSetting = -1;
+static int     g_LevelSetting = WELS_LOG_ERROR;
 
 int ParseLayerConfig (CReadConfig& cRdLayerCfg, const int iLayer, SEncParamExt& pSvcParam, SFilesSet& sFileSet) {
   if (!cRdLayerCfg.ExistFile()) {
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/995/

----
**testing_done**:  
**description**:  use WELS_LOG_ERROR as default trace level in console app

----
 reviewer Sijia Chen (sijchen) 
 time 2014-12-16T06:27:20+01:00 
2014-12-16T06:27:20+01:00 Sijia Chen (sijchen):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----
 time 2014-12-16T08:08:05+01:00 
2014-12-16T08:08:05+01:00 Martin Storsj (mstorsjo):
><p>Since we want to have WELS_LOG_ERROR as the default everywhere (both in the console app and in the unit tests), why not just have it as the default in the library as well? Right now every single user will have to add code to change it back to WELS_LOG_ERROR.</p>
Since we want to have WELS_LOG_ERROR as the default everywhere (both in the console app and in the unit tests), why not just have it as the default in the library as well? Right now every single user will have to add code to change it back to WELS_LOG_ERROR.

2014-12-16T08:28:52+01:00 Karina Li (ruil2):
><p>thanks for this suggestion, I will use WELS_LOG_WARNING as the default level in library.</p>
thanks for this suggestion, I will use WELS_LOG_WARNING as the default level in library.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/995/

PR #1640, opened 2014-12-16
Fix occasional failures in some tests

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/1001/.

### DIFF
--- a/test/api/encode_decode_api_test.cpp
+++ b/test/api/encode_decode_api_test.cpp
@@ -2057,11 +2057,11 @@ TEST_F (EncodeDecodeTestAPI, Engine_SVC_Switch_P) {
 }
 
 TEST_F (EncodeDecodeTestAPI, SetOptionEncParamExt) {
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1001/

----
**testing_done**:  
**description**:  <p>Increase the minimum frame size in a test with multiple spatial layers</p><p>Make sure the input frame size is large enough that the downsampled<br />layers also have a nonzero resolution.</p><p></p><p>Use an inline function instead of a macro for clipping which contains rand()</p><p>When using a macro, the macro parameters get evaluated<br />multiple times, which means that the rand() value compared<br />actually isn't the same that is used as return value.</p><p>This makes sure that clipping works as intended for the<br />random tests.</p><p></p><p>Use the common WELS_CLIP3 macro instead of a custom reimplementation</p>

----
 reviewer Karina Li (ruil2) 
 time 2014-12-17T01:29:15+00:00 
2014-12-17T01:29:15+00:00 Karina Li (ruil2):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1001/

PR #1641, opened 2014-12-16
Make sure the random test bitrate is high enough

mstorsjo:
The ForceIntraFrame test will fail (giving up after 100 skipped
frames) if the bitrate is not set high enough.

Set the minimum bitrate to w*h/50, which is a very low value,
but which still should allow getting a non-skipped frame within
a few attempts.

Review at https://rbcommons.com/s/OpenH264/r/1002/.

### DIFF
--- a/test/encoder/EncUT_EncoderExt.cpp
+++ b/test/encoder/EncUT_EncoderExt.cpp
@@ -455,6 +455,8 @@ void GetValidEncParamBase (SEncParamBase* pEncParamBase) {
   pEncParamBase->iPicWidth = VALID_SIZE (pEncParamBase->iPicWidth);
   pEncParamBase->iPicHeight = VALID_SIZE (pEncParamBase->iPicHeight);
   pEncParamBase->iTargetBitrate = rand() + 1; //!=0
+  // Force a bitrate of at least w*h/50, otherwise we will only get skipped frames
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1002/

----
**testing_done**:  
**description**:  Make sure the random test bitrate is high enoughThe ForceIntraFrame test will fail (giving up after 100 skippedframes) if the bitrate is not set high enough.Set the minimum bitrate to w*h/50, which is a very low value,but which still should allow getting a non-skipped frame withina few attempts.

----
 time 2014-12-17T07:28:12+00:00 
2014-12-17T07:28:12+00:00 Martin Storsj (mstorsjo):
><p>Any comments on this one? This fixes test failures if the randomly chosen bitrate is too low (so that the encoder outputs more than 100 skipped frames, when trying to force an intra frame).</p>
Any comments on this one? This fixes test failures if the randomly chosen bitrate is too low (so that the encoder outputs more than 100 skipped frames, when trying to force an intra frame).


----
 reviewer Sijia Chen (sijchen) 
 time 2014-12-17T07:49:57+00:00 
2014-12-17T07:49:57+00:00 Sijia Chen (sijchen):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----
 reviewer Forrest Shi (huade) 
 time 2014-12-17T07:50:14+00:00 
2014-12-17T07:50:14+00:00 Forrest Shi (huade):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1002/

PR #1651, opened 2014-12-18
Improve version header

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/1012/.

### DIFF
--- a/codec/api/svc/codec_api.h
+++ b/codec/api/svc/codec_api.h
@@ -531,6 +531,11 @@ long WelsCreateDecoder (ISVCDecoder** ppDecoder);
 */
 void WelsDestroyDecoder (ISVCDecoder* pDecoder);
 
+/** @brief   Get codec version
+ *  @return  The linked codec version
+*/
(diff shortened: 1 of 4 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1012/

----
**testing_done**:  
**description**:  Add a public function for getting the codec versionIf the codec library is loaded dynamically, it may be necessary thatthe loaded library actually matches the version that was used whenit was built, to avoid issues due to ABI differences. Therefore adda public function to allow getting the version of the currentlyloaded library.Captizalize OpenH264 correctly in the version header

----
 reviewer Sijia Chen (sijchen) 
 time 2014-12-19T01:32:46+00:00 
2014-12-19T01:32:46+00:00 Sijia Chen (sijchen):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1012/

PR #1652, opened 2014-12-19
use picture-level new seq begin detection for output functions

huili2:
add picture-level new-seq-begin for correctly output log information.
see:
https://rbcommons.com/s/OpenH264/r/1011/

### DIFF
--- a/codec/decoder/core/inc/picture.h
+++ b/codec/decoder/core/inc/picture.h
@@ -78,6 +78,7 @@ int32_t		iLongTermFrameIdx;					//id for long term ref pic
 int32_t     iSpsId; //against mosaic caused by cross-IDR interval reference.
 int32_t     iPpsId;
 unsigned long long uiTimeStamp;
+bool bNewSeqBegin;
 int32_t iMbEcedNum;
 int32_t iMbEcedPropNum;
 int32_t iMbNum;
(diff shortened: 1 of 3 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1011/

----
**testing_done**:  
**description**:  <p>current output function will errorly judge new seq begin with the following case:<br />1. SPS/PPS arrives, but IDR is lost<br />2. only part of P picture is received<br />In this case, after the incomplete P picture is reconstructed, pCtx-&gt;bNewSeqBegin was set to false. Then after EC, in output function DecodeFrameConstruction(), it was still not the seq begin, leading to error log information.</p><p>Add picture level new seq begin flag, to correct and give the proper information in DecodeFrameConstruction().</p>

----
 time 2014-12-18T07:30:53+00:00 
2014-12-18T07:30:53+00:00 dong zhang (dongzhang):
 on **codec/decoder/core/src/decoder_core.cpp line1934**: 
 id="comment200283"><p>pCtx-&gt;pDec-&gt;bNewSeqBegin = pCtx-&gt;bNewSeqBegin; should not be added here, consider the case N slices input N times. Each time DecodeFrame (NULL) will set pCtx-&gt;bNewSeqBegin to false,<br />so pCtx-&gt;pDec-&gt;bNewSeqBegin will become false from the second to N th slice.</p>
pCtx-&gt;pDec-&gt;bNewSeqBegin = pCtx-&gt;bNewSeqBegin; should not be added here, consider the case N slices input N times. Each time DecodeFrame (NULL) will set pCtx-&gt;bNewSeqBegin to false,<br />so pCtx-&gt;pDec-&gt;bNewSeqBegin will become false from the second to N th slice.


----
2014-12-18T09:24:51+00:00 status-change:
make pDec always NULL after decoding picture finished.


----
 time 2014-12-18T09:28:23+00:00 
2014-12-18T09:28:23+00:00 dong zhang (dongzhang):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1011/

PR #1666, opened 2014-12-23
fix a spatial layer index error

lyao2:
review at:
https://rbcommons.com/s/OpenH264/r/1020/

### DIFF
--- a/test/encoder/EncUT_EncoderExt.cpp
+++ b/test/encoder/EncUT_EncoderExt.cpp
@@ -210,7 +210,7 @@ TEST_F (EncoderInterfaceTest, EncoderAdditionalOptionSetTest) {
   eOptionId = ENCODER_OPTION_BITRATE;
   SBitrateInfo sInfo, sReturn;
   sInfo.iBitrate = rand() % 100000 - 100;
-  sInfo.iLayer = static_cast<LAYER_NUM> (pParamExt->iSpatialLayerNum);
+  sInfo.iLayer = static_cast<LAYER_NUM> (pParamExt->iSpatialLayerNum - 1);
   iResult = pPtrEnc->SetOption (eOptionId, &sInfo);
   pPtrEnc->GetOption (ENCODER_OPTION_SVC_ENCODE_PARAM_EXT, pParamExt);
   if (sInfo.iBitrate <= 0)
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1020/

----
**testing_done**:  
**description**:  <p>fix a spatial layer index error</p>

----
 reviewer Sijia Chen (sijchen) 
 time 2014-12-23T03:18:01+00:00 
2014-12-23T03:18:01+00:00 Sijia Chen (sijchen):
><p>@Karina please double-check</p>
@Karina please double-check
Ship it!


----
 reviewer Karina Li (ruil2) 
 time 2014-12-23T05:24:50+00:00 
2014-12-23T05:24:50+00:00 Karina Li (ruil2):
 on **test/encoder/EncUT_EncoderExt.cpp line213**: 
 id="comment201213"><p>this is different concepts between sInfo.iLayer and iSpatialLayerNum.</p><ol><li>sInfo.iLayer means that the number of current spatial layer.</li><li>iSpatialLayerNum means that the total number of spatial layers.</li></ol><p>it is not correct assigment.</p>
this is different concepts between sInfo.iLayer and iSpatialLayerNum.
it is not correct assigment.

2014-12-23T06:01:48+00:00 Li Yao (lyao2):
><p>yes, I know that old code isn't correct. though current modification can't cause other issue, it will cause misunderstanding on this.<br />if you know there is only one layer, you should just use sInfo.iLayer = SPATIAL_LAYER_0</p>
in this test case, in InitializeParamExt pParamExt-&gt;iSpatialLayerNum = 1; there is only one spatial layer.<br />so the sInfo.iLayer should be 0<br />it is not correct use sInfo.iLayer = static_cast&lt;LAYER_NUM&gt; (pParamExt-&gt;iSpatialLayerNum);
yes, I know that old code isn't correct. though current modification can't cause other issue, it will cause misunderstanding on this.<br />if you know there is only one layer, you should just use sInfo.iLayer = SPATIAL_LAYER_0


----

----

----
 reviewer Karina Li (ruil2) 
 time 2014-12-23T09:19:43+00:00 
2014-12-23T09:19:43+00:00 Karina Li (ruil2):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1020/

PR #1668, opened 2014-12-23
Fix build warnings with GCC

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/1021/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1021/

----
**testing_done**:  
**description**:  Initialize a local variableThis fixes warnings with GCC about a variable that may be useduninitialized (if iResProperty was an incorrect value).Use an unsigned variable for comparison with unsigned valuesThis avoids warnings about comparison between signed and unsigned.Remove a write-only variableThis fixes warnings with GCC about a variable that is set butnot used.

----
 reviewer ling zhu (lingzhu) 
 time 2014-12-24T00:50:30+00:00 
2014-12-24T00:50:30+00:00 ling zhu (lingzhu):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1021/

PR #1691, opened 2015-01-02
Improve pkg-config files for use with static libraries

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/1038/.


sijchen:
@zaheerm

zaheerm:
This is only positive change, thanks for the contribution. Looks good for merging.

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1038/

----
**testing_done**:  
**description**:  Include private dependencies in the Libs line in pkg-config if only installing a static libraryIf only a static library is installed, the user of the library maynot know that only a static version of this particular library isavailable (and doesn&#39;t know that --static should be added to thepkg-config call). Therefore, one common practice is to include privatedependencies in the public Libs line if a static-only library isinstalled.Include -lstdc++ under Libs.private in the pkg-config fileThis allows linking statically to the library, by passing --staticto pkg-config.Don&#39;t install the pkg-config file as readonlyThis file should be installed just as any other file, protected bynormal user permissions, not by making the file readonly.Simplify the makefile rule for generating the pkg-config file

----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1038/

PR #1692, opened 2015-01-02
Add a new public function WelsGetCodecVersionEx

mstorsjo:
This function doesn't return a struct, but fills in a struct via
a provided pointer.

The ABI of returning a struct is different between MSVC and mingw.

This allows using the same function from mingw, even though the DLL
has been built with MSVC.

Review at https://rbcommons.com/s/OpenH264/r/1039/.

### DIFF
--- a/codec/api/svc/codec_api.h
+++ b/codec/api/svc/codec_api.h
@@ -532,9 +532,15 @@ long WelsCreateDecoder (ISVCDecoder** ppDecoder);
 void WelsDestroyDecoder (ISVCDecoder* pDecoder);
 
 /** @brief   Get codec version
+ *           Note, this function isn't ABI compatible between MSVC and Mingw.
  *  @return  The linked codec version
 */
 OpenH264Version WelsGetCodecVersion ();
+
+/** @brief   Get codec version
+ *  @param   pVersion  struct to fill in with the version
+*/
+void WelsGetCodecVersionEx (OpenH264Version *pVersion);
 
 #ifdef __cplusplus
 }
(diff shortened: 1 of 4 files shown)


mstorsjo:
Actually, this seems to only be a bug in older versions of gcc in mingw, and only is an issue in 32 bit mode. With gcc 4.7 or newer, the issue seems to be gone. So it might not be necessary at all to add a new function just for this reason.

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1039/

----
**testing_done**:  
**description**:  Add a new public function WelsGetCodecVersionExThis function doesn&#39;t return a struct, but fills in a struct viaa provided pointer.The ABI of returning a struct is different between MSVC and mingw.This allows using the same function from mingw, even though the DLLhas been built with MSVC.

----
 time 2015-01-03T19:08:41-05:00 
2015-01-03T19:08:41-05:00 Martin Storsj (mstorsjo):
><p>Actually, this seems to only be a bug in older versions of gcc in mingw, and only is an issue in 32 bit mode. With gcc 4.7 or newer, the issue seems to be gone. So it might not be necessary at all to add a new function just for this reason.</p>
Actually, this seems to only be a bug in older versions of gcc in mingw, and only is an issue in 32 bit mode. With gcc 4.7 or newer, the issue seems to be gone. So it might not be necessary at all to add a new function just for this reason.


----
 reviewer Karina Li (ruil2) 
 time 2015-01-03T20:14:21-05:00 
2015-01-03T20:14:21-05:00 Karina Li (ruil2):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1039/

PR #1702, opened 2015-01-06
Clarify the comment about ABI issues with WelsGetCodecVersion

mstorsjo:
The incompatibility between mingw and msvc is smaller than it seemed
at first; this turned out to only be a bug in older versions.

Review at https://rbcommons.com/s/OpenH264/r/1046/.

### DIFF
--- a/codec/api/svc/codec_api.h
+++ b/codec/api/svc/codec_api.h
@@ -532,7 +532,9 @@ long WelsCreateDecoder (ISVCDecoder** ppDecoder);
 void WelsDestroyDecoder (ISVCDecoder* pDecoder);
 
 /** @brief   Get codec version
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1046/

----
**testing_done**:  
**description**:  Clarify the comment about ABI issues with WelsGetCodecVersionThe incompatibility between mingw and msvc is smaller than it seemedat first; this turned out to only be a bug in older versions.

----
 time 2015-01-07T20:07:00+03:00 
2015-01-07T20:07:00+03:00 Ethan Hugg (ethanhugg):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1046/

PR #1704, opened 2015-01-06
build: Support DESTDIR

lu-zero:
Distribution rely on that to properly package software.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ V=Yes
 PREFIX=/usr/local
 SHARED=-shared
 OBJ=o
+DESTDIR=
 SHAREDLIB_DIR=$(PREFIX)/lib
 PROJECT_NAME=openh264
 MODULE_NAME=gmpopenh264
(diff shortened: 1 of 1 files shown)


zaheerm:
Agreed, this is important. Commit looks good, please merge this.

huili2:
can’t automatically merge this pull request.

lu-zero:
Rebased again. I can update it to use more standard-ish variable names (such as `libdir` and `prefix`) if there is

mstorsjo:
Looks good to me. Due to the extra slash between $(DESTDIR) and $(PREFIX), you can't use a relative PREFIX any longer though. I'm trying to think whether a relative PREFIX is a nice-to-have or not important at all...

lu-zero:
Actually I did it on purpose to prevent relative prefixes.

mstorsjo:
Ah, ok then - I guess they seldom make sense. (For OSX, the used PREFIX gets built into the binary as well, so there it probably makes sense to have it absolute. Relative library paths can be set afterwards with install_name_tool anyway, if that's desired (like @executable_path/libopenh264.dylib).)

Thus, no objections from me.

mstorsjo:
Luca: The extra slash here seems to cause problems; on MSYS (on windows) it fails to do "make install", since it tries to write stuff to //foo/bar, where // is interpreted as the windows network share identifier.

Is it generally ok to assume that DESTDIR ends with a slash, or do we need some make magic to add one if destdir is non-empty and doesn't end with a slash?

lu-zero:
Let's keep it simple and drop the /.

PR #1710, opened 2015-01-08
prevent error propagation for parse only when frame is not complete

huili2:
see:
https://rbcommons.com/s/OpenH264/r/1051/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1051/

----
**testing_done**:  
**description**:  <p>if frame incomplete for parse only usage, mark current decoded frame as error to prevent succeeding frame decoding.</p>

----
 time 2015-01-08T07:29:49+00:00 
2015-01-08T07:29:49+00:00 Juanny Wang (JuannyWang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1051/

PR #1715, opened 2015-01-13
modify parse only flowchart to use DecodeFrameConstruction()

huili2:
let parse only usage call DecodeFrameConstruction() to be consistent with normal decoding.
Modify some fixes and flowcharts with parse only.
Fix a early-return bug before construction related to active sps reset.
see:
https://rbcommons.com/s/OpenH264/r/1057/

### DIFF
--- a/codec/decoder/core/src/decoder_core.cpp
+++ b/codec/decoder/core/src/decoder_core.cpp
@@ -45,8 +45,6 @@
 
 namespace WelsDec {
 static inline int32_t DecodeFrameConstruction (PWelsDecoderContext pCtx, uint8_t** ppDst, SBufferInfo* pDstInfo) {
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1057/

----
**testing_done**:  
**description**:  <p>Add bug fix for active sps clean before error return.<br />add some protections in DecodeFrameConstruction() to involve parse only flowchart.<br />And modify corresponding codes before this call.</p>

----
 time 2015-01-13T02:23:46+00:00 
2015-01-13T02:23:46+00:00 Juanny Wang (JuannyWang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1057/

PR #1716, opened 2015-01-13
add option for generating pdb in windows builds

sijchen:
being reviewed at: https://rbcommons.com/s/OpenH264/r/1059/


ethanhugg:
I would be more inclined to put this in SHLDFLAGS and use PROJECT_NAME like this:

-SHLDFLAGS=-link -def:openh264.def -implib:$(EXTRA_LIBRARY)
+SHLDFLAGS=-Zi -Fd$(PROJECT_NAME).pdb -link -def:$(PROJECT_NAME).def -implib:$(EXTRA_LIBRARY)

Does that still make a usable PDB for you?

mstorsjo:
Ethan: I think this is needed while building every object file, as well as while linking. Also, as I pointed out on reviewboard, the -Zi is duplicated (it's in CFLAGS_OPT/CFLAGS_DEBUG already), and there's no point in adding that flag to the linker (iirc that's a compiler-only flag).

sijchen:
Ethan @ethanhugg , at least in my trial, only have that -Fd in linker line seemed cannot generate the full pdb
I will update the reviewboard according to the comments.

sijchen:
open another pull to avoid many commits: https://github.com/cisco/openh264/pull/1722

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1059/

----
**testing_done**:  <h1>this way is not working (cannot see symbols)</h1><p>CFLAGS += -nologo -Zi -Fd$(PROJECT_NAME).pdb -W3 -EHsc -fp:precise -Zc:wchar_t -Zc:forScope -D_VARIADIC_MAX=10<br />CXX_LINK_O=-nologo -Fd$(PROJECT_NAME).pdb -Fe$@<br />CFLAGS_OPT=-O2 -Ob1 -Oy- -GF -Gm- -GS -Gy -DNDEBUG<br />CFLAGS_DEBUG=-Od -Oy- -RTC1 -D_DEBUG<br />SHLDFLAGS=-link -def:$(PROJECT_NAME).def -implib:$(EXTRA_LIBRARY)<br />=============</p>
**description**:  add option for generating pdb in windows builds

----
 time 2015-01-13T09:33:57+00:00 
2015-01-13T09:33:57+00:00 Martin Storsj (mstorsjo):
 on **build/msvc-common.mk line25**: 
 id="comment205236"><p>This duplicates the -Zi parameter, which is already set in CFLAGS_OPT/CFLAGS_DEBUG below.</p>
This duplicates the -Zi parameter, which is already set in CFLAGS_OPT/CFLAGS_DEBUG below.

2015-01-14T04:13:37+00:00 Sijia Chen (sijchen):
><p>interestingly that when I tried to have "-Zi" in CFLAGS and remote the two in CFLAGS_OPT/CFLAGS_DEBUG, it cannot generate the useful pdb as previous;  any hint on the reason? Thanks</p>
interestingly that when I tried to have "-Zi" in CFLAGS and remote the two in CFLAGS_OPT/CFLAGS_DEBUG, it cannot generate the useful pdb as previous;  any hint on the reason? Thanks


----
2015-01-14T04:11:32+00:00 status-change:
CFLAGS += -nologo -Zi -Fd$(PROJECT_NAME).pdb -W3 -EHsc -fp:precise -Zc:wchar_t -Zc:forScope -D_VARIADIC_MAX=10<br/></td></tr><tr class="insert"> <td class="marker">&nbsp;</td> <td class="marker">+</td> <td class="line rich-text">CXX_LINK_O=-nologo -Fd$(PROJECT_NAME).pdb -Fe$@<br/></td></tr><tr class="insert"> <td class="marker">&nbsp;</td> <td class="marker">+</td> <td class="line rich-text">CFLAGS_OPT=-O2 -Ob1 -Oy- -GF -Gm- -GS -Gy -DNDEBUG<br/></td></tr><tr class="insert"> <td class="marker">&nbsp;</td> <td class="marker">+</td> <td class="line rich-text">CFLAGS_DEBUG=-Od -Oy- -RTC1 -D_DEBUG<br/></td></tr><tr class="insert"> <td class="marker">&nbsp;</td> <td class="marker">+</td> <td class="line rich-text">SHLDFLAGS=-link -def:$(PROJECT_NAME).def -implib:$(EXTRA_LIBRARY)<br/></td></tr><tr class="insert"> <td class="marker">&nbsp;</td> <td class="marker">+</td> <td class="line rich-text">=============


----
 time 2015-01-14T07:13:07+00:00 
2015-01-14T07:13:07+00:00 Martin Storsj (mstorsjo):
 on **build/msvc-common.mk line16**: 
 id="comment205660"><p>Don't set PROJECT_NAME here, it is already set in the main Makefile (and should work just fine here without setting it again). By setting it here it is much harder to change this variable, if that ever is desired.</p>
Don't set PROJECT_NAME here, it is already set in the main Makefile (and should work just fine here without setting it again). By setting it here it is much harder to change this variable, if that ever is desired.

2015-01-14T07:13:07+00:00 Martin Storsj (mstorsjo):
 on **build/msvc-common.mk line44**: 
 id="comment205661"><p>I'd rather keep the -def parameter as it is right now, because it refers to an actual file with this name that is committed in the repo, not what the name of a new generated file should be. If you want to do a build with PROJECT_NAME set to something else, this would fail the build since it wouldn't find such a .def file.</p>
I'd rather keep the -def parameter as it is right now, because it refers to an actual file with this name that is committed in the repo, not what the name of a new generated file should be. If you want to do a build with PROJECT_NAME set to something else, this would fail the build since it wouldn't find such a .def file.


----
2015-01-14T08:29:06+00:00 status-change:
fix according to comments


----
 time 2015-01-14T08:50:25+00:00 
2015-01-14T08:50:25+00:00 Martin Storsj (mstorsjo):
>Please add *.pdb to the clean rule in Makefile as well (sorry for not pointing this out earlier).

2015-01-14T08:50:25+00:00 Martin Storsj (mstorsjo):
 on **build/msvc-common.mk line43**: 
 id="comment205669"><p>This causes warnings like "LINK : warning LNK4044: unrecognized option '/Fdopenh264.pdb'; ignored". Parameters before the "-link" flag are handled by cl.exe (and translated to the link.exe form), while parameters set after the "-link" flag are sent unchanged to link.exe. The cl.exe parameter is -Fd&lt;filename.pdb&gt; while the link.exe parameter is -pdb:&lt;filename.pdb&gt;.</p><p>So either change it to "-Fd$(PROJECT_NAME).pdb -link" or into "-link -pdb:$(PROJECT_NAME).pdb" (both work equally well).</p><p>(Sorry I didn't notice this in the previous version - I only noticed it now when I tested building this code locally.)</p>
This causes warnings like "LINK : warning LNK4044: unrecognized option '/Fdopenh264.pdb'; ignored". Parameters before the "-link" flag are handled by cl.exe (and translated to the link.exe form), while parameters set after the "-link" flag are sent unchanged to link.exe. The cl.exe parameter is -Fd&lt;filename.pdb&gt; while the link.exe parameter is -pdb:&lt;filename.pdb&gt;.
So either change it to "-Fd$(PROJECT_NAME).pdb -link" or into "-link -pdb:$(PROJECT_NAME).pdb" (both work equally well).
(Sorry I didn't notice this in the previous version - I only noticed it now when I tested building this code locally.)

2015-01-14T10:04:46+00:00 Sijia Chen (sijchen):
><p>I used the former one, since I tried -pdb before which seemingly does not work, but I forget if I tried that independently.</p>
I used the former one, since I tried -pdb before which seemingly does not work, but I forget if I tried that independently.


----
2015-01-14T10:05:12+00:00 status-change:
interestingly that when I tried to have "-Zi" in CFLAGS and remote the two in CFLAGS_OPT/CFLAGS_DEBUG, it cannot generate the useful pdb as previous;  any hint on the reason? Thanks


----
 time 2015-01-14T10:08:36+00:00 
2015-01-14T10:08:36+00:00 Martin Storsj (mstorsjo):
>Ship It!
Ship It!
Ship it!


----
 time 2015-01-14T15:35:36+00:00 
2015-01-14T15:35:36+00:00 Ethan Hugg (ethanhugg):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1059/

PR #1725, opened 2015-01-14
gmp-openh264 should avoid writing outside the input frame buffer with start code.

ethanhugg:
Review here - https://rbcommons.com/s/OpenH264/r/1064/

### DIFF
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -622,7 +622,9 @@ class OpenH264VideoDecoder : public GMPVideoDecoder {
 
     case GMP_BufferLength32: {
       uint8_t* start_code = inputFrame->Buffer();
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1064/

----
**testing_done**:  
**description**:  <p>This was reported in this bug - https://bugzilla.mozilla.org/show_bug.cgi?id=1058887</p><p>It is not considered critical because our input here is Firefox's webrtc code, but we should prevent overwrites here in case that changes.</p>

----

----
 reviewer Karina Li (ruil2) 
 time 2015-01-15T02:11:52+00:00 
2015-01-15T02:11:52+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1064/

PR #1726, opened 2015-01-15
Fix the decode failed status update bug under EC off mode

HaiboZhu:
Fix the bug that the pDec not set to NULL when early exit


HaiboZhu:
UT failed, need to modify the UT code

PR #1731, opened 2015-01-15
Fixes for some new compiler warnings

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/1066/.

### DIFF
--- a/codec/encoder/core/src/ratectl.cpp
+++ b/codec/encoder/core/src/ratectl.cpp
@@ -1367,7 +1367,7 @@ void  WelsRcPictureInitGomTimeStamp (sWelsEncCtx* pEncCtx, long long uiTimeStamp
 
   RcInitSliceInformation (pEncCtx);
   RcInitGomParameters (pEncCtx);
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1066/

----
**testing_done**:  
**description**:  Declare a float constant as float instead of as doubleThis avoids warnings about converting from double to floatin MSVC.Add explicit casts when converting from double to int32_tThis fixes build warnings with MSVC.Consistently use the right type for log printsThese fields were originally int64_t. Even though long long isthe same size as int64_t in all setups, some compilers treatit as a different type than long long (in 64 bit environments,int64_t is a long, not a long long, even if they are the samesize).%&quot;PRId64&quot; is used for printing int64_t, while %lld is usedfor printing long long.This fixes build warnings with gcc.

----
 reviewer Karina Li (ruil2) 
 time 2015-01-16T01:00:28+00:00 
2015-01-16T01:00:28+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1066/

PR #1734, opened 2015-01-16
Avoid a warning about a variable set but not used

mstorsjo:
The return values may be useful later, so instead of removing the
variable, just silence the warning.

This fixes a warning in builds with GCC.

Review at https://rbcommons.com/s/OpenH264/r/1067/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1067/

----
**testing_done**:  
**description**:  Avoid a warning about a variable set but not usedThe return values may be useful later, so instead of removing thevariable, just silence the warning.This fixes a warning in builds with GCC.

----
 reviewer Sijia Chen (sijchen) 
 time 2015-01-19T02:51:19+00:00 
2015-01-19T02:51:19+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1067/

PR #1735, opened 2015-01-16
Fix build warnings with MSVC

mstorsjo:
After this, the whole project, including the testsuite, builds without warnings with MSVC, for both ARM, x86 and x86_64.

Review at https://rbcommons.com/s/OpenH264/r/1068/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1068/

----
**testing_done**:  
**description**:  <p>After this, the whole project, including the testsuite, builds without warnings with MSVC, for both ARM, x86 and x86_64.</p><p></p><p>Use VerifyVersionInfo instead of GetVersionEx</p><p>The GetVersionEx function has been deprecated since Windows 8.1.<br />VerifyVersionInfo exists at least since Windows 2000.</p><p>Move a conversion from size_t to int into one single place within BaseDecoderTest.cpp</p><p></p><p>Add a cast when converting from uint64_t to int</p><p>Add casts when converting from size_t or std::streamsize to int</p><p></p><p>Add casts when converting from float to int</p><p></p><p>Silence warnings about insecure CRT functions in the test suite</p><p>There is no problem in using the "insecure" CRT functions, as long as<br />they are used correctly - especially not within the test suite where<br />they are only are exposed to the test suite input data.</p><p>Within the library, these are used via the library internal wrappers<br />in codec/common/src/crt_util_safe_x.cpp, but we'd rather not use them<br />in the test suite - just use the normal standard C functions here.</p><p>Allow setting cflags specifically for the unit test suite</p><p></p><p>Use the right data type for a variable</p><p>There is little point in using an int32_t variable for something<br />that will only be used for passing in as a bool parameter to a<br />function.</p><p>Fix a cast to actually cast to the right type</p><p>The existing cast didn't make much sense. The expression itself is<br />already int, there's little point in casting it to int, especially<br />when assigning to a float.</p><p>Add a comparison in ASSERT_TRUE</p><p>This makes sure the compared value actually is bool, not any other<br />type that can be used as condition.</p><p>Consistently compare integers with integers and bools with bools</p><p>MSVC warns when comparing integers with booleans in with the gtest<br />assertion macros.</p><p>Ignore armasm warnings about instructions that are deprecated in ARMv8</p><p>This disables armasm warning A4509, "This form of conditional<br />instruction is deprecated".</p><p>The conditional instructions (such as movcs, addcs, subscs)<br />have been deprecated in ARMv8, in favor of conditional branches.<br />This isn't something that we need to take immediate action about,<br />though, therefore silence the warning.</p>

----
 time 2015-01-19T01:29:10+00:00 
2015-01-19T01:29:10+00:00 dong zhang (dongzhang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1068/

PR #1736, opened 2015-01-16
Build using -Werror in travis unit tests

mstorsjo:
This allows catching warnings that are introduced, that may only be
visible with some compilers.

Review at https://rbcommons.com/s/OpenH264/r/1069/.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ ifeq ($(BUILDTYPE), Release)
 CFLAGS += $(CFLAGS_OPT)
 USE_ASM = Yes
 else
-CFLAGS = $(CFLAGS_DEBUG)
+CFLAGS += $(CFLAGS_DEBUG)
 USE_ASM = No
 endif
 
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -62,7 +62,10 @@
 #endif
 
 // This is for supporting older versions which do not have support for nullptr.
-#if defined(__clang__)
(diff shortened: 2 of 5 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1069/

----
**testing_done**:  
**description**:  Build using -Werror in travis unit testsThis makes all compiler warnings to be treated as errors.This allows catching warnings that are introduced, that may only bevisible with some compilers.This option is only added within travis unit tests, not during normalbuilds, so there is no risk of unnecessarily breaking builds onnew/untested systems.Avoid warnings about an unused variable in debug modeCheck for an already defined nullptr macroIn addition to the existing compiler version/feature checks, alsocheck if there is such a macro.This avoids warnings when building with Xcode 6.0.Don&#39;t override the externally set CFLAGS variable in debug modeThis is similar to how it is done in release mode already.The fact that this behaviour differed between release and debugcan be traced back to commit cf92e8d6.Avoid a warning about a variable set but not usedThe return values may be useful later, so instead of removing thevariable, just silence the warning.This fixes a warning in builds with GCC.

----
 reviewer Sijia Chen (sijchen) 
 time 2015-01-19T02:53:47+00:00 
2015-01-19T02:53:47+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1069/

PR #1740, opened 2015-01-19
Fix issues in the ParameterSetStrategy tests

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/1073/.

### DIFF
--- a/test/api/encode_decode_api_test.cpp
+++ b/test/api/encode_decode_api_test.cpp
@@ -149,11 +149,11 @@ class EncodeDecodeTestBase : public ::testing::TestWithParam<EncodeDecodeFilePar
 
 
   virtual int GetRandWidth() {
-    return (WELS_CLIP3 ((((rand() % MAX_WIDTH) >> 1) + 1) << 1, 2, MAX_WIDTH));
+    return WelsClip3 ((((rand() % MAX_WIDTH) >> 1) + 1) << 1, 2, MAX_WIDTH);
   }
 
   virtual int GetRandHeight() {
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1073/

----
**testing_done**:  
**description**:  <p>Undefine a define for writing debug files</p><p></p><p>Use WelsClip3 instead of WELS_CLIP3 for random values</p><p>Since WELS_CLIP3 is a macro, using it will cause the arguments<br />to be evalated multiple times, which means that the clip will<br />not work as intended. If using a rand() argument, always use<br />WelsClip3 instead, which is an inline function.</p><p></p><p>Compare macro block widths instead of picture widths</p><p>Since cropping is disabled (prepareParam does<br />memset (pParam, 0, sizeof (SEncParamExt)); which clears the<br />cropping flag), we need to make sure that the number of<br />macroblocks differ, not only the number of pixels.</p><p></p><p>Fix a comparison in ParameterSetStrategy_SPS_LISTING_AND_PPS_INCREASING3</p><p>Previously, this test picked a width at random until it found<br />a width that had been used before, meaning that all tests were<br />made with the same single width.</p>

----
 reviewer Sijia Chen (sijchen) 
 time 2015-01-19T08:20:50+00:00 
2015-01-19T08:20:50+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1073/

PR #1742, opened 2015-01-20
[Encoder] improve the para checking of max-br

sijchen:
being reviewed at: https://rbcommons.com/s/OpenH264/r/1075/

improve the para checking of max-br, so as to avoid unknown behavior when the setting is very large

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -76,11 +76,6 @@ int32_t WelsBitRateVerification (SLogContext* pLogCtx, SSpatialLayerConfig* pLay
              pLayerParam->iSpatialBitrate);
     return ENC_RETURN_UNSUPPORTED_PARA;
   }
-  if (pLayerParam->iMaxSpatialBitrate == pLayerParam->iSpatialBitrate) {
-    WelsLog (pLogCtx, WELS_LOG_INFO,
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1075/

----
**testing_done**:  
**description**:  <p>improve the para checking of max-br, so as to avoid unknown behavior when the setting is very large</p>

----
 reviewer Karina Li (ruil2) 
 time 2015-01-20T03:39:46+00:00 
2015-01-20T03:39:46+00:00 Karina Li (ruil2):
><p>I don't think it should be encoder logic.</p>
I don't think it should be encoder logic.

2015-01-20T05:15:21+00:00 Sijia Chen (sijchen):
><p>encoder needs to reponse correctly if the setting exceeds expected range, or we can set it to UNSPECIFIED rather than level5-2 max br?</p>
encoder needs to reponse correctly if the setting exceeds expected range, or we can set it to UNSPECIFIED rather than level5-2 max br?


----
2015-01-21T08:56:20+00:00 status-change:
update the code diff which added some enhancements


----
 reviewer Karina Li (ruil2) 
 time 2015-01-21T09:26:23+00:00 
2015-01-21T09:26:23+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1075/

PR #1746, opened 2015-01-21
Various fixes for the GeckoMediaPlugin wrapper

kinetiknz:
These fixes (except for the SVideoProperty inititalization, which doesn't seem to be used anywhere) are required to use the OpenH264 GMP plugin in Firefox as a decoder for the video playback engine.  The Firefox patches to implement this are here: https://bugzilla.mozilla.org/show_bug.cgi?id=1121258

### DIFF
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -603,6 +603,7 @@ class OpenH264VideoDecoder : public GMPVideoDecoder {
     param.eOutputColorFormat = videoFormatI420;
     param.uiTargetDqLayer = UCHAR_MAX;  // Default value
     param.eEcActiveIdc = ERROR_CON_SLICE_MV_COPY_CROSS_IDR_FREEZE_RES_CHANGE; // Error concealment on.
+    param.sVideoProperty.size = sizeof(param.sVideoProperty);
     param.sVideoProperty.eVideoBsType = VIDEO_BITSTREAM_DEFAULT;
 
     if (decoder_->Initialize (&param)) {
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -548,6 +548,16 @@ class OpenH264VideoEncoder : public GMPVideoEncoder {
   FrameStats stats_;
 };
 
+uint16_t readU16BE(const uint8_t* in) {
+  return in[0] << 8 | in[1];
+}
+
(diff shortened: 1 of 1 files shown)


ethanhugg:
I built this and tried it out with the WebRTC loopback test here - http://mozilla.github.io/webrtc-landing/pc_test.html with H264 selected  I could not get it to work on either Release on Nightly.

I then applied your patch from Bug 1121258 but it doesn't build by itself so I pulled down the nightly from your Try run and still I couldn't get the WebRTC test to work.  

Does WebRTC still work for you with this patch?  Will this require a coordinated rollout of Firefox and OpenH264 plugins?

kinetiknz:
It looks like WebrtcGmpVideoCodec needs some modification before the modified OpenH264 GMP will work.  Specifically, WebrtcGmpVideoCodec.cpp#646 needs to store the buffer length in big-endian since this is what the modified GMP expects (since that's what's in AVCC, it seemed logical).  I had a bit of a debate with Randell on irc/#media about the definition of GMP_BufferLength32 is, etc. so this needs to be clarified one way or another.  The other change is that the MOZ_CRASH() calls need to be removed from WebrtcGmpVideoCodec.h in WebrtcGmpVideoDecoder's InputDataExhausted, DrainComplete, and ResetComplete implementations.  With those changes, it works.

kinetiknz:
I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1124542 for the MOZ_CRASH() issue in WebrtcGmpVideoDecoder.

kinetiknz:
8e0138d contains a bit of cleanup to remove the hardcoded offset in favour of using offsetof; also used pointer arithmetic rather than array notation which ended up more readable.

sijchen:
seems this pull request cannot be automatically merged

kinetiknz:
I've removed the controversial GMP_BufferLength32 meaning change from the set of changes, so the rest of these can be merged whenever you're ready if you think they're suitable.

I'll open a new pull request for the GMP_BufferLength32 issue once I've talked to the appropriate people and found an agreeable solution.

ethanhugg:
This one tests OK with WebRTC on Nightly so I'm OK with this change.  I'm going to ping Randell to see if we wants to review or comment.

ethanhugg:
Taking this now so it should be in v1.4.

PR #1760, opened 2015-01-26
Add const to pointers to an inline function within MC

mstorsjo:
Review at https://rbcommons.com/s/OpenH264/r/1091/.


mstorsjo:
I'll fold this into the other pull request about MC.

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1091/

----
**testing_done**:  
**description**:  Add const to pointers to an inline function within MC

----
2015-01-26T22:01:07+00:00 status-change:
I needed this change as part of the optimizer bug workaround in review request 1090, so integrated it there.


----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1091/

PR #1763, opened 2015-01-27
Remove the forceC parameter from DEF_MCCOPYTEST

mstorsjo:
Even if there actually is no SIMD optimized version of the
width==2 cases, luma function for SIMD still needs to handle
it by calling McCopyWidthEq2_c for these cases.

This simplifies the UT code a little, and makes sure that
those codepaths are tested properly.

Review at https://rbcommons.com/s/OpenH264/r/1093/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1093/

----
**testing_done**:  
**description**:  Remove the forceC parameter from DEF_MCCOPYTESTEven if there actually is no SIMD optimized version of thewidth==2 cases, luma function for SIMD still needs to handleit by calling McCopyWidthEq2_c for these cases.This simplifies the UT code a little, and makes sure thatthose codepaths are tested properly.

----
 time 2015-01-28T01:06:11+00:00 
2015-01-28T01:06:11+00:00 dong zhang (dongzhang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1093/

PR #1772, opened 2015-01-28
Move the MC implementation to the common library

mstorsjo:
This avoids code duplication.

Review at https://rbcommons.com/s/OpenH264/r/1102/.


mstorsjo:
I'll try to have a look to see why the travis build failed - it does seem like something is crashing, but it does seem to run clean in valgrind locally, so I'm not sure if it depends on the compiler version...

ethanhugg:
I will restart the failed runs now.

mstorsjo:
I finally found a compiler config locally where I can reproduce the issue, I'll try to look into it...

mstorsjo:
Ok, finally found the bug. I had missed a small detail in one of the MC functions when replacing the encoder functions with the ones from the decoder (that are a bit more generic than the special cased encoder ones, but one of the encoder functions required more scratch space in a temp buffer).

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1102/

----
**testing_done**:  
**description**:  Remove the now unnecessary mc_test_common.hMerge all MC unit tests into one fileSince both encoder and decoder use the same MC implementation now,one set of tests is enough.Move the MC routines to the common libraryUse the decoder versions of the functions (which are capableof handling widths 4/8/16 for luma, not only 16 as in theencoder). By using the more generic versions, there may be a smallperformance loss since the functions need to check the widthin every call. Actual measurements show that the actual change isvery small (and the shared routines turn out to actually be fasterthan the existing ones in ARM NEON setups).Hook up McChromaWidthEq8_ssse3 into use in the decoder as wellChange pfSampleAveraging to be a single function with internal width handlingThis makes it match the behaviour of pMcLumaFunc and pMcChromaFunc.Simplify the MC unit tests now that the naming is made consistentRename the encoder MC chroma/luma functions to match the naming in the decoderSimplify the MC unit tests now that the luma functions have a similar signatureMove the qpel function table into the luma function itselfThis unifies the luma MC interface to match the decoder side.

----
2015-01-28T22:04:34+00:00 status-change:
Fixed a crash in one case, by making sure McHorVer22_c has got a large enough tmp buffer (it was based on the decoder version that had a smaller buffer).


----
 time 2015-01-29T01:26:16+00:00 
2015-01-29T01:26:16+00:00 dong zhang (dongzhang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1102/

PR #1778, opened 2015-01-30
Don't overwrite the git-tracked version.h

mstorsjo:
When generating a new version of the header, that includes the
actual git hash, don't overwrite the file that is tracked by git.
Instead create a new file, and include this only if the build system
indicates that it exists (by setting a define). This allows the
untouched source tree to be built from within an IDE even if make
has not been run.

This reduces the hassle with a file that needs to be ignored in the
git configuration.

The downside is that the generated file isn't used if building
from within an IDE, if the header has been updated by calling make
before (since the IDE configuration doesn't know whether the user
actually has run make). Since users of the IDE might not build via
make in the command line at all (in the same source checkout at least),
this should not be an issue in practice. The previous way things worked,
the version hash (generated by make) when used in an IDE could actually
be outdated and misleading.

Review at https://rbcommons.com/s/OpenH264/r/1107/.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ SHAREDLIBVERSION=0
 include $(SRC_PATH)build/platform-$(OS).mk
 
 
-CFLAGS +=
+CFLAGS += -DGENERATED_VERSION_HEADER
 LDFLAGS +=
 
 ifeq (Yes, $(GCOV))
--- a/codec/common/generate_version.sh
+++ b/codec/common/generate_version.sh
@@ -14,13 +14,13 @@ fi
 GIT_VERSION='"'$GIT_VERSION'"'
 rm -f config.git-hash
 
-cat codec/common/inc/version.h.template | sed "s/\$FULL_VERSION/$GIT_VERSION/g" > codec/common/inc/version.h.new
-if cmp codec/common/inc/version.h.new codec/common/inc/version.h > /dev/null 2>&1; then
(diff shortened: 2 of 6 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1107/

----
**testing_done**:  
**description**:  Don&#39;t overwrite the git-tracked version.hWhen generating a new version of the header, that includes theactual git hash, don&#39;t overwrite the file that is tracked by git.Instead create a new file, and include this only if the build systemindicates that it exists (by setting a define). This allows theuntouched source tree to be built from within an IDE even if makehas not been run.This reduces the hassle with a file that needs to be ignored in thegit configuration.The downside is that the generated file isn&#39;t used if buildingfrom within an IDE, if the header has been updated by calling makebefore (since the IDE configuration doesn&#39;t know whether the useractually has run make). Since users of the IDE might not build viamake in the command line at all (in the same source checkout at least),this should not be an issue in practice. The previous way things worked,the version hash (generated by make) when used in an IDE could actuallybe outdated and misleading.

----
 reviewer Sijia Chen (sijchen) 
 time 2015-01-30T03:58:57-05:00 
2015-01-30T03:58:57-05:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1107/

PR #1785, opened 2015-02-02
fix it

syureyi:
Just fix a typo


sijchen:
please use meaningful commit contents so other can get brief understanding on the commit purpose before trying to read the code diff

PR #1797, opened 2015-02-05
Wp8 arm

zhilwang:


sijchen:
close this because it can be achieved with other better approach

PR #1799, opened 2015-02-05
Allow forcing copying in downscale

mstorsjo:
Normally, the DownsamplePadding skips scaling if the target
size is the same as the source size, assuming that the caller
will use the source data pointer in that case. This is true
for the base layer (the first call to DownsamplePadding in
SingleLayerPreprocess), but when downsampling the other layers,
there is no special handling for the case when the target
is the same size as the source.

Previously, the encoding of such spatial layers will use
completely uninitialized data, encoding complete garbage.

Instead force DownsamplePadding to make a copy if no scaling
is required, for the dependency layers. The base layer still
avoids a copy unless scaling of that layer is required.

Whether it actually makes sense to have lower spatial layers
the same size as the original one is a different question
though - currently the code allows it, and
EncodeDecodeTestAPI.SetOptionEncParamExt will try to use it.

Review at https://rbcommons.com/s/OpenH264/r/1125/.

### DIFF
--- a/codec/encoder/core/inc/wels_preprocess.h
+++ b/codec/encoder/core/inc/wels_preprocess.h
@@ -153,7 +153,8 @@ class CWelsPreProcess {
   void	BilateralDenoising (SPicture* pSrc, const int32_t iWidth, const int32_t iHeight);
   bool  DetectSceneChange (SPicture* pCurPicture, SPicture* pRefPicture);
   int32_t DownsamplePadding (SPicture* pSrc, SPicture* pDstPic,  int32_t iSrcWidth, int32_t iSrcHeight,
-                             int32_t iShrinkWidth, int32_t iShrinkHeight, int32_t iTargetWidth, int32_t iTargetHeight);
+                             int32_t iShrinkWidth, int32_t iShrinkHeight, int32_t iTargetWidth, int32_t iTargetHeight,
+                             bool bForceCopy);
 
   void    VaaCalculation (SVAAFrameInfo* pVaaInfo, SPicture* pCurPicture, SPicture* pRefPicture, bool bCalculateSQDiff,
                           bool bCalculateVar, bool bCalculateBGD);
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1125/

----
**testing_done**:  
**description**:  Allow forcing copying in downscaleNormally, the DownsamplePadding skips scaling if the targetsize is the same as the source size, assuming that the callerwill use the source data pointer in that case. This is truefor the base layer (the first call to DownsamplePadding inSingleLayerPreprocess), but when downsampling the other layers,there is no special handling for the case when the targetis the same size as the source.Previously, the encoding of such spatial layers will usecompletely uninitialized data, encoding complete garbage.Instead force DownsamplePadding to make a copy if no scalingis required, for the dependency layers. The base layer stillavoids a copy unless scaling of that layer is required.Whether it actually makes sense to have lower spatial layersthe same size as the original one is a different questionthough - currently the code allows it, andEncodeDecodeTestAPI.SetOptionEncParamExt will try to use it.

----
 reviewer Sijia Chen (sijchen) 
 time 2015-02-06T05:17:27+00:00 
2015-02-06T05:17:27+00:00 Sijia Chen (sijchen):
><p>Hi, I will rather add parameter checking to forbid "have lower spatial layers the same size as the original one", at least for now. We did not fully test that case, and moreover, allowing this will give a misleading impression that we supports something like CGS in SVC.</p>
Hi, I will rather add parameter checking to forbid "have lower spatial layers the same size as the original one", at least for now. We did not fully test that case, and moreover, allowing this will give a misleading impression that we supports something like CGS in SVC.


----
 reviewer Karina Li (ruil2) 
 time 2015-02-06T05:18:58+00:00 
2015-02-06T05:18:58+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----
 reviewer Sijia Chen (sijchen) 
 time 2015-02-06T06:21:42+00:00 
2015-02-06T06:21:42+00:00 Sijia Chen (sijchen):
><p>anyway, this is a harmless protection to complete the function's logic</p>
anyway, this is a harmless protection to complete the function's logic
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1125/

PR #1803, opened 2015-02-06
[Encoder] fix potential overflow (will change bs but little impact on bs)

sijchen:
https://rbcommons.com/s/OpenH264/r/1105/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1105/

----
**testing_done**:  <p>Tested that it will change bs in some cases but impacts little in RD, will update SHA later</p>
**description**:  <p>fix potential overflow under large bit rate</p>

----
2015-01-30T08:40:00+00:00 status-change:
Tested that it will change bs in some cases but impacts little in RD, will update SHA later


----

----
 reviewer Karina Li (ruil2) 
 time 2015-02-03T06:17:14+00:00 
2015-02-03T06:17:14+00:00 Karina Li (ruil2):
 on **codec/processing/src/complexityanalysis/ComplexityAnalysis.cpp line106**: 
 id="comment211183"><p>suggest use int64 type to avoid the overflow</p>
suggest use int64 type to avoid the overflow

2015-02-03T06:28:56+00:00 Sijia Chen (sijchen):
><p>I did not think we need that since INT_MAX already represents large complexity. But anyway I can fix it.</p>
I did not think we need that since INT_MAX already represents large complexity. But anyway I can fix it.


----
2015-02-03T06:58:56+00:00 status-change:
use longer type for complexity in preprocess modules and remove previous clipping, no BS change with this


----
 reviewer Karina Li (ruil2) 
 time 2015-02-06T02:27:55+00:00 
2015-02-06T02:27:55+00:00 Karina Li (ruil2):
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1105/

PR #1816, opened 2015-02-19
Avoid trying to downscale layers to dimensions below 2

mstorsjo:
When calculating what resolution to actually downscale to,
it can end up smaller than what the caller set. When scaling
down to resolutions close to the limit of allowed values,
this can end up setting values lower than the limit.

Previously, e.g. a downscale from 2266x8 to 566x2 will end
up as 566x1 after this calculation. When scaling to a
566x1, the chroma plane gets a height of 0, which doesn't
make sense, and which breaks e.g. the SSE2 scaler. Therefore,
make sure none of the dimensions end up set below 2.

Review at https://rbcommons.com/s/OpenH264/r/1131/.

### DIFF
--- a/codec/encoder/core/src/wels_preprocess.cpp
+++ b/codec/encoder/core/src/wels_preprocess.cpp
@@ -430,9 +430,9 @@ bool JudgeNeedOfScaling (SWelsSvcCodingParam* pParam, Scaled_Picture* pScaledPic
 
     if (iInputWidthXDstHeight > iInputHeightXDstWidth) {
       pScaledPicture->iScaledWidth[iSpatialIdx] = iCurDstWidth;
-      pScaledPicture->iScaledHeight[iSpatialIdx] = iInputHeightXDstWidth / kiInputPicWidth;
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1131/

----
**testing_done**:  
**description**:  Avoid trying to downscale layers to dimensions below 2When calculating what resolution to actually downscale to,it can end up smaller than what the caller set. When scalingdown to resolutions close to the limit of allowed values,this can end up setting values lower than the limit.Previously, e.g. a downscale from 2266x8 to 566x2 will endup as 566x1 after this calculation. When scaling to a566x1, the chroma plane gets a height of 0, which doesn&#39;tmake sense, and which breaks e.g. the SSE2 scaler. Therefore,make sure none of the dimensions end up set below 2.

----
 time 2015-02-23T21:53:48+00:00 
2015-02-23T21:53:48+00:00 Ethan Hugg (ethanhugg):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1131/

PR #1823, opened 2015-02-28
Don't downsample to anything smaller than 4x4 pixels

mstorsjo:
This makes for a chroma plane of 2x2. The SIMD versionf of generic
downscalers assume that the width and height is at least 2, since
it does an unconditional loop for the body of the image, and a
separate step for the last pixel and last row. The SIMD versions
assume that (width-1) and (height-1) are larger than zero.

This fixes spurious crashes in EncodeDecodeTestAPI.SetOptionEncParamExt.

Review at https://rbcommons.com/s/OpenH264/r/1135/.

### DIFF
--- a/codec/encoder/core/src/wels_preprocess.cpp
+++ b/codec/encoder/core/src/wels_preprocess.cpp
@@ -430,9 +430,9 @@ bool JudgeNeedOfScaling (SWelsSvcCodingParam* pParam, Scaled_Picture* pScaledPic
 
     if (iInputWidthXDstHeight > iInputHeightXDstWidth) {
       pScaledPicture->iScaledWidth[iSpatialIdx] = iCurDstWidth;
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1135/

----
**testing_done**:  
**description**:  Don&#39;t downsample to anything smaller than 4x4 pixelsThis makes for a chroma plane of 2x2. The SIMD versionf of genericdownscalers assume that the width and height is at least 2, sinceit does an unconditional loop for the body of the image, and aseparate step for the last pixel and last row. The SIMD versionsassume that (width-1) and (height-1) are larger than zero.This fixes spurious crashes in EncodeDecodeTestAPI.SetOptionEncParamExt.

----
 time 2015-03-02T00:43:50+00:00 
2015-03-02T00:43:50+00:00 zhiliang wang (zhilwang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1135/

PR #1830, opened 2015-03-04
allow slice-level data come in for parse only

huili2:
allow slice-level inpout of one AU into parseonly decoder.
see:
https://rbcommons.com/s/OpenH264/r/1136/


huili2:
close because of warning exists. will open after fixed.

PR #1831, opened 2015-03-04
[Encoder] parameter checking related to loop-filter-idc

sijchen:
https://rbcommons.com/s/OpenH264/r/1141/

### DIFF
--- a/codec/api/svc/codec_def.h
+++ b/codec/api/svc/codec_def.h
@@ -121,6 +121,17 @@ enum ENalPriority {
 #define FRAME_NUM_PARAM_SET		(-1)
 #define FRAME_NUM_IDR			0
 
+/**
+ * @brief eDeblockingIdc
+ */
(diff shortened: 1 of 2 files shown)


ethanhugg:
I'm guessing you were opening/closing this pull req to restart the travis run since they were gray.  I can restart travis runs for you on the travis site so if you get gray runs you can leave me a note and I can retry them for you.   I'm retrying 1833 and 1834 now so hopefully they'll be complete by the time you come on.

ethanhugg:
Also, anyone with merge rights to cisco/openh264 can restart jobs by signing into travis-ci.org.  The restart button is in the upper right when looking at a travis run.  That would be preferable to open/close the pull request because no email is sent to everyone in that case.

sijchen:
I see that, yes that was what I want to try. Thanks for telling me the way.

PR #1835, opened 2015-03-05
[UT] add more condition to improve the coverage of DecoderCrashTest

sijchen:
https://rbcommons.com/s/OpenH264/r/1132/


dongzha:
Should this be two commits?

PR #1845, opened 2015-03-12
use each spatial layer qp instead of global qp and update traces

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/1149/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1149/

----
**testing_done**:  
**description**:  use each spatial layer qp instead of global qp and update traces

----
 reviewer Sijia Chen (sijchen) 
 time 2015-03-12T07:28:30+00:00 
2015-03-12T07:28:30+00:00 Sijia Chen (sijchen):
Ship it!

2015-03-12T07:28:30+00:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/ratectl.cpp line1318**: 
 id="comment223482"><p>will this change affect bs?</p>
will this change affect bs?

2015-03-12T08:49:58+00:00 Karina Li (ruil2):
><p>yes, it will change bs in some cases for RC_TIMESTAMP_MODE. But UT doesn't cover this case.</p>
yes, it will change bs in some cases for RC_TIMESTAMP_MODE. But UT doesn't cover this case.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1149/

PR #1867, opened 2015-03-22
Update .gitignore

GabrielDuque:


HaiboZhu:
What is the meaning of the parameter -id? Thanks

HaiboZhu:
It seems no reply till now.

PR #1876, opened 2015-03-25
fix vui writting

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/1165/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1165/

----
**testing_done**:  
**description**:  fix vui writting

----
 reviewer Sijia Chen (sijchen) 
 time 2015-03-25T02:27:43-04:00 
2015-03-25T02:27:43-04:00 Sijia Chen (sijchen):
Ship it!

2015-03-25T02:27:43-04:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/au_set.cpp line284**: 
 id="comment227472"><p>it will be easier for maintainance if we use pSps-&gt;bVuiParamPresentFlag rather than directly using a 'true'</p>
it will be easier for maintainance if we use pSps-&gt;bVuiParamPresentFlag rather than directly using a 'true'


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1165/

PR #1883, opened 2015-03-27
Remove duplication in arm assembly

mstorsjo:
Remove duplicated macros for the apple specific syntax, when apple's assembler these days can handle the binutils syntax just fine (and we already require such a new assembler, by using other gnu binutils style directives that older apple assemblers don't support).

Review at https://rbcommons.com/s/OpenH264/r/1169/.


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1169/

----
**testing_done**:  
**description**:  <p>Remove apple specific versions of arm macros with arguments</p><p>The apple assembler for arm can handle the gnu binutils style<br />macros just fine these days, so there is no need to duplicate all<br />of these macros in two syntaxes, when the new one works fine in all cases.</p><p>We already require a new enough assembler to support the gnu binutils<br />style features since we use the .rept directive in a few places.</p><p></p><p></p><p>Remove apple specific versions of arm64 macros with arguments</p><p>The apple assembler for arm64 can handle the gnu binutils style<br />macros just fine, so there is no need to duplicate all of these<br />macros in two syntaxes, when the new one works fine in all cases.</p><p>We already require a new enough assembler to support the gnu binutils<br />style features since we use the .rept directive in a few places.</p><p></p><p></p><p>Remove a stray, unused parameter to a gnu binutils style macro in arm assembly</p><p></p><p>Remove superfluous .text directives at the start of arm assembly files</p><p>This directive can be set by the common include header that is<br />included by all files anyway.</p>

----
 time 2015-03-30T07:59:51+00:00 
2015-03-30T07:59:51+00:00 zhiliang wang (zhilwang):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1169/

PR #1888, opened 2015-04-07
add Codec demo and auto build script for WP8

shihuade:
1. add Codec demo and auto build script for WP8	

2. review board:
  https://rbcommons.com/s/OpenH264/r/1170/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1170/

----
**testing_done**:  <p>--Add Codec demo for WP8 (openh264 Dll will be generated by script )<br />--Add UT demo for WP8<br />--Auto bat script for generating dll (debug/release c-only/asm)<br />--please ignore the win32 project change</p>
**description**:  Add WP8 codec/UT demo and auto script for building dll

----
 time 2015-04-02T02:08:16-07:00 
2015-04-02T02:08:16-07:00 Martin Storsj (mstorsjo):
 on **.gitattributes line1**: 
 id="comment230138"><p>Why is this new .gitattributes file needed? I don't think it should be committed, and especially not as part of this change. If you want to add it, add it in a separate change and explain why you think it should be added.</p>
Why is this new .gitattributes file needed? I don't think it should be committed, and especially not as part of this change. If you want to add it, add it in a separate change and explain why you think it should be added.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
><p>OK,As there will be ending line issue (LF and CRLF in different OS),some binary file such as .png/.264  or something like that should be declare as binary file,thus  LF or CRLF symbol change in such kind of file can be ignored and will not be add to  the commit .<br />Actually, this file is automation generated in previous commit.<br />I think it is better to use .gitattributes file in the future.</p>
OK,As there will be ending line issue (LF and CRLF in different OS),some binary file such as .png/.264  or something like that should be declare as binary file,thus  LF or CRLF symbol change in such kind of file can be ignored and will not be add to  the commit .<br />Actually, this file is automation generated in previous commit.<br />I think it is better to use .gitattributes file in the future.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
 on **.gitignore line25**: 
 id="comment230140"><p>What .pl file needs to be ignored within the openh264 project here? I don't think this is needed.</p>
What .pl file needs to be ignored within the openh264 project here? I don't think this is needed.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
><p>as wp8 build with make, need gas-script,and most of time gas-preprocessor is located under root dir.so with this ingore pattern,no *pl file will be committed.<br />And now there is PATH setting in the auto script,at this time, I can remove the ignore pattern.</p>
as wp8 build with make, need gas-script,and most of time gas-preprocessor is located under root dir.so with this ingore pattern,no *pl file will be committed.<br />And now there is PATH setting in the auto script,at this time, I can remove the ignore pattern.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
 on **codec/build/windowsphone/all/codec.def line1**: 
 id="comment230141"><p>Why is this file needed? You don't need a .def file when creating Windows Runtime Component DLL files.</p>
Why is this file needed? You don't need a .def file when creating Windows Runtime Component DLL files.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
><p>it is for export function of dll,which used in previous codec project,and I have moved it from the solution.<br />Forgot to removed it, will remove it later.Thanks for your remind.</p>
it is for export function of dll,which used in previous codec project,and I have moved it from the solution.<br />Forgot to removed it, will remove it later.Thanks for your remind.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
 on **codec/console/dec/src/h264dec.cpp line458**: 
 id="comment230142"><p>Why do you need to load openh264.dll dynamically? You can just as well link to openh264.lib; then you don't need to do this extra change right now. If you specifically want to test it, sure, that's a different thing, but this review request is already huge, so try to make sure it has the absolute minimum number of changes that actually are necessary. If you want to add extra features you can add them in separate review requests later.</p>
Why do you need to load openh264.dll dynamically? You can just as well link to openh264.lib; then you don't need to do this extra change right now. If you specifically want to test it, sure, that's a different thing, but this review request is already huge, so try to make sure it has the absolute minimum number of changes that actually are necessary. If you want to add extra features you can add them in separate review requests later.

2015-04-03T00:24:58-07:00 Martin Storsj (mstorsjo):
><p>Yes, this feature is for test the dll. sorry for  huge review request.<br />And will split this review request into small pieces</p>
I meant openh264_dll.lib, not openh264.lib which is the static version.
Yes, this feature is for test the dll. sorry for  huge review request.<br />And will split this review request into small pieces

2015-04-03T00:24:58-07:00 Martin Storsj (mstorsjo):
 on **codec/console/enc/src/welsenc.cpp line471**: 
 id="comment230143"><p>This is an unrelated change</p>
This is an unrelated change

2015-04-03T00:24:58-07:00 Martin Storsj (mstorsjo):
 on **codec/console/enc/src/welsenc.cpp line1052**: 
 id="comment230144"><p>Same thing here, please link to openh264.lib in the project file, instead of dynamically loading it. Then you don't need the <code>_LOAD_CODEC_DLL_</code> changes at all, right now.</p>
Same thing here, please link to openh264.lib in the project file, instead of dynamically loading it. Then you don't need the <code>_LOAD_CODEC_DLL_</code> changes at all, right now.

2015-04-03T00:24:58-07:00 Martin Storsj (mstorsjo):
 on **codec/processing/build/win32/WelsVP.vcproj line1**: 
 id="comment230145"><p>Here you are committing git conflict marrkers into the file, breaking normal visual studio builds. This is not ok.</p>
Here you are committing git conflict marrkers into the file, breaking normal visual studio builds. This is not ok.

2015-04-03T00:24:58-07:00 Martin Storsj (mstorsjo):
 on **test/api/simple_test.cpp line8**: 
 id="comment230146"><p>Why are you using a different define here (WELS_WIN_PHONE) than for the console tools (where you used WINDOWS_PHONE instead)? Please use the same define in both places.</p>
Why are you using a different define here (WELS_WIN_PHONE) than for the console tools (where you used WINDOWS_PHONE instead)? Please use the same define in both places.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
><p>OK, all change to WINDOWS_PHONE</p>
OK, all change to WINDOWS_PHONE

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
 on **test/build/windowsphone/codec_ut/Codec_UT.def line2**: 
 id="comment230147"><p>Why does this module need a .def file? I don't think it is needed, please do not include it unless it is absolutely needed.</p>
Why does this module need a .def file? I don't think it is needed, please do not include it unless it is absolutely needed.

2015-04-07T00:03:50-07:00 Forrest Shi (huade):
><p>Yes - if you build gtest as part of the make build process, it works just fine - you don't need any msvc project file for it. I suggested that you build the whole unit test suite with make, just as you do for the codec library itself, and just as you do for libut.a for iOS (named e.g. "ut.lib" for msvc). Then just have the unit test project file link in ut.lib. This should require very little modifications to the current build system, and would avoid duplicating all of the unit test listings.</p>
3.WP8 based UT cases will be added in the future, so for convinience debug,adding new cases, Codec_UT_Dll.vcxproj is need for the UT demo solution.
4.by the way, does the make test command link gtest and all other related lib into one dll, so that it can be used in WP8_UT demo<br />  thanks!
Hi,matin,<br />    You said there is successful build for latest gtest on windows phone 8/8.1. are there any special setting for gtest build?<br />    There is no macro WINAPI_FAMILY=WINAPI_FAMILY_PHONE_APP.<br />    And the only macro for windows phone is GTEST_OS_WINDOWS_MOBILE.
Yes - if you build gtest as part of the make build process, it works just fine - you don't need any msvc project file for it. I suggested that you build the whole unit test suite with make, just as you do for the codec library itself, and just as you do for libut.a for iOS (named e.g. "ut.lib" for msvc). Then just have the unit test project file link in ut.lib. This should require very little modifications to the current build system, and would avoid duplicating all of the unit test listings.

2015-04-07T00:03:50-07:00 Forrest Shi (huade):
 on **test/build/windowsphone/codec_ut/Codec_UT_Dll.vcxproj line148**: 
 id="comment230148"><p>Here you are duplicating all of the unit test definitions again. Please do not duplicate such things unless absolutely necessary. For the unit tests on iOS, we build a separate libut.a, which the xcode project file then uses. I think we should do the same here, to avoid duplicating all of the unit test file listings here. Every time you duplicate it, you add more work burden to everybody who will add/rename/remove unit tests.</p><p>Also, since this probably requires changes to the makefile, I would very much prefer if you split your changes. First submit the autobuild bat file and the project files for building the normal console tools (which really should not need almost any changes to existing code at all, only adding new project files). Then once that is done and is merged, we can add project files for unit tests on windows phone in a separate step.</p>
Here you are duplicating all of the unit test definitions again. Please do not duplicate such things unless absolutely necessary. For the unit tests on iOS, we build a separate libut.a, which the xcode project file then uses. I think we should do the same here, to avoid duplicating all of the unit test file listings here. Every time you duplicate it, you add more work burden to everybody who will add/rename/remove unit tests.
Also, since this probably requires changes to the makefile, I would very much prefer if you split your changes. First submit the autobuild bat file and the project files for building the normal console tools (which really should not need almost any changes to existing code at all, only adding new project files). Then once that is done and is merged, we can add project files for unit tests on windows phone in a separate step.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
><p>OK,right!</p>
OK,right!

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
 on **test/build/windowsphone/codec_ut/Codec_UT_Lib.vcxproj line2**: 
 id="comment230150"><p>Here you are adding a separate vcxproj file called Codec_UT_Lib, but the .sln file doesn't seem to refer to it? Is it added by accident, or why do you need two .vcxproj files that do the exact same thing?</p>
Here you are adding a separate vcxproj file called Codec_UT_Lib, but the .sln file doesn't seem to refer to it? Is it added by accident, or why do you need two .vcxproj files that do the exact same thing?

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
><p>Forgot to remove the file. sorry for my careless!<br />removed and actually only demo project is need in the future, right!</p>
Forgot to remove the file. sorry for my careless!<br />removed and actually only demo project is need in the future, right!

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
 on **test/build/windowsphone/codec_ut/Codec_UT_RTComponent.cpp line41**: 
 id="comment230154"><p>Please link the library via a .lib file in the project file instead of loading it at runtime.</p>
Please link the library via a .lib file in the project file instead of loading it at runtime.

2015-04-03T00:24:58-07:00 Forrest Shi (huade):
 on **test/build/windowsphone/codec_ut/GTestRTLib.vcxproj line1**: 
 id="comment230156"><p>If we build all of the unit tests via the makefile as I suggested, you can remove the GTestRTLib.vcxproj file.</p>
If we build all of the unit tests via the makefile as I suggested, you can remove the GTestRTLib.vcxproj file.


----
 time 2015-04-02T02:16:03-07:00 
2015-04-02T02:16:03-07:00 Martin Storsj (mstorsjo):
><p>Please do not commit unrelated changes to files - there were even changes to x86 asm source files here! Even if it is just whitespace changes, do not commit and include such changes here. This means that you actually will need to look at what you commit, instead of telling git to commit every single change without looking. You had accidentally committed git conflict markers into a project file at one place as well.</p><p>Please remove all the unrelated extra changes, link to openh264_dll.lib instead of loading openh264.dll at runtime (remove the LOAD_CODEC_DLL code), and split this commit and review request so that you only add what is absolutely necessary for building the codec app (the autobuild bat file and the new files under codec/build/windowsphone), but nothing else. Then once that is ok, you can try to add project files for the unit tests, but by following my advice. But please defer that to a later review request, because this one is already huge, and the larger the changes are, the harder it is to find all the extra issues that you accidentally are adding.</p>
Please do not commit unrelated changes to files - there were even changes to x86 asm source files here! Even if it is just whitespace changes, do not commit and include such changes here. This means that you actually will need to look at what you commit, instead of telling git to commit every single change without looking. You had accidentally committed git conflict markers into a project file at one place as well.
Please remove all the unrelated extra changes, link to openh264_dll.lib instead of loading openh264.dll at runtime (remove the LOAD_CODEC_DLL code), and split this commit and review request so that you only add what is absolutely necessary for building the codec app (the autobuild bat file and the new files under codec/build/windowsphone), but nothing else. Then once that is ok, you can try to add project files for the unit tests, but by following my advice. But please defer that to a later review request, because this one is already huge, and the larger the changes are, the harder it is to find all the extra issues that you accidentally are adding.

2015-04-03T00:25:09-07:00 Forrest Shi (huade):
><p>thanks for your suggestions and your patience.</p>
thanks for your suggestions and your patience.


----
2015-04-03T01:33:47-07:00 status-change:
--split privous review request into small pieces--Codec demo/UT demo<br />--only codec demo for WP8 and auto build script <br />--auto build script,please refer to: build/AutoBuildForWP8VC2013.bat<br />-- will add UT demo once gtest for wp8 with make file ready


----
 time 2015-04-03T04:17:12-07:00 
2015-04-03T04:17:12-07:00 Martin Storsj (mstorsjo):
><p>Thanks, this looks ok.</p>
Thanks, this looks ok.
Ship it!


----

----

----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1170/

PR #1891, opened 2015-04-09
Indentation in PrintHelp() for better reading flow

darealshinji:
I think it's easier to read it this way.
By the way, in line 410 of welsenc.cpp is `(Level)` correct or shouldn't it be `(Layer)`?


sijchen:
line 410: this is the level of trace. see "Enumerate the type of wels log" in codec_app_def.h.

PR #1893, opened 2015-04-10
bug fix for nodelay API for buffer storage

huili2:
save necessary dst buffer as well.
see:
https://rbcommons.com/s/OpenH264/r/1172/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1172/

----
**testing_done**:  
**description**:  <p>Old API does not save the pointer "ppDst" for output buffer. Once the second call DecodeFrame2() returns NULL, it will overwrite the first ppDst.<br />This will return NULL pointer for output buffer, which results in missing output frames.</p>

----
 reviewer Haibo Zhu (haibozhu) 
 time 2015-04-10T08:37:45+00:00 
2015-04-10T08:37:45+00:00 Haibo Zhu (haibozhu):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1172/

PR #1899, opened 2015-04-17
add ut demo for wp8

shihuade:
1. change make file to generate ut dll files
2. add ut demo and runtime component project
3. gtest with latest version can now be used for windows phone 
4. review board:
    https://rbcommons.com/s/OpenH264/r/1176/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1176/

----
**testing_done**:  <p>error with latest  gtest-1.7.0 list as below when build with make file</p><p>cl -Od -Oy- -Zi -RTC1 -D_DEBUG -MDd -DHAVE_NEON -nologo -Fdopenh264.pdb -W3 -EHsc -fp:precise -Zc:wchar_t -Zc:forScope -D_VARIADIC_MAX=10 -DWINAPI_FAMILY=WINAPI_FAMILY_PHONE_APP -DWINDOWS_PHONE -DDEBUG -DGENERATED_VERSION_HEADER -ZW -wd4447 -I./codec/api/svc -I./codec/common/inc -D_CRT_SECURE_NO_WARNINGS -I./gtest/include -I./codec/common/inc -I./test -I./codec/decoder/core/inc -I./codec/decoder/plus/inc -c -Fotest/decoder/DecUT_Deblock.obj test/decoder/DecUT_Deblock.cpp<br />DecUT_Deblock.cpp<br />E:\Project\Openh264\Huade-04-16\gtest\include\gtest/internal/gtest-port.h(1770) : error C3861: 'chdir': identifier not found<br />E:\Project\Openh264\Huade-04-16\gtest\include\gtest/internal/gtest-port.h(1802) : error C3861: 'getenv': identifier not found</p>
**description**:  <p>add codec ut demo (based on gtest version 690)<br />   --update make file to build ut dll for wp8 demo<br />   --add ut.def for dll export<br />   --UT demo project<br />   --gtest with latest gtest-1.7.0 still have build error. but version with 690 is OK.<br />to do:<br />  issue caused by using latest gtest-1.7.0, there are build error for building ut related source files.</p>

----
 time 2015-04-16T10:25:13+00:00 
2015-04-16T10:25:13+00:00 Martin Storsj (mstorsjo):
><p>The rest of it looks good. Unfortunately, it's impossible to make this work with gtest 1.7.0 - that release is from 2013 and didn't have any support for windows phone.</p>
The rest of it looks good. Unfortunately, it's impossible to make this work with gtest 1.7.0 - that release is from 2013 and didn't have any support for windows phone.

2015-04-17T04:58:02+00:00 Forrest Shi (huade):
><p>update:<br />I thought that gtest-1.7.0 was the latest one which is incorrect.<br />And now I clone the latest gtest, it works!<br />Thanks!</p>
so it means that the latest gtest does not support for windows phone?
update:<br />I thought that gtest-1.7.0 was the latest one which is incorrect.<br />And now I clone the latest gtest, it works!<br />Thanks!

2015-04-17T04:58:02+00:00 Forrest Shi (huade):
 on **Makefile line196**: 
 id="comment235727"><p>If you'd use $(LIBPREFIX)ut.$(LIBSUFFIX) (that is, ut.lib instead of ut.dll), you would have a bit simpler code (no need for LoadPackagedLibrary) and would require a bit less modifications (it wouldn't require ut.def at all either).</p><p>I would prefer it that way but I won't stop you if you really prefer it this way.</p>
If you'd use $(LIBPREFIX)ut.$(LIBSUFFIX) (that is, ut.lib instead of ut.dll), you would have a bit simpler code (no need for LoadPackagedLibrary) and would require a bit less modifications (it wouldn't require ut.def at all either).
I would prefer it that way but I won't stop you if you really prefer it this way.

2015-04-17T06:07:29+00:00 Forrest Shi (huade):
><p>OK,then keep dll as final solution. <br />Thanks :)</p>
Yes.I tried, but it did not work.<br />----tried 1.<br />    -- ut.lib contain all ut related object files( under /test/<strong> and /codec  and gtest/*)<br />    -- and RuntimeComponent project link to<br />           ut.lib<br />    -- extern int CodecUtMain (int argc , char</strong> argv) <br />        in source code Codec_UT_RTComponent.cpp under RuntimeComponent project
----tried 2.<br />    -- ut.lib only contain  ut related object files( only under /test/<strong> )<br />    -- and RuntimeComponent project link to<br />            ut.lib  gtest.lib and openh264.lib<br />    -- extern int CodecUtMain (int argc , char</strong> argv) <br />        in source code Codec_UT_RTComponent.cpp under RuntimeComponent project
and both above solutions did not work.<br />the final result when running the ut is,  there is zero cases showed in the ut result xml file.
Oh, I see - there's probably nothing that forces the individual object files from the static library to be linked in. (With gcc you can use the -Wl,--whole-library option, but it's a bit harder for MSVC.) I don't mind keeping this as a DLL then.
OK,then keep dll as final solution. <br />Thanks :)

2015-04-17T06:07:29+00:00 Forrest Shi (huade):
 on **build/platform-msvc-wp.mk line5**: 
 id="comment235726"><p>Don't add -DDEBUG here</p>
Don't add -DDEBUG here

2015-04-17T01:24:52+00:00 Forrest Shi (huade):
><p>thanks, forgot to remove it when I used this macro for test</p>
thanks, forgot to remove it when I used this macro for test


----

----
2015-04-17T05:42:53+00:00 status-change:
add runtime project <br />    ----project files has been filtered out by<br />         build/windowsphone/.gitignore
update  build/windowsphone/.gitignore
update auto script for copying ut files moudle


----
 time 2015-04-17T05:54:30+00:00 
2015-04-17T05:54:30+00:00 Martin Storsj (mstorsjo):
><p>This looks ok to me.</p>
This looks ok to me.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1176/

PR #1900, opened 2015-04-17
Hook up the x86_32-only assembly as well

mstorsjo:
This makes sure the code is at least somewhat tested. Ideally those
assembly functions should be upgraded to work in 64 bit mode as well,
but until then it's better to actually use them in the configurations
where they can be built.

Review at https://rbcommons.com/s/OpenH264/r/1177/.

### DIFF
--- a/build/x86-common.mk
+++ b/build/x86-common.mk
@@ -17,6 +17,9 @@ ASMFLAGS_PLATFORM = -DX86_32
 endif
 ifeq ($(USE_ASM),Yes)
 CFLAGS += -DX86_ASM
+ifneq ($(ENABLE64BIT), Yes)
+CFLAGS += -DX86_32_ASM
+endif
 ASM_ARCH = x86
 endif
 ASM = nasm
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1177/

----
**testing_done**:  
**description**:  Hook up the x86_32-only assembly as wellThis makes sure the code is at least somewhat tested. Ideally thoseassembly functions should be upgraded to work in 64 bit mode as well,but until then it&#39;s better to actually use them in the configurationswhere they can be built.

----
 time 2015-04-20T01:42:20+00:00 
2015-04-20T01:42:20+00:00 zhiliang wang (zhilwang):
>Ship It!
Ship It!
Ship it!


----
 reviewer Forrest Shi (huade) 
 time 2015-04-20T01:46:35+00:00 
2015-04-20T01:46:35+00:00 Forrest Shi (huade):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1177/

PR #1909, opened 2015-04-21
Mention the "all" target before the extra dependencies for openh264.dll in msvc.mk

mstorsjo:
This makes sure that all components get built, not only the DLL,
if doing a plain "make OS=msvc" after 2b33c749939.

Review at https://rbcommons.com/s/OpenH264/r/1185/.

### DIFF
--- a/build/platform-msvc.mk
+++ b/build/platform-msvc.mk
@@ -3,4 +3,7 @@ LDFLAGS += user32.lib
 CFLAGS_OPT += -MT
 CFLAGS_DEBUG += -MTd -Gm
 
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1185/

----
**testing_done**:  
**description**:  Mention the &quot;all&quot; target before the extra dependencies for openh264.dll in msvc.mkThis makes sure that all components get built, not only the DLL,if doing a plain &quot;make OS=msvc&quot; after 2b33c749939.

----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1185/

PR #1922, opened 2015-04-30
Disable HP tools in decoder

HaiboZhu:


HaiboZhu:
Some code need to be astyled

PR #1958, opened 2015-05-25
Fix install targets for out-of-tree builds.

ghismary:


mstorsjo:
Thanks for fixing this! The extra $(SRC_PATH) isn't needed though, see #1967.

PR #1965, opened 2015-05-28
Fix source files path for out-of-tree build.

ghismary:


ghismary:
Please ignore this pull request, this is not correct. I will provide an other pull request to fix this correctly.

PR #1966, opened 2015-05-28
Fix source files path for out-of-tree build.

ghismary:


mstorsjo:
Can you:

1. Explain what issue there is that you're trying to fix. I see that you're generating the version header in the build tree instead of in the source tree, that's a useful change, but the rest of it I don't immediately see what issue you're having and trying to fix..
2. Don't revert your own changes on the branch, but rebase and remove the commits that you don't want to include (or create a completely new branch and pull request if you're unable to update this one properly)
3. Don't manually edit targets.mk, they're autogenerated - you need to change the generator to make sure the changes aren't accidentally removed later.
4. Don't manually add `$(SRC_PATH)` to the build rules - the `vpath` entries in `Makefile` should take care of that, no?

mstorsjo:
The changes for adding $(SRC_PATH) for the .def files probably is correct though, but I don't understand why you're editing all the targets.mk files.

ghismary:
I am trying to build out-of-tree because I'm building for several platforms with the same source tree in parallel.
I didn't know the targets.mk were autogenerated, thanks for pointing that.
You are right the vpath entries in Makefile indeed takes care of that. Except in a specific case: when building out-of-tree on Windows (with MSVC) giving the absolute path to the Makefile, eg. make -f C:/code/openh264/Makefile BUILDTYPE="Release" OS="msvc"
When putting $(SRC_PATH) in the targets.mk it is working correctly in all the cases.

mstorsjo:
Ok, thanks for explaining. You are building in some variant of msys/msys2/cygwin shell, right? I would expect that it would work if you'd use a relative path to the openh264 Makefile. Does it work right (without manually adding all the extra $(SRC_PATH)) if you use an absolute unix-style path instead? At least msys does a lot of work to convert the unix style paths back to windows, when calling windows binaries, so I think that would work.

I'd be very reluctant to add these extra $(SRC_PATH) just because vpath doesn't work with absolute windows style paths - I would think that's a pretty unsupported usecase, other things in make might fail just as well then.

ghismary:
Yes it works when putting a relative path. I think I would be able to stick with it. I will also try to see if this works with a unix-style path.
Then, let's close this pull request and I'll provide a new one with the changes to generate the version header in the build tree, add $(SRC_PATH) for the .def files and also generate the .res file in the build tree instead of the source tree.

ghismary:
Ok, just to let you know, this does not work with unix-style paths because the linker expects a dos-style path when providing the .def file.

mstorsjo:
Ah, I see.

The thing is, normally msys does such path translations automatically; if you provide the option `-I/absolute/path/to/include` to cl.exe, msys will notice that cl.exe is a non-msys binary and will translate the command argument into `-Ic:\absolute\path\to\include`. This only works as long as the option is like `-X/path`, i.e. only a single char after the dash - it can't figure out to do this when doing e.g. `-Fo/absolute/path` or `-def:/absolute/path`.

Recent cl.exe versions actually can take `-Fo /absolute/path` as two separate arguments, which I think was done to help fix this issue, but I'm not sure if one can do `-def /path` instead of `-def:/path`.

But if it works with relative paths, that's probably good enough for now, without having to change all of the makefiles just for this.

ghismary:
I have added a new pull-request with the changes I was talking about: #1969

mstorsjo:
I guess this pull request should be closed now?

ghismary:
This is ok for me to close it now.

mstorsjo:
I think that you (as opener of the pull request) can close it as well, instead of waiting for someone else to do it.

PR #1969, opened 2015-05-29
More fixes for out-of-tree build:

ghismary:
- Generate version header in the build tree
 - Generate .res file in the build tree on Windows
 - Add the $(SRC_PATH) prefix to the .def files on Windows

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,7 +87,7 @@ ifneq ($(V),Yes)
 endif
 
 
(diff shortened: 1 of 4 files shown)


mstorsjo:
Except for the detail about `//fo` vs `-fo`, these changes look good to me.

sijchen:
@ghismary could you kindly add a commit to fix the detail as mentioned above in comments? Thanks a lot.

ghismary:
I have added the commit to replace the /fo option by -fo.

sijchen:
@ghismary thanks a lot.

PR #1977, opened 2015-06-03
Fix race condition by dispatching Enc/Decode complete.

ethanhugg:
This should prevent the caller from deleting while a encode or decode is in flight.

Review is here:
https://rbcommons.com/s/OpenH264/r/1227/

### DIFF
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -363,7 +363,10 @@ class OpenH264VideoEncoder : public GMPVideoEncoder {
   }
 
   virtual void EncodingComplete() {
-    delete this;
+    // Synchronously delete on main thread.
+    g_platform_api->syncrunonmainthread (WrapTask (
+      this,
+      &OpenH264VideoEncoder::EncodingComplete_m));
   }
 
  private:
@@ -540,6 +543,9 @@ class OpenH264VideoEncoder : public GMPVideoEncoder {
     frame->Destroy();
   }
 
+  void EncodingComplete_m() {
+    delete this;
+  }
 
  private:
   GMPVideoHost* host_;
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1227/

----
**testing_done**:  <p>Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1164740<br />Race condition found between Decode_m and DecodingComplete.</p>
**description**:  Fix race condition by dispatching Enc/Decode complete.

----
 reviewer Sijia Chen (sijchen) 
 time 2015-06-02T02:26:30+00:00 
2015-06-02T02:26:30+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!

2015-06-03T22:56:05+00:00 Ethan Hugg (ethanhugg):
><p>Do the sync dispatches your adding here actually cause the tasks to be run? In Firefox's syncrunonmainthread implementation, we fail and return an error if this is called on the main thread, and DecodingComplete() is supposed to be called on the main thread, so your sync dispatch to the main thread should be failing here and not running the task.</p><p>Also, sync dispatches in general are risky, as they can block tasks in the calling thread's queue from running that will unblock the other thread to which you're dispatching to, i.e. they are a source of deadlocks. That's may not be possible here, but it bit both us and Adobe with our EME GMPs, and for future proofing your better to avoid them.</p>
I did not submit a pull request yet for this.   Still discussing with Mozilla folks over whether this fix is needed.
understand
Do the sync dispatches your adding here actually cause the tasks to be run? In Firefox's syncrunonmainthread implementation, we fail and return an error if this is called on the main thread, and DecodingComplete() is supposed to be called on the main thread, so your sync dispatch to the main thread should be failing here and not running the task.
Also, sync dispatches in general are risky, as they can block tasks in the calling thread's queue from running that will unblock the other thread to which you're dispatching to, i.e. they are a source of deadlocks. That's may not be possible here, but it bit both us and Adobe with our EME GMPs, and for future proofing your better to avoid them.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1227/

PR #1997, opened 2015-06-18
Enable deterministic mode for ar

deadNightTiger:
Note that this is using GNU ar option which is not in POSIX. If this isn't okay, I'll strip stuff manually in a shell script.


sijchen:
@ethanhugg  it looks okay?

ethanhugg:
I'm okay with the D option, but as-is, this will break the OSX builds where ar does not have it.  Perhaps add it in platform-linux instead.

mstorsjo:
Don't include merge commits in your pull requests; rebase this on top of the latest master instead of merging master into your branch.

sijchen:
@deadNightTiger could you please change the current one or create another request according to the comments? thanks

sijchen:
@deadNightTiger , interestingly that my Windows build will fail with the below error after I merged this pull request, the build can succeed with commit c51975acdf73ad6f6c7cb708056f865c6f79e99a (right before merging this pull request), and will fail with commit a4fb978bb74817e49e5f42166fef4e1fc61ef141 (right after the merging this pull request)
**Any suggestions? I may have to create a pull request to revert this merging if we can't find a solution to the build failure.** Thanks in advance.

lib -nologo -out:common.lib common.lib codec/common/src/common_tables.obj codec/common/src/copy_mb.obj codec/common/src/cpu.obj codec/common/src/crt_util_safe_x.obj codec/common/src/deblocking_common.obj codec/common/src/expand_pic.obj codec/common/src/intra_pred_common.obj codec/common/src/mc.obj codec/common/src/memory_align.obj codec/common/src/sad_common.obj codec/common/src/utils.obj codec/common/src/welsCodecTrace.obj codec/common/src/WelsThreadLib.obj
LINK : fatal error LNK1104: cannot open file 'common.lib'
make: *** [common.lib] Error 80
cp: cannot stat `openh264.dll': No such file or directory

mstorsjo:
Sijia, I can fix this, I'll send a pull request.

The issue is that you can't append to ASFLAGS like this - the reason why $@ is in the variable is because for MSVC, you need to have it in the exact form `-out:$@`.

mstorsjo:
There's a proper fix in #2011.

sijchen:
@mstorsjo Thanks a lot!

mstorsjo:
FYI, I guess you can remove the branch "revert-1997-deterministic-build" from the repo, it doesn't serve any purpose any longer.

PR #1999, opened 2015-06-19
Make the static library directory configurable

bmiklautz:
When running "make install" the destination of pkgconfig/openh264.pc and
lib/libopenh264.a was hard coded to "lib". When building packages it
might be required to change the install location of those files to
the systems default library location (like lib64 or lib/x86_64-linux-gnu
for example).

This commit adds an additional make variable "LIBDIR_NAME" that allows
it to change the destination directory for those files.
Without setting LIBDIR_NAME the installation location is the same as before
(lib).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ PREFIX=/usr/local
 SHARED=-shared
 OBJ=o
 DESTDIR=
+LIBDIR_NAME=lib
 SHAREDLIB_DIR=$(PREFIX)/lib
 PROJECT_NAME=openh264
 MODULE_NAME=gmpopenh264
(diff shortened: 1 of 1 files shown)


mstorsjo:
This looks ok to me.

nasa1286:
261797_174854645912351_8143515_n
1488092_10151860650971658_749218915_n

nasa1286:
1388133_10151695458816658_1265435350_n
1388007_10151695541021658_662115846_n
1384744_10151695458811658_57917410_n
1082778_10151586077741658_1801127914_n
1006222_552394444825034_589148037_n

PR #2008, opened 2015-06-25
    use the external qp setting for screen content

ruil2:


sijchen:
context ok, but could the commit with CI failure be removed? Thx

PR #2011, opened 2015-06-26
Don't concatenate $@ to AR_OPTS

mstorsjo:
The MSVC version of AR_OPTS already contains $@, in a very specific
way.

This fixes MSVC builds after c8bed8a9.


deadNightTiger:
I don't think that works for OS X now, you probably forgot to change default AR_OPTS back to `cr $@` as it was before.

mstorsjo:
Oh, right, of course - updated.

PR #2019, opened 2015-07-02
put pSliceBsBuffer to thread buffer rather than per slice, so as to save memory usage

sijchen:
https://rbcommons.com/s/OpenH264/r/1252/

put pSliceBsBuffer to thread buffer rather than per slice, so as to save memory usage
in some test, the change can save memory usage by 30% under SM_DYN_SLICE

### DIFF
--- a/codec/encoder/core/inc/mt_defs.h
+++ b/codec/encoder/core/inc/mt_defs.h
@@ -87,6 +87,7 @@ int32_t*                        pSliceComplexRatio[MAX_DEPENDENCY_LAYER]; // *IN
 FILE*                           pFSliceDiff;    // file handle for debug
 #endif//MT_DEBUG
 
+uint8_t*                        pThreadBsBuffer[MAX_THREADS_NUM]; //actual memory for slice buffer
 } SSliceThreading;
 
 #endif//MULTIPLE_THREADING_DEFINES_H__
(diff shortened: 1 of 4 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1252/

----
**testing_done**:  
**description**:  <p>put pSliceBsBuffer to thread buffer rather than per slice, so as to save memory usage<br />in some test, the change can save memory usage by 30% under SM_DYN_SLICE</p>

----
 reviewer wayne liu (huili2) 
 time 2015-07-02T07:17:12+00:00 
2015-07-02T07:17:12+00:00 wayne liu (huili2):
 on **codec/encoder/core/src/slice_multi_threading.cpp line298**: 
 id="comment264641"><ol><li>should be "iSlice0Len".</li><li>move line 298, line 310 into top row in this func to prevent redundant memory alloc.</li></ol>

2015-07-02T07:44:39+00:00 Sijia Chen (sijchen):
><p>yeah, 2 is not an issue, but early check the input parameters is a good habit, and can save some unnecessary operations inside.</p>
will fix 1; for 2, when the return is not SUCCESS, there is freememeory in which it will free all !=NULL pointers so there will be no proble,
yeah, 2 is not an issue, but early check the input parameters is a good habit, and can save some unnecessary operations inside.


----

----
 reviewer wayne liu (huili2) 
 time 2015-07-03T03:16:29+00:00 
2015-07-03T03:16:29+00:00 wayne liu (huili2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1252/

PR #2020, opened 2015-07-03
Allow compilation to JavaScript via emscripten

brion:
Currently the emscripten JavaScript environment for C/C++ programs cannot do multithreading, and doesn't have a sysctl implementation to get hardware CPU count. Hardcode to 1.

Builds using the Linux platform settings; to build with emscripten (tested 1.30.0):

  emmake make OS=linux ARCH=asmjs

The resulting libopenh264.so is LLVM bitcode which can be further linked into an emscripten project for final JavaScript output.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -30,6 +30,7 @@ Adam Roach
 Sawyer Shan
 Siping Tao
 Martin Storsjö
+Brion Vibber
 James Wang
 Juanny Wang
 Zhiliang Wang
--- a/codec/common/src/WelsThreadLib.cpp
+++ b/codec/common/src/WelsThreadLib.cpp
@@ -483,6 +483,12 @@ WELS_THREAD_ERROR_CODE    WelsQueryLogicalProcessInfo (WelsLogicalProcessInfo* p
 
   return WELS_THREAD_ERROR_OK;
 
+#elif defined(__EMSCRIPTEN__)
+
+  // There is not yet a way to determine CPU count in emscripten JS environment.
+  pInfo->ProcessorCount = 1;
+  return WELS_THREAD_ERROR_OK;
+
 #else
 
   size_t len = sizeof (pInfo->ProcessorCount);


sijchen:
@ethanhugg

sijchen:
quite curious how the performance will be with the JavaScript version. There was some attempts we made with this direction, but the performance reports were not attractive.

brion:
I haven't benchmarked it yet, but if it's comparable to libvpx for VP8 it should achieve acceptable playback at SD resolutions on reasonably fast computers with current-version browsers. I get much better performance on Ogg Theora, which is an older-generation codec.

See [demos of Ogg and WebM JS playback](https://brionv.com/misc/ogv.js/demo/); I haven't finished the MP4 bits yet but have confirmed emscripten'd OpenH264 decoding in isolation is working via some [crazy Mozilla patches](https://bugzilla.mozilla.org/show_bug.cgi?id=1165180).

Making use of SIMD extensions for JavaScript should help both libvpx and OpenH264, but that's still experimental in the browsers and will take some fiddling with vector intrinsics on the C/C++ side.

(Most browsers support MP4/H.264 via system, hardware, or bundled codecs these days so I'm actually most interested in JavaScript OpenH264 for encoding in browser-side tasks such as animation widgets.)

mstorsjo:
The change looks acceptable, but just out of curiosity - is there anything that strictly needs to be different from the linux makefile? I.e., would an `emmake make OS=linux ARCH=asmjs` right now work as well, or does it fail due to `-lpthread` or `-fPIC`? You might be interested in adding `-MMD -MP` as well to get dependency generation/tracking (if that makes sense at all for emscripten builds - I'm not familiar with it).

brion:
Confirmed the Linux platform settings seem to work without exploding; dropping the extra platform file and cutting the patch down to just the thread count fallback. :)

mstorsjo:
Thanks! You probably want to update the commit message at the same time though :-)

ethanhugg:
This looks fine.  Glad to hear it works.

PR #2029, opened 2015-07-13
Remove unnecessary tabs in the middle of lines

mstorsjo:
At 6ff4ef3b38 all the (stray, mixed) tabs had been cleaned out from
the source code; remove newly added tabs and convert to spaces.

Review at https://rbcommons.com/s/OpenH264/r/1264/.

### DIFF
--- a/codec/encoder/core/inc/rc.h
+++ b/codec/encoder/core/inc/rc.h
@@ -272,7 +272,7 @@ PWelsCheckFrameSkipBasedMaxbrFunc pfWelsCheckSkipBasedMaxbr;
 PWelsUpdateBufferWhenFrameSkippedFunc pfWelsUpdateBufferWhenSkip;
 PWelsUpdateMaxBrCheckWindowStatusFunc pfWelsUpdateMaxBrWindowStatus;
 
-PWelsRCPostFrameSkippingFunc			pfWelsRcPostFrameSkipping;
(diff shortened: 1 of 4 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1264/

----
**testing_done**:  
**description**:  Remove unnecessary tabs in the middle of linesAt 6ff4ef3b38 all the (stray, mixed) tabs had been cleaned out fromthe source code; remove newly added tabs and convert to spaces.

----
 reviewer Sijia Chen (sijchen) 
 time 2015-07-14T01:29:52+00:00 
2015-07-14T01:29:52+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!

2015-07-14T07:05:20+00:00 Sijia Chen (sijchen):
><p>That's probably the best solution, I'll send a review request for that. (I originally didn't use that when I cleaned up the tabs, since it would have ruined the formatting in some cases where the spacing needs to be done manually.)</p>
@Martin, any change needed for astyle setting/version to avoid similar problems? seems the astyle we are using now did not auto-correct those you mentioned. Thanks in advance.
That's probably the best solution, I'll send a review request for that. (I originally didn't use that when I cleaned up the tabs, since it would have ruined the formatting in some cases where the spacing needs to be done manually.)


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1264/

PR #2061, opened 2015-08-04
Cj build1

8v060htwyc:


build/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+RUN tar xvfz nasm-2.11.06.tar.gz 
+WORKDIR /tmp/nasm-2.11.06/
+RUN ./configure
+RUN make
+RUN make install 
+
+WORKDIR /tmp
+RUN git clone https://github.com/cisco/openh264.git
+WORKDIR /tmp/openh264
+RUN git checkout v1.1
```
sijchen: intentionally checkout v1.1 here -- is this needed to be updated with newer versions or not?

shihuade:
1. what about cent-OS, apt-get should be replace with yum command, is it possible that automatically change  command like apt-get to yum?

2. for below command, the update may take such a long time to update the entire packages, 
       RUN apt-get update
       RUN apt-get upgrade -y
       RUN apt-get install -y \

shihuade:
for below RUN instruction, it seems that it is not follow the best practices guid  which  you can refer to https://docs.docker.com/articles/dockerfile_best-practices/

+RUN apt-get update
+RUN apt-get upgrade -y
+RUN apt-get install -y bison flex g++ gcc git libgmp3-dev libmpc-dev libmpfr-dev libz-dev make wget
+

shihuade:
1. There are some errors when I used an totally new Ubuntu machine with  version(14.04 or 15.04).
  it seems that blocked by below instruction:
   RUN apt-get install -y bison flex g++ gcc git libgmp3-dev libmpc-dev libmpfr-dev libz-dev make wget
 but when only used apt-get outside the dockerfile, everything is OK.

***********************************************************************************
The following packages have unmet dependencies:
 g++ : Depends: cpp (>= 4:4.8.2-1ubuntu6) but it is not going to be installed
       Depends: g++-4.8 (>= 4.8.2-5~) but it is not going to be installed
       Depends: gcc-4.8 (>= 4.8.2-5~) but it is not going to be installed
 gcc : Depends: cpp (>= 4:4.8.2-1ubuntu6) but it is not going to be installed
       Depends: gcc-4.8 (>= 4.8.2-5~) but it is not going to be installed
       Recommends: libc6-dev but it is not going to be installed or
                   libc-dev
 zlib1g-dev : Depends: libc6-dev but it is not going to be installed or
                       libc-dev
E: Unable to correct problems, you have held broken packages.
***********************************************************************************

HaiboZhu:
It is too complex for our codec, thank you for your contribution

PR #2063, opened 2015-08-05
Add protection for memcpy overlap

HaiboZhu:
See code review at: https://www.rbcommons.com/s/OpenH264/r/1281/

### DIFF
--- a/codec/decoder/core/src/manage_dec_ref.cpp
+++ b/codec/decoder/core/src/manage_dec_ref.cpp
@@ -131,14 +131,16 @@ int32_t WelsInitRefList (PWelsDecoderContext pCtx, int32_t iPoc) {
         bCopyPrevious = bCopyPrevious && (pRef->iWidthInPixel == pCtx->pPreviousDecodedPictureInDpb->iWidthInPixel)
                         && (pRef->iHeightInPixel == pCtx->pPreviousDecodedPictureInDpb->iHeightInPixel);
 
-        if (bCopyPrevious) {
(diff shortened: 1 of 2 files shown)


codec/decoder/core/src/manage_dec_ref.cpp
```diff
@@ -131,14 +131,16 @@ int32_t WelsInitRefList (PWelsDecoderContext pCtx, int32_t iPoc) {
...
-        } else {
+        if (!bCopyPrevious) {
           memset (pRef->pData[0], 128, pRef->iLinesize[0] * pRef->iHeightInPixel);
           memset (pRef->pData[1], 128, pRef->iLinesize[1] * pRef->iHeightInPixel / 2);
           memset (pRef->pData[2], 128, pRef->iLinesize[2] * pRef->iHeightInPixel / 2);
+        } else if (pRef == pCtx->pPreviousDecodedPictureInDpb) {
+          WelsLog (& (pCtx->sLogCtx), WELS_LOG_WARNING, "WelsInitRefList()::EC memcpy overlap.");
+        } else {
+          memcpy (pRef->pData[0], pCtx->pPreviousDecodedPictureInDpb->pData[0], pRef->iLinesize[0] * pRef->iHeightInPixel);
+          memcpy (pRef->pData[1], pCtx->pPreviousDecodedPictureInDpb->pData[1], pRef->iLinesize[1] * pRef->iHeightInPixel / 2);
```
licaiguo: I think you can use memmove to replace memcpy

HaiboZhu:
In fact, the pSrc & pDst should not have the same address. Now the overlap is caused by the design of EC behavior about NULL picture. If we re-design it, we do not need this protection.

PR #2072, opened 2015-08-13
[Encoder] fix the range of writable cfg to avoid problem

sijchen:
https://rbcommons.com/s/OpenH264/r/1285/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1285/

----
**testing_done**:  
**description**:  fix the range of writable cfg to avoid problem

----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1285/

PR #2082, opened 2015-08-26
Force reset decoder when out of memory error

HaiboZhu:
See code review at: https://www.rbcommons.com/s/OpenH264/r/1299/
When out of memory error, force reset the decoder


HaiboZhu:
The current trace point has only been protected outside

PR #2088, opened 2015-08-27
replace LayerBsInfo with FrameBsInfo in thread-based private data str…

shihuade:
https://rbcommons.com/s/OpenH264/r/1295/

----for more easy operation on FrameBS data during multiple-thread encoding, replace
     LayerBsInfo with FrameBsInfo in thread-private data structure
----LayerBsInfo is part of FrameBsInfo with plus iLayerBsIndex in Ctx-pOut;

----it is for later dynamic re-allocate function in multi-thread encoding under slice mode 4

### DIFF
--- a/codec/encoder/core/inc/extern.h
+++ b/codec/encoder/core/inc/extern.h
@@ -112,6 +112,7 @@ void WelsEncoderApplyFrameRate (SWelsSvcCodingParam* pParam);
 int32_t WelsEncoderApplyBitRate (SLogContext* pLogCtx, SWelsSvcCodingParam* pParam, int32_t iLayer);
 int32_t WelsEncoderApplyBitVaryRang(SLogContext* pLogCtx, SWelsSvcCodingParam* pParam, int32_t iRang);
 int32_t WelsEncoderApplyLTR (SLogContext* pLogCtx, sWelsEncCtx** ppCtx, SLTRConfig* pLTRValue);
+int32_t DynSliceRealloc(sWelsEncCtx* pCtx,SFrameBSInfo* pFrameBsInfo,SLayerBSInfo* pLayerBsInfo);
 int32_t FilterLTRRecoveryRequest (sWelsEncCtx* pCtx, SLTRRecoverRequest* pLTRRecoverRequest);
 void CheckProfileSetting (SLogContext* pLogCtx,SWelsSvcCodingParam* pParam,int32_t iLayer, EProfileIdc uiProfileIdc);
 void CheckLevelSetting (SLogContext* pLogCtx,SWelsSvcCodingParam* pParam,int32_t iLayer, ELevelIdc uiLevelIdc);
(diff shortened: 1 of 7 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1295/

----
**testing_done**:  <p>----for more easy operation on FrameBS data during multiple-thread encoding, replace<br />     LayerBsInfo with FrameBsInfo in thread-private data stucture<br />----LayerBsInfo is part if FrameBsInfo with plus iLayerBsIndex in Ctx-pOut;</p><p>----it is for later dynamic re-allocate function in multi-thread encoding under slice mode 4</p><p>----and merge previous bug fixed which can refer to https://rbcommons.com/s/OpenH264/r/1291/</p>
**description**:  replace LayerBs with FrameBs

----
 reviewer Sijia Chen (sijchen) 
 time 2015-08-26T23:51:27+00:00 
2015-08-26T23:51:27+00:00 Sijia Chen (sijchen):
><p>I kind of not sure if we are moving toward different directions.... hopefulling not. Are you planning to do buffer reallocating for the <em>whole</em> frame bs, in a child-encoding thread?</p>
I kind of not sure if we are moving toward different directions.... hopefulling not. Are you planning to do buffer reallocating for the <em>whole</em> frame bs, in a child-encoding thread?

2015-08-28T04:51:52+00:00 Forrest Shi (huade):
><p>the first code review I sent include a variable of list, which may cause build problem as noted by Martin, so I have to fix it first. Same as the sleep on Winphone.</p>
yes, step by step.<br />it seems there are lots of thing to do.<br />will based on your thread-pool code merge
It may not be a good idea to reallocate the whole frame bs buffer in the child-encoding thread since that may mean in one thread, there is a possibility to modify the context of the other threads. The threads should be as independent as possible. By design such cross-thread modification should be avoided.
and if we need to rellocate many times in a child thread, this way may mean copying encoded data for many times.
Thinking quickly, my proposal will be we create one set of target buffer for each thread, and re-allocate that if needed, and we can instead use memcpy to put data from thread'tartgget buffer to frame bs buffer. Regarding that we are already doing memmove, there is little increased speed burden.
yes, it is a good idea to do so. so waite for your code merge.<br />the data structure also need to change.<br />and here in this review, FrameBs used to replace LayerBs, re-allocated is part of this change. Additionally, FrameBs gives more info than LayerBs.
Actually you can do part of it now ("create one set of target buffer for each thread, and re-allocate that if needed"), without needing to wait for my changes. Check the existing "pThreadBsBuffer", which is used for intermediate writing for now. 
According to Martin's suggestion, I still need to add implementation for a "list" structrure, and find a substitution for "sleep" on WinPhone, before I can go on merging the current changes. Unfortunately this cannot be done very soon during to other work at hand....
got it.<br />but it seems that there are no logical change based on your thread-pool,it is independence. if so, I suggest you merge it and refine the small detail step by step later
the first code review I sent include a variable of list, which may cause build problem as noted by Martin, so I have to fix it first. Same as the sleep on Winphone.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1295/

PR #2099, opened 2015-09-02
Use stdbool.h with Visual Studio 2013 or higher

nathankidd:
2013 is the first version to include the header.

Avoids "codec_api.h(40): error C2632: 'char' followed by 'bool' is illegal"
when building C code.

### DIFF
--- a/codec/api/svc/codec_api.h
+++ b/codec/api/svc/codec_api.h
@@ -36,7 +36,7 @@
 #define WELS_VIDEO_CODEC_SVC_API_H__
 
 #ifndef __cplusplus
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && (_MSC_VER < 1800)
 typedef unsigned char bool;
 #else
 #include <stdbool.h>


mstorsjo:
This probably is ok, assuming that the full testsuite still passes.

nathankidd:
*make test* says:
> ...
> [----------] Global test environment tear-down
> [==========] 447 tests from 53 test cases ran. (35703 ms total)
> [  PASSED  ] 447 tests.

Is that what "full testsuite" means? Or something else?

mstorsjo:
Yes, that's it. In particular, there's a test that makes sure the C ABI still works, although I'm not sure it tests the size of bool. Is sizeof(bool) still equal in both C and C++ after this?

nathankidd:
Yes, bool size is 1 everywhere.

```
$ cat booltest.c
#include <stdio.h>
#include <stdbool.h>

int main (void)
{
                printf("sizeof(bool)=%d\n", sizeof(bool));
                return 0;
}

$ cl booltest.c
Microsoft (R) C/C++ Optimizing Compiler Version 18.00.40629 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

booltest.c
Microsoft (R) Incremental Linker Version 12.00.40629.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:booltest.exe
booltest.obj

$ ./booltest
sizeof(bool)=1
```

and

```
$ cat booltest.cpp
#include <iostream>
#include <stdbool.h> // same result if not included

int main (void)
{
                std::cout << "sizeof(bool)=" << sizeof(bool) << "\n";
                return 0;
}

$ cl booltest.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 18.00.40629 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

booltest.cpp
Microsoft (R) Incremental Linker Version 12.00.40629.0
Copyright (C) Microsoft Corporation.  All rights reserved.

/out:booltest.exe
booltest.obj

$ ./booltest
sizeof(bool)=1
```

mstorsjo:
Ok, great, then this should absolutely be ok. (I don't have privileges to merge it myself though.)

nathankidd:
@mstorsjo Thanks for the review.

HaiboZhu:
Thank you for your contribution for openh264 project

nathankidd:
Thanks for merging.

PR #2104, opened 2015-09-08
  exception case for width or height is less than 16

ruil2:


mstorsjo:
This breaks a lot of tests, I've seen test errors in ParameterSetStrategy_SPS_PPS_LISTING3, SimulcastAVC, SimulcastAVC_SPS_PPS_LISTING and probably a lot of other tests. This causes spurious errors for others' pull requests, such as in #2115.

PR #2112, opened 2015-09-12
Revert "Enable deterministic mode for ar"

8v060htwyc:


sijchen:
Hi pls give reason on the reverting request.

HaiboZhu:
It seems no sense about it

PR #2113, opened 2015-09-12
Support NetBSD and OpenBSD; consolidate BSD build.

fstd:
This adds support for NetBSD and OpenBSD by consolidating platform-freebsd.mk to platform-bsd.mk.
I verified that it still builds on FreeBSD.

Cheers,
Timo Buhrmester


fstd:
This pull request is flawed, I'll reopen with a correct version.

PR #2115, opened 2015-09-12
Support NetBSD and OpenBSD; consolidate BSD build.

fstd:
This adds support for NetBSD and OpenBSD by consolidating platform-freebsd.mk to platform-bsd.mk.
I verified that it still builds on FreeBSD.

Cheers,
Timo Buhrmester


fstd:
Hmm, I'm at a loss to explain how these minor changes to the build system could possibly lead to that unit test failing.

mstorsjo:
I think the travis build failure is unrelated - if you e.g. touch either of these commits (amend them and overwrite your branch), it will trigger a rerun of travis and it should succeed.

The root cause of the test failure seems to be this:

    [ RUN      ] EncodeDecodeTestAPI.SimulcastAVC_SPS_PPS_LISTING
    [OpenH264] this = 0x0xb94700, Error:Don't support width(3532) or height(4) which is less than 16 

So if someone wants to fix it, one could try to make sure that such invalid sizes can't be chosen in the SimulcastAVC_SPS_PPS_LISTING test, or make sure that they don't make the test fail.

mstorsjo:
(Also, the changes look good to me, so someone with the right privileges could merge it.)

mstorsjo:
Actually, this error seems to pop up a lot in a lot of different tests, so someone really should look at that.

fstd:
>amend them and overwrite your branch
That seems to have worked, thanks @mstorsjo

HaiboZhu:
Thank you for your contribution.

PR #2191, opened 2015-10-28
Avoid warnings in the cabac code

mstorsjo:
Use int32_t for a parameter that is always 0 or 1, because it is
negated. This fixes "warning C4146: unary minus operator applied
to unsigned type, result still unsigned" in MSVC.

Also add casts to silence MSVC warnings about "conversion from
'WelsEnc::cabac_low_t' to 'uint8_t', possible loss of data".

The generated code still is identical to before, on both gcc
and clang.

@saamas - does it seem ok to you?


saamas:
+1 lgtm.

PR #2204, opened 2015-10-30
[UT] add a .template file for codec UT

sijchen:
https://rbcommons.com/s/OpenH264/r/1323/

### DIFF
--- a/test/api/encode_decode_api_test.h
+++ b/test/api/encode_decode_api_test.h
@@ -1,0 +1,163 @@
+#include <gtest/gtest.h>
+#include "codec_def.h"
+#include "utils/BufferedData.h"
+#include "utils/FileInputStream.h"
+#include "BaseDecoderTest.h"
+#include "BaseEncoderTest.h"
+#include "wels_common_defs.h"
+#include "utils/HashFunctions.h"
+#include <string>
+#include <vector>
+using namespace WelsCommon;
+
+//TODO: some content here in this file is the same with encode_decode_api_test.cpp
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1323/

----
**testing_done**:  
**description**:  <p>add a .template file for codec UT<br />with this people can know the basic way of adding UT and we can generate UTs basing on it</p>

----
 reviewer wayne liu (huili2) 
 time 2015-10-30T02:31:52+00:00 
2015-10-30T02:31:52+00:00 wayne liu (huili2):
><p>OK to extract .h file from current .cpp for easy UT addition.<br />expect following changes to include this file.</p><p>template is not enough for encode-to-decode process, but is OK for some cases.</p>
OK to extract .h file from current .cpp for easy UT addition.<br />expect following changes to include this file.
template is not enough for encode-to-decode process, but is OK for some cases.

2015-11-03T02:09:03+00:00 Sijia Chen (sijchen):
><p>Just keep it.</p>
the codes in template already include the process of encoding and decoding?
yes it contains. But not enough for various requirements of enc-dec. Not sure if it should be a template.
this can be considered the most basic usage of encoder and decoder, as a template.<br />To cover which requirements will be considered 'enough' then?
If being a template to cover decoder side can mean many cases, maybe I can change the name to "encoder_api_test.template"
so from your standpoint, I agree this is OK.
so do you suggest to change the name to "encoder_api_test.template" or keep it?
Just keep it.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1323/

PR #2269, opened 2015-11-24
remove pSliceComplexRatio from SliceThreading

shihuade:
https://rbcommons.com/s/OpenH264/r/1409/

### DIFF
--- a/codec/encoder/core/inc/mt_defs.h
+++ b/codec/encoder/core/inc/mt_defs.h
@@ -80,8 +80,6 @@ WELS_EVENT                      pThreadMasterEvent[MAX_THREADS_NUM];    // event
 
 WELS_MUTEX                      mutexSliceNumUpdate;    // for dynamic slicing mode MT
 
-int32_t*                        pSliceComplexRatio[MAX_DEPENDENCY_LAYER]; // *INT_MULTIPLY
-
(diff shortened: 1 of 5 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1409/

----
**testing_done**:  
**description**:  <ol><li>remove pSliceComplexRatio from pSliceThreading</li><li>the same with https://rbcommons.com/s/OpenH264/r/1404/</li></ol>

----
2015-11-23T08:27:06+00:00 status-change:
fixed travis warnning error


----
 reviewer Sijia Chen (sijchen) 
 time 2015-11-23T19:37:31+00:00 
2015-11-23T19:37:31+00:00 Sijia Chen (sijchen):
><p>I am okay with this change but keep in mind that when no loadbalancing is needed, previously there is no memory needed but now the memory is built within the slice structure. Fortunately that for now this increase is not big. More twisting may be considered in saving unnecessary memory.</p>
I am okay with this change but keep in mind that when no loadbalancing is needed, previously there is no memory needed but now the memory is built within the slice structure. Fortunately that for now this increase is not big. More twisting may be considered in saving unnecessary memory.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1409/

Issue #2252, opened 2015-11-18
No debug symbols for non-Windows release builds


luser:
https://github.com/cisco/openh264/commit/b5792a09f9c80c0f0bf3a6faacd8883c086817e6 fixed the Windows build so we get PDB files when building release, but on other platforms we're not building with debug symbols, so we won't get good symbols for crash reporting etc.

This should be as simple as adding -g to C[XX]FLAGS, but we'll probably also want a make target to strip the binaries for release.

HaiboZhu:
Thanks for your suggestion and we will add it soon. Welcome to contribute on openh264 project.

pengyanhai:
Hi @HaiboZhu, this fix should be included in v1.5.2 which was supposed to be released on Nov 24th according to Maire. I talked to Sijia and others in SJ and we are not able to make it done before the date. So I asked @luser to submit a PR for us. Please contact @luser about how to get the fix done as soon as possible.

ruil2:
added debug symbols

PR #2295, opened 2015-12-07
Add DEBUGSYMBOLS option for makefile under release mode

HaiboZhu:
https://rbcommons.com/s/OpenH264/r/1443/
for git-hub issue https://github.com/cisco/openh264/issues/2252

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,6 +52,10 @@ endif
 ifeq ($(BUILDTYPE), Release)
 CFLAGS += $(CFLAGS_OPT)
 USE_ASM = Yes
+ifeq ($(DEBUGSYMBOLS), True)
+CFLAGS += -g
+CXXFLAGS += -g
+endif
 else
 CFLAGS += $(CFLAGS_DEBUG)
 USE_ASM = No


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1443/

----
**testing_done**:  <p>https://github.com/cisco/openh264/issues/2252</p>
**description**:  Add DEBUGSYMBOLS option for makefile

----
2015-12-04T01:49:03+00:00 status-change:
Remove the tabs in the line


----
 reviewer Forrest Shi (huade) 
 time 2015-12-04T01:49:46+00:00 
2015-12-04T01:49:46+00:00 Forrest Shi (huade):
>Ship It!
Ship It!
Ship it!


----
 reviewer Hank Peng (hankpeng) 
 time 2015-12-07T06:23:14+00:00 
2015-12-07T06:23:14+00:00 Hank Peng (hankpeng):
><p>If I understand in the right way, I need to run the command 'make DEBUGSYMBOLS=True' to generate the binary containing symbols for debugging first and then run 'make' to build the binary for the end user. Does the symbol resolution process check the timestamp of the binary? If so, I am afraid this method doesn't work as the timestamp of the symbol binary and release binary might not match.</p>
If I understand in the right way, I need to run the command 'make DEBUGSYMBOLS=True' to generate the binary containing symbols for debugging first and then run 'make' to build the binary for the end user. Does the symbol resolution process check the timestamp of the binary? If so, I am afraid this method doesn't work as the timestamp of the symbol binary and release binary might not match.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1443/

PR #2328, opened 2016-01-13
fix build issue when some macro turn on

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/1469/


ruil2:
will submit a new request

PR #2336, opened 2016-01-18
Add protection for wrong API call without decoder Initialize

HaiboZhu:
See code review at: https://www.rbcommons.com/s/OpenH264/r/1476/


HaiboZhu:
Add another judgement,

Issue #2348, opened 2016-01-29
openh264 v1.5.1 crashes when a thread can not be created


hardening:
If we're running out of resources and we can't create a thread, then the full context is freed but the threads that have been created aren't killed, leading to segfault.

Patch follows.

sijchen:
thank you for the input. In fact the function CreateSliceThreads is going to retire and its counterpart is WELS_THREAD_ERROR_CODE CWelsThread::Start() under common/src/WelsThread.cpp Ln76. We need to figure out how to perform a same protection as you mentioned in the latter function....

HaiboZhu:
Thanks for hardening's contribution on it.

PR #2349, opened 2016-01-29
Stop created threads if the creation of thread fails

hardening:
If we fail to create a thread, we must stop the one that have been already created or we have these threads that work on freed memory.

This is a fix for #2348.


sijchen:
thank you for the input. In fact the function CreateSliceThreads is going to retire and its counterpart is  WELS_THREAD_ERROR_CODE CWelsThread::Start() under common/src/WelsThread.cpp Ln76. We need to figure out how to perform a same protection as you mentioned in the latter function....

PR #2351, opened 2016-02-02
fix frame size constraints for width and height

huili2:
make constraint as total MB per frame instead of single width or height.
see https://rbcommons.com/s/OpenH264/r/1503/

### DIFF
--- a/codec/common/inc/utils.h
+++ b/codec/common/inc/utils.h
@@ -43,8 +43,7 @@
 #include "typedefs.h"
 
 #define MAX_LOG_SIZE    1024
-#define MAX_WIDTH      (4096)
-#define MAX_HEIGHT     (2304)//MAX_FS_LEVEL51 (36864); MAX_FS_LEVEL51*256/4096 = 2304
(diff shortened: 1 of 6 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1503/

----
**testing_done**:  
**description**:  <p>fix frame size constraints for width and height, with only Width*Height limit rather than width or height</p>

----
 reviewer Sijia Chen (sijchen) 
 time 2016-02-02T00:47:56+00:00 
2016-02-02T00:47:56+00:00 Sijia Chen (sijchen):
><p>I am okay with this one but can we bring in some views from assembly side to make sure that expanding this has no impact on assembly codes?</p>
I am okay with this one but can we bring in some views from assembly side to make sure that expanding this has no impact on assembly codes?
Ship it!


----
 time 2016-02-02T01:59:40+00:00 
2016-02-02T01:59:40+00:00 zhiliang wang (zhilwang):
><p>OK with assembly code</p>
OK with assembly code
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1503/

PR #2384, opened 2016-02-26
modify return value check inside decoder

huili2:
make internal return value as ERROR type instead of direct instant value, for more convenient error tracking. And add missing protection for malloc failure case for some variables.
see:
https://rbcommons.com/s/OpenH264/r/1526/

PR #2385, opened 2016-02-26
modify return value check inside decoder

huili2:
same as pull request #2384 but fixing build warnings.


huili2:
will open a new pull request for this build warning issue.

PR #2411, opened 2016-03-17
fix memory leak when alloc failed in decoder

huili2:
to keep consistence with that in UninitialDqLayersContext(), so that the failure in between malloc can lead to correct free operation.
see:
https://rbcommons.com/s/OpenH264/r/1568/

### DIFF
--- a/codec/decoder/core/src/decoder_core.cpp
+++ b/codec/decoder/core/src/decoder_core.cpp
@@ -1247,6 +1247,7 @@ int32_t InitialDqLayersContext (PWelsDecoderContext pCtx, const int32_t kiMaxWid
     if (pDq == NULL)
       return ERR_INFO_OUT_OF_MEMORY;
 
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1568/

----
**testing_done**:  
**description**:  <p>fix memory leak when alloc failed in decoder<br />the variable sequence should be consistent as in UninitialDqLayersContext(), so as to prevent possible memory leak issue once malloc failed in between.</p>

----
 reviewer Haibo Zhu (haibozhu) 
 time 2016-03-17T04:28:58+00:00 
2016-03-17T04:28:58+00:00 Haibo Zhu (haibozhu):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1568/

PR #2413, opened 2016-03-17
Sync commits to other branches

HaiboZhu:
As same as #2411 #2410


HaiboZhu:
Remove assert makes some variable unused

PR #2418, opened 2016-03-18
fix preprocessing initialization logic

ruil2:
reviewed at:https://rbcommons.com/s/OpenH264/r/1572/

### DIFF
--- a/codec/encoder/core/inc/wels_preprocess.h
+++ b/codec/encoder/core/inc/wels_preprocess.h
@@ -121,7 +121,7 @@ class CWelsPreProcess {
   virtual  ~CWelsPreProcess();
 
  public:
-  int32_t WelsPreprocessReset (sWelsEncCtx* pEncCtx);
+  int32_t WelsPreprocessReset (sWelsEncCtx* pEncCtx,int32_t iWidth,int32_t iHeight);
   int32_t AllocSpatialPictures (sWelsEncCtx* pCtx, SWelsSvcCodingParam* pParam);
   void    FreeSpatialPictures (sWelsEncCtx* pCtx);
   int32_t BuildSpatialPicList (sWelsEncCtx* pEncCtx, const SSourcePicture* kpSrcPic);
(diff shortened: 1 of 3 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1572/

----
**testing_done**:  
**description**:  <ol><li>preprocessing initilization should not be done when encoder is initialized, sometimes the input resolution isn't the actual input resolution here.</li><li>when input resolution changes, preprocessing needs to be reset.</li></ol>

----
 reviewer Sijia Chen (sijchen) 
 time 2016-03-18T21:19:30+00:00 
2016-03-18T21:19:30+00:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/encoder_ext.cpp line4631**: 
 id="comment369360"><p>I am not quite sure on description1 since it should be common that the initialization para will be used in following encoding, although there is chance that it may be updated before actual encoding.</p>
I am not quite sure on description1 since it should be common that the initialization para will be used in following encoding, although there is chance that it may be updated before actual encoding.

2016-03-21T01:15:53+00:00 Karina Li (ruil2):
><p>this initialization will use input width and height. so we should not put this here. it is a potential bug.<br />if the input resolution isn't same with the highest layer resolution.</p>
this initialization will use input width and height. so we should not put this here. it is a potential bug.<br />if the input resolution isn't same with the highest layer resolution.

2016-03-21T01:15:53+00:00 Karina Li (ruil2):
 on **codec/encoder/core/src/wels_preprocess.cpp line212**: 
 id="comment369364"><p>doing such Reset in BuildSpatialPicList is somehow weird, (but good as a protection maybe). can we do it in parameter updating function? (such as WelsEncoderParamAdjust)</p>
doing such Reset in BuildSpatialPicList is somehow weird, (but good as a protection maybe). can we do it in parameter updating function? (such as WelsEncoderParamAdjust)

2016-03-21T01:15:53+00:00 Karina Li (ruil2):
><p>only after the encoder frame is input, the spatial picture will be allocated.  and when the input frame changes, the spatial pictures should be reallocated. it doesn't mean input encoder parameters.</p>
only after the encoder frame is input, the spatial picture will be allocated.  and when the input frame changes, the spatial pictures should be reallocated. it doesn't mean input encoder parameters.


----
2016-03-21T02:51:04+00:00 status-change:
update reset function


----
 reviewer Sijia Chen (sijchen) 
 time 2016-03-21T18:08:17+00:00 
2016-03-21T18:08:17+00:00 Sijia Chen (sijchen):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1572/

PR #2431, opened 2016-03-29
fix frame rate issue 

ruil2:
reviewed at:https://rbcommons.com/s/OpenH264/r/1583/

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -1177,7 +1177,7 @@ static inline int32_t InitSliceThreadInfo (sWelsEncCtx** ppCtx,
     pSliceThreadInfo->iEncodedSliceNumInThread[iIdx] = 0;
     pSliceThreadInfo->pSliceInThread[iIdx]           = (SSlice*)pMa->WelsMallocz (sizeof (SSlice) *
         iMaxSliceNumInThread, "pSliceInThread");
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1583/

----
**testing_done**:  
**description**:  <p>fix temporal layer issue when output frame rate is different from input frame rate<br />when the current frame isn't encoded becasue of frame rate exactor, codingIdx should be updated. otherwise the frames will be skipped</p>

----
 reviewer Sijia Chen (sijchen) 
 time 2016-03-29T18:28:40+00:00 
2016-03-29T18:28:40+00:00 Sijia Chen (sijchen):
><p>don't quite understand but the change is harmless.</p><p>"when the current frame isn't encoded becasue of frame rate exactor" then it is actually skipped? so it is the same as "the frames will be skipped"</p>
don't quite understand but the change is harmless.
"when the current frame isn't encoded becasue of frame rate exactor" then it is actually skipped? so it is the same as "the frames will be skipped"
Ship it!

2016-03-30T16:55:02+00:00 Karina Li (ruil2):
><p>I see, thx</p>
for an example.<br />the input frame rate is 30fps. output frame rate is 15fps.<br />codingIndex is used to decide which frame should be skipped. if the frame is skipped,but the codingIdx keeps no change, the next input frame will also be skipped. So the codingIdx should be updated in this case.
I see, thx


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1583/

Issue #2434, opened 2016-04-06
Got the encDemo crashed on iOS


fastxyf:
Test environment:
openh264 version : 1.4,1.5, 1.51
iOS version: 9.3
iOS Device: iPhone 5s, iPhone 6S

Test step:
1. Open codec/build/iOS/enc/encDemo/encDemo.xcodeproj
2. Disable AutoTestEnc(), 
  /*
  if (AutoTestEnc() == 0)
    NSLog (@"Auto testing running sucessfully");
  else
    NSLog (@"Auto testing running failed");
  abort();*/
3.Run encDemo , click Simple Encode Test,

APP crash, EXC_BAD_ACCESS, block in this line:pfDeblocking->pfLumaDeblockingLT4Hor (pPix, iStride, iAlpha, iBeta, iTc);

#0	0x000000010013d93c in DeblockLumaLt4H_AArch64_neon ()
#1	0x00000001001067ac in WelsEnc::FilteringEdgeLumaV(WelsEnc::tagDeblockingFunc*, WelsEnc::TagDeblockingFilter*, unsigned char*, int, unsigned char*) at /Users/Joe/temp/openh264-1.4.0/codec/encoder/core/src/deblocking.cpp:260
#2	0x0000000100106cf8 in WelsEnc::DeblockingInterMb(WelsEnc::tagDeblockingFunc*, WelsEnc::TagMB*, WelsEnc::TagDeblockingFilter*, unsigned char (*) [4][4]) at /Users/Joe/temp/openh264-1.4.0/codec/encoder/core/src/deblocking.cpp:397
#3	0x0000000100107cf4 in WelsEnc::DeblockingMbAvcbase(WelsEnc::TagWelsFuncPointerList*, WelsEnc::TagMB*, WelsEnc::TagDeblockingFilter*) at /Users/Joe/temp/openh264-1.4.0/codec/encoder/core/src/deblocking.cpp:651
#4	0x0000000100107e30 in WelsEnc::DeblockingFilterFrameAvcbase(WelsEnc::TagDqLayer*, WelsEnc::TagWelsFuncPointerList*) at /Users/Joe/temp/openh264-1.4.0/codec/encoder/core/src/deblocking.cpp:684
[...]

PR #2439, opened 2016-04-12
add missing sign extension for arm64 on deblocking_aarch64_neon.S

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/1589/

### DIFF
--- a/codec/common/arm64/arm_arch64_common_macro.S
+++ b/codec/common/arm64/arm_arch64_common_macro.S
@@ -62,3 +62,7 @@ ret
 .endm
 
 #endif
+
+.macro SIGN_EXTENSION arg0, arg1
+  sxtw \arg0, \arg1
+.endm
--- a/codec/common/arm64/deblocking_aarch64_neon.S
+++ b/codec/common/arm64/deblocking_aarch64_neon.S
@@ -305,6 +305,7 @@ WELS_ASM_AARCH64_FUNC_END
 WELS_ASM_AARCH64_FUNC_BEGIN DeblockLumaLt4V_AArch64_neon //uint8_t* pPix, int32_t iStride, int32_t iAlpha, int32_t iBeta, int8_t* tc
     dup v16.16b, w2 //alpha
     dup v17.16b, w3 //beta
+    SIGN_EXTENSION x1,w1
     add x2, x1, x1, lsl #1
     sub x2, x0, x2
     movi v23.16b, #128
@@ -363,8 +364,8 @@ WELS_ASM_AARCH64_FUNC_END
 WELS_ASM_AARCH64_FUNC_BEGIN DeblockLumaEq4V_AArch64_neon
     dup     v16.16b, w2 //alpha
     dup     v17.16b, w3 //beta
+    SIGN_EXTENSION x1,w1
     sub     x3, x0, x1, lsl #2
-
     ld1     {v0.16b}, [x3], x1
     ld1     {v4.16b}, [x0], x1
     ld1     {v1.16b}, [x3], x1
@@ -431,7 +432,7 @@ WELS_ASM_AARCH64_FUNC_BEGIN DeblockLumaLt4H_AArch64_neon //uint8_t* pPix, int32_
     dup v17.16b, w3 //beta
     sub x2, x0, #3
     movi v23.16b, #128
-
+    SIGN_EXTENSION x1,w1
     LOAD_LUMA_DATA_3      v0, v1, v2, v3, v4, v5, 0
     LOAD_LUMA_DATA_3      v0, v1, v2, v3, v4, v5, 1
     LOAD_LUMA_DATA_3      v0, v1, v2, v3, v4, v5, 2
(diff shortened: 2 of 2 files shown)


ruil2:
fix #2434

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1589/

----
**testing_done**:  
**description**:  <p>fix arm64 crash issue on iOS9.3 and Xcode7.3</p><p>git issue: https://github.com/cisco/openh264/issues/2434</p><p>I am not sure whehter it is the lastest apple compiler issue.</p><ol><li>this crash only happens on the lastest iOS version.</li><li>no crash when use -O0 optimization level on the lastest iOS version</li><li>no crash when using -O1,-O2,-O3,-Os.</li><li>X1 register isn't set zero before using it. when 32-bit variable is put into X1 register, the higher 32-bit isn't zero. which will get the very large stride and cause memory overflow. In order to fix this issue, set 0 to higher 32-bit for X1 register.</li></ol><p>if you have any good idea or suggestion, please add your comments. thanks a lot!</p>

----
 time 2016-04-12T06:26:42+00:00 
2016-04-12T06:26:42+00:00 Martin Storsj (mstorsjo):
><p>This isn't the cleanest possible fix.</p><p>The issue is that the parameter is an <code>int32_t</code>, which means that there is no guarantee that the upper half of a 64 bit register is zero - you are only supposed to look at the lower 32 bit of it.</p><p>To fix this, you can do either of these:<br />- Use <code>uxtw x1, w1</code> or <code>sxtw x1, w1</code> to do unsigned or signed extension of a 32 bit register into a 64 bit register. (In x86_64, we have the macro <code>SIGN_EXTENSION</code>, which is equivalent to using sxtw. Recently a similar fix was needed there due to newer clang versions as well, see a4e71d6662bf806e6fc55ff5d4b2ec6743dae1e3.)<br />- Make sure to use the register via the name <code>w1</code> only, not <code>x1</code>. Then you only use the lower 32 bit half.<br />- Change the parameter data type from <code>int32_t</code> to <code>ptrdiff_t</code> or maybe <code>intptr_t</code>; then you can be sure that it is a proper 64 bit type here, and the compiler will fill the whole register properly. (We don't use <code>ptrdiff_t</code> anywhere at the moment, so I don't know if that works well on all older MSVC versions or not, but we at least use <code>uintptr_t</code> already, so I think <code>intptr_t</code> should be fine as well.) If we do this, we can remove the use of the <code>SIGN_EXTENSION</code> macro in x86_64 as well, for the cases where it is changed.</p>
This isn't the cleanest possible fix.
The issue is that the parameter is an <code>int32_t</code>, which means that there is no guarantee that the upper half of a 64 bit register is zero - you are only supposed to look at the lower 32 bit of it.
To fix this, you can do either of these:<br />- Use <code>uxtw x1, w1</code> or <code>sxtw x1, w1</code> to do unsigned or signed extension of a 32 bit register into a 64 bit register. (In x86_64, we have the macro <code>SIGN_EXTENSION</code>, which is equivalent to using sxtw. Recently a similar fix was needed there due to newer clang versions as well, see a4e71d6662bf806e6fc55ff5d4b2ec6743dae1e3.)<br />- Make sure to use the register via the name <code>w1</code> only, not <code>x1</code>. Then you only use the lower 32 bit half.<br />- Change the parameter data type from <code>int32_t</code> to <code>ptrdiff_t</code> or maybe <code>intptr_t</code>; then you can be sure that it is a proper 64 bit type here, and the compiler will fill the whole register properly. (We don't use <code>ptrdiff_t</code> anywhere at the moment, so I don't know if that works well on all older MSVC versions or not, but we at least use <code>uintptr_t</code> already, so I think <code>intptr_t</code> should be fine as well.) If we do this, we can remove the use of the <code>SIGN_EXTENSION</code> macro in x86_64 as well, for the cases where it is changed.

2016-04-12T07:35:48+00:00 Karina Li (ruil2):
><p>very appreciate your comments, I will use #1 to fix this issue.</p>
very appreciate your comments, I will use #1 to fix this issue.


----
2016-04-12T08:30:55+00:00 status-change:
update by using SIGN_EXTENSION macro


----
 time 2016-04-12T08:35:08+00:00 
2016-04-12T08:35:08+00:00 Martin Storsj (mstorsjo):
><p>This looks ok to me.</p><p>We should probably also check all the places where SIGN_EXTENSION is used for x86, to see if we need to add it in more places. Right now, the code just works by coincidence, but any compiler update can break it, just as this bug report showed. We should also check any other assembly function that uses 32 bit parameters, if they also need a similar sign extension, to avoid future bugs.</p>
This looks ok to me.
We should probably also check all the places where SIGN_EXTENSION is used for x86, to see if we need to add it in more places. Right now, the code just works by coincidence, but any compiler update can break it, just as this bug report showed. We should also check any other assembly function that uses 32 bit parameters, if they also need a similar sign extension, to avoid future bugs.
Ship it!

2016-04-12T08:43:16+00:00 Karina Li (ruil2):
><p>good suggestion, we will checking the places to avoid the same issue.</p>
good suggestion, we will checking the places to avoid the same issue.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1589/

PR #2464, opened 2016-05-17
MVC aka Stereoscopic 3D support

bumblebritches57:
it's very basic support just to get started.


bumblebritches57:
Awesome! I was starting to wonder if you guys would ever accept it!

PR #2465, opened 2016-05-18
fix temporal layer skip issue

ruil2:
reviewed at:https://rbcommons.com/s/OpenH264/r/1602/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1602/

----
**testing_done**:  
**description**:  fix temporal layer skip issue

----
 reviewer Sijia Chen (sijchen) 
 time 2016-05-18T16:33:53+00:00 
2016-05-18T16:33:53+00:00 Sijia Chen (sijchen):
><p>how can we design test case to avoid regression on such hidden problem in the future?</p>
how can we design test case to avoid regression on such hidden problem in the future?
Ship it!

2016-05-19T01:17:15+00:00 Karina Li (ruil2):
><ol><li>we need to check the output as video engine calling.</li><li>use application parameters to design our TA case.</li></ol>


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1602/

PR #2473, opened 2016-05-24
modify the interface that use a independent subseqID for each layer

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/1606/

### DIFF
--- a/codec/api/svc/codec_app_def.h
+++ b/codec/api/svc/codec_app_def.h
@@ -578,6 +578,11 @@ typedef struct {
   EVideoFrameType eFrameType;
   unsigned char uiLayerType;
 
+  /**
+   * The sub sequence layers are ordered hierarchically based on their dependency on each other so that any picture in a layer shall not be
+   * predicted from any picture on any higher layer.
+  */
+  int   iSubSeqId;                ///< refer to D.2.11 Sub-sequence information SEI message semantics
   int   iNalCount;              ///< count number of NAL coded already
   int*  pNalLengthInByte;       ///< length of NAL size in byte from 0 to iNalCount-1
   unsigned char*  pBsBuf;       ///< buffer of bitstream contained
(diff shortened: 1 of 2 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1606/

----
**testing_done**:  
**description**:  <p>modify the interface that use a independent subseqID for each layer<br />1. remove uiTemporalId in SFrameBSInfo. this isn't be used.<br />2. each spatial layer can be controlled seperately.so we need to use a independent subseqID for each layer</p>

----
 reviewer Sijia Chen (sijchen) 
 time 2016-05-19T16:54:03+00:00 
2016-05-19T16:54:03+00:00 Sijia Chen (sijchen):
><p>if we are changing the interface this way, it will not be backward compatiable, then we need a major version upgrading along with this change.</p>
if we are changing the interface this way, it will not be backward compatiable, then we need a major version upgrading along with this change.

2016-05-20T16:02:46+00:00 Karina Li (ruil2):
><p>the existing one is 1.6. although we did not push a release for that. I am not quite sure whether we need a major version upgrade or minor version upgrade for this change. maybe you guyes can discuss in daily. but having a version upgrade will definitely align with the interface change so I will prefer to have some rather than no.</p>
1.6 version is ok?
the existing one is 1.6. although we did not push a release for that. I am not quite sure whether we need a major version upgrade or minor version upgrade for this change. maybe you guyes can discuss in daily. but having a version upgrade will definitely align with the interface change so I will prefer to have some rather than no.


----
 reviewer Haibo Zhu (haibozhu) 
 time 2016-05-25T01:59:27+00:00 
2016-05-25T01:59:27+00:00 Haibo Zhu (haibozhu):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1606/

PR #2476, opened 2016-05-30
    fix dependency ID mapping issue

ruil2:
reviewed at: https://rbcommons.com/s/OpenH264/r/1620/


mstorsjo:
After merging this, I see use of uninitialized data in `codec_unittest` in `EncoderInterfaceTest.MemoryCheckTest`.

ruil2:
@mstorsjo how to see " uninitialized data in codec_unittest in EncoderInterfaceTest.MemoryCheckTest"?

mstorsjo:
Try this, on linux:

    make BUILDTYPE=Debug
    valgrind --track-origins=yes ./codec_unittest --gtest_filter=EncoderInterfaceTest.MemoryCheckTest

ruil2:
@mstorsjo #2488 fixed this issue. 
very appreciate your checking carefully. thanks!

mstorsjo:
Thanks, it seems to be fixed in my testing now as well!

mstorsjo:
Even after the latest fixes, I see issues in the same test, if I run it with a specific seed:
`valgrind --track-origins=yes ./codec_unittest --seed=1465181492`

(Note that this seed only trigger the issue if you don't add `--gtest_filter`, so you need to wait for the full run.)

ruil2:
I run UT as you specified. 
valgrind --track-origins=yes ./codec_unittest --seed=1465181492

there is no any issues report when finished the full run by using the latest code on master.

could you give more info in your testing?

mstorsjo:
The problem with issues that require a specific seed is that it depends on you having the exact same random number generator (which probably is the same on most common linux versions), and run the exact same tests before hitting the one with the error. If you run more or less tests before (e.g. you don't have AVX but I have, or vice versa) it won't hit it.

I tried reproducing it by only running the right test (to hopefully avoid such issues), and tested a number of seeds until I found one that failed for me. Try this:

    valgrind ./codec_unittest --seed=3 --gtest_filter=EncoderInterfaceTest.MemoryCheckTest

If you don't get errors with this either, try adjusting seed up/down until you hit it - I tried with 0, 1, and 2 before, until I hit the errors with seed=3. So you should probably not need to test that many until you hit it.

mstorsjo:
Were you able to reproduce the issue with this command?

    valgrind ./codec_unittest --seed=3 --gtest_filter=EncoderInterfaceTest.MemoryCheckTest

ruil2:
I can't not reproduce this issue from seed 0 to 200 in debug mode.
but I found there is some "uninitialized value" issue in release mode, and these issues also existed in older version before this commit. I am checking these issues.

Could you provide the parameters on your side,thanks!

mstorsjo:
Sorry, you're right, I can't reproduce the issue in debug mode any longer, only in release mode - thanks for taking the time to test with lots of seeds.

So that's a separate issue then; it was introduced around the same time though (plus/minus a few days), since valgrind with release builds used to run without errors up until around when this (and some other things) were merged - I appreciate if you can take time to try to pinpoint it.

I can reproduce it with the following kind of build, though:

    make BUILDTYPE=Debug USE_ASM=Yes -j4
    valgrind ./codec_unittest --seed=3 --gtest_filter=EncoderInterfaceTest.MemoryCheckTest

So it's clearly affected by some of the newly added/changed x86 asm routines. Maybe @saamas can help check? I think it might be related to one of his changes. It should be pretty easy to pin down by disabling the newly added/changed asm routines one by own.

saamas:
This appears to be the result of a valgrind shortcoming: It appears that entire vectors are considered to be undefined if a single lane contains uninitialized data; so even if the uninitialized part is later overwritten, valgrind will still complain about uninitialized values. A workaround is to zero-initialize input buffers (or at least any padding used).

mstorsjo:
Thanks for checking!

Even if it's only a valgrind false positive, would you consider some hack to silence it, to ease valgrind testing? (I've got a regression test setup to catch issues like this, alerting me of all valgrind errors in the testsuite.) We did this before in 44bce08d6076 for a similar issue/valgrind shortcoming, although it's arguably quite ugly.

In case you aren't interested in such a hack, can you point me to the place where it arises (in case you debugged it that far already), so I can patch it locally before running valgrind?

ruil2:
@mstorsjo @saamas  thanks you for debugging and fixing this issue.

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1620/

----
**testing_done**:  
**description**:  <p>fix did mapping issue</p>

----
 reviewer Haibo Zhu (haibozhu) 
 time 2016-05-30T02:42:33+00:00 
2016-05-30T02:42:33+00:00 Haibo Zhu (haibozhu):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1620/

PR #2488, opened 2016-06-02
    fix codingIdx update issue

ruil2:
reviewed at:https://rbcommons.com/s/OpenH264/r/1631/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1631/

----
**testing_done**:  
**description**:  <ol><li>codingIdx should be updated after encoding</li><li>put iContinualSkipFrames calculation in rc module to make the function clear</li><li>becasue codingIdx isn't updated correclty in the last version, so the SHA table needs to be updated.</li></ol>

----

----
 reviewer Sijia Chen (sijchen) 
 time 2016-06-02T16:38:12+00:00 
2016-06-02T16:38:12+00:00 Sijia Chen (sijchen):
Ship it!

2016-06-02T16:38:12+00:00 Sijia Chen (sijchen):
 on **codec/encoder/core/src/encoder_ext.cpp line4141**: 
 id="comment402594"><p>It looks like this is such a ease-to-have-bug operation. Is it possible to hide these details, such as "UpdateCodingIndex" function, in which we do the SAVC/SSVC difference?</p>
It looks like this is such a ease-to-have-bug operation. Is it possible to hide these details, such as "UpdateCodingIndex" function, in which we do the SAVC/SSVC difference?

2016-06-03T01:03:41+00:00 Karina Li (ruil2):
><p>yes, it is very easy to make issues. 1. will think about an other solution 2.needs to add more UTs</p>
yes, it is very easy to make issues. 1. will think about an other solution 2.needs to add more UTs


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1631/

PR #2489, opened 2016-06-02
[Processing] DyadicBilinearDownsample optimizations

saamas:
https://rbcommons.com/s/OpenH264/r/1632/
https://rbcommons.com/s/OpenH264/r/1633/
https://rbcommons.com/s/OpenH264/r/1634/

### DIFF
--- a/codec/common/x86/asm_inc.asm
+++ b/codec/common/x86/asm_inc.asm
@@ -485,7 +485,7 @@ SECTION .note.GNU-stack noalloc noexec nowrite progbits ; Mark the stack as non-
 %endmacro
 
 %macro WELS_EXTERN 1
-    ALIGN 16
+    ALIGN 16, nop
     %ifdef PREFIX
         global _%1
         %define %1 _%1
--- a/codec/processing/src/common/util.h
+++ b/codec/processing/src/common/util.h
@@ -83,10 +83,6 @@ WELSVP_NAMESPACE_BEGIN
 #define WELS_CLAMP(x, minv, maxv)  WELS_MIN(WELS_MAX(x, minv), maxv)
 
 #define ALIGNBYTES         (16)       /* Worst case is requiring alignment to an 16 byte boundary */
-#define WELS_ALIGN(iInput)   ((iInput+(ALIGNMENT-1)) & ~(ALIGNMENT-1))
-#define WELS_ALIGN2(iInput)  ((iInput+1) & ~1)
-#define WELS_ALIGN4(iInput)  ((iInput+3) & ~3)
-#define WELS_ALIGN8(iInput)  ((iInput+7) & ~7)
 
 #define WelsCastFromPointer(p)      (reinterpret_cast<intptr_t>(p))
 #define WelsStaticCast(type, p)  (static_cast<type>(p))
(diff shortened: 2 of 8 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1632/

----
**testing_done**:  
**description**:  The WELS_ALIGN macro here aliases the WELS_ALIGN macro in macros.hwhich is inconvenient. Just remove these unused macros.

----
 reviewer Karina Li (ruil2) 
 time 2016-06-03T05:37:44+00:00 
2016-06-03T05:37:44+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1632/

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1633/

----
**testing_done**:  
**description**:  Assume that data can be written into the padding area following eachline. This enables the use of faster routines for more cases.Align downsample buffer stride to a multiple of 32.With this all strides used should be a multiple of 16, which meansthat use of narrower downsample routines can be dropped altogether.

----
2016-06-03T10:08:08+00:00 status-change:
Clear LSB before aligning source width.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1633/

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1634/

----
**testing_done**:  
**description**:  Average vertically before horizontally; horizontal averaging is moreworksome. Doing the vertical averaging first reduces the number ofhorizontal averages by half.Use pmaddubsw and pavgw to do the horizontal averaging for a slightperformance improvement.Minor tweaks.Improve the SSSE3 dyadic downsample routines and drop the SSE4 routines.The non-temporal loads used in the SSE4 routines do nothing for cache-backed memory AFAIK.Adjust tests because averaging vertically first gives slightly differentoutput.~2.39x speedup for the widthx32 routine on Haswell when not memory-bound.~2.20x speedup for the widthx16 routine on Haswell when not memory-bound.Note that the widthx16 routine can be unrolled for further speedup.

----
 reviewer Karina Li (ruil2) 
 time 2016-06-06T04:50:26+00:00 
2016-06-06T04:50:26+00:00 Karina Li (ruil2):
 on **test/api/decode_api_test.cpp line763**: 
 id="comment403515"><p>why need to change this?</p>
why need to change this?

2016-06-08T01:27:01+00:00 Sindre Aam (saamas):
><p>got it. If the checksum change is only caused by the order of horizontally and  vertically. could you add some comments on Ifdef? thanks!<br />Because there is some benefit on the change, we also can use the same implementation on Nero assembly.</p>
From what I can see, the test encodes multiple layers and then decodes and verifies that the output matches the expected checksum. The output is slightly different with this change as a result of averaging vertically first when downsampling the input.
Average horizontally firstly, then average vertically in C funciton. it doesn't trigger UT failed. why is it?
The checksum change is ifdef’ed on X86_ASM. It will fail if we run the UT on X86 w/o SSSE3 support and ASM enabled though. An option is to modify the test to accept both checksums so that this cannot happen. I think I prefer that in fact. I’ll add that change if you otherwise find this acceptable unless you disagree.
got it. If the checksum change is only caused by the order of horizontally and  vertically. could you add some comments on Ifdef? thanks!<br />Because there is some benefit on the change, we also can use the same implementation on Nero assembly.


----
2016-06-08T19:57:17+00:00 status-change:
Add comment on X86_ASM checksum ifdef.


----
 reviewer Karina Li (ruil2) 
 time 2016-06-12T02:02:17+00:00 
2016-06-12T02:02:17+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1634/

PR #2498, opened 2016-06-15
Use assert.h instead of cassert

mstorsjo:
This fixes building for android differently than in f5e483ce.

On android, <cassert> isn't available in the normal include path,
only when the STL headers are available.

We intentionally avoid using STL within the main libopenh264.so, to
simplify dependency chains for users of the library (which otherwise
could run into conflicts if the surrounding app would want to use
a different STL implementation).

The previous fix only provided headers, not actually linking
against STL, so at this point it's not a real issue yet, but it's
still a very slippery slope towards accidentally starting relying on
STL within the core library.

Instead explicitly avoid using STL within the core library, by not
even providing the include path.

Review at https://rbcommons.com/s/OpenH264/r/1642/.

FYI @saamas @GuangweiWang @HaiboZhu


HaiboZhu:
Thank you very much. Your comments are very professional.

sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1642/

----
**testing_done**:  
**description**:  Use assert.h instead of cassertThis fixes building for android differently than in f5e483ce.On android, &lt;cassert&gt; isn&#39;t available in the normal include path,only when the STL headers are available.We intentionally avoid using STL within the main libopenh264.so, tosimplify dependency chains for users of the library (which otherwisecould run into conflicts if the surrounding app would want to usea different STL implementation).The previous fix only provided headers, not actually linkingagainst STL, so at this point it&#39;s not a real issue yet, but it&#39;sstill a very slippery slope towards accidentally starting relying onSTL within the core library.Instead explicitly avoid using STL within the core library, by noteven providing the include path.

----
 reviewer Haibo Zhu (haibozhu) 
 time 2016-06-16T01:20:02+00:00 
2016-06-16T01:20:02+00:00 Haibo Zhu (haibozhu):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1642/

PR #2499, opened 2016-06-16
[Encoder] Avoid valgrind downsampling false positives

saamas:
X86 SIMD downsampling routines may, for convenience, read slightly beyond
the input data and into the alignment padding area beyond each line. This
causes valgrind to warn about uninitialized values even if these values
only affect lanes of a SIMD vector that are effectively never used.

Avoid these false positives by zero-initializing the padding area beyond
each line of the source buffer used for downsampling.

Reviewed at: https://rbcommons.com/s/OpenH264/r/1643/

### DIFF
--- a/codec/encoder/core/src/wels_preprocess.cpp
+++ b/codec/encoder/core/src/wels_preprocess.cpp
@@ -36,6 +36,17 @@
 #include "utils.h"
 #include "encoder.h"
 
+namespace {
+
+void ClearEndOfLinePadding (uint8_t* pData, int32_t iStride, int32_t iWidth, int32_t iHeight) {
+  if (iWidth < iStride) {
+    for (int32_t i = 0; i < iHeight; ++i)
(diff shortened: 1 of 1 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1643/

----
**testing_done**:  
**description**:  <p>X86 SIMD downsampling routines may, for convenience, read slightly beyond<br />the input data and into the alignment padding area beyond each line. This<br />causes valgrind to warn about uninitialized values even if these values<br />only affect lanes of a SIMD vector that are effectively never used.</p><p>Avoid these false positives by zero-initializing the padding area beyond<br />each line of the source buffer used for downsampling.</p>

----
 time 2016-06-20T12:26:34-07:00 
2016-06-20T12:26:34-07:00 Martin Storsj (mstorsjo):
><p>Ship It!</p>
Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1643/

PR #2511, opened 2016-07-04
Update v1.6 information

HaiboZhu:


Makefile
```diff
@@ -34,7 +34,7 @@ CCASFLAGS=$(CFLAGS)
 STATIC_LDFLAGS=-lstdc++
 
 VERSION=1.6
-SHAREDLIBVERSION=2
+SHAREDLIBVERSION=0
```
mstorsjo: Why did you change SHAREDLIBVERSION here? This new version is not compatible with the old libraries that had SHAREDLIBVERSION=0. Please fix this (both for the openh264v1.6 branch and for master).

Also, please make a tag (v1.6.0) for the version that was used for building the binaries that were released - having a branch (where new commits can be added) isn't quite enough - there have been tags for all previous releases.

HaiboZhu:
Thanks for your support on openh264 branch. (1) For SHAREDLIBVERSION problem, there was some confusion in openh264 version management and now we want to sync it with the official release version. (2) now there is a tag (v1.6.0) for release v1.6.0

mstorsjo:
For the library version, I don't understand what you are saying. Is it a mistake that you will fix? If not, why do you think it's ok to reuse shared library version 0 that was already used for version 1.4?

mstorsjo:
If you mean that you want to sync if with the versions in the official binary builds, it doesn't make sense. Those binaries don't include any shared library version at all, it's only used when you do "make install", to create libopenh264.so.1, to make sure it doesn't overwrite the older version lib (when the libraries aren't binary compatible). Version 1.4 (the first version to use a shared library version suffix) had shared library version 0, and version 1.5 was version 1. You can't reset this to 0 for 1.6, it needs to be a new, unique number not used before. The only case when you can reuse an old number is if the binary interface hasn't changed since the last version.

HaiboZhu:
I updated the SHAREDLIBVERSION to 3 in openh264v1.6 branch (https://github.com/cisco/openh264/pull/2522). But I don't know the impact of the this mistake. Should I update the binary files which we generated from the previous code? If yes, which platforms should I update?

mstorsjo:
Thanks! Technically, you didn't need to set it to 3, 2 would have been ok as well. (The last release, 1.5, used version 1, and sometimes after that, when binary compatibility was broken with the last release, it was changed to 2, and there hasn't been any release using version 2 yet. But 3 is also fine.)

Please also update master similarly.

Yes, you need to update the files, the linux32/64 and osx32/64 ones - the rest don't need to be rebuilt I think.

huili2:
It is NOT compatible, so we will fix the issue soon.

PR #2536, opened 2016-07-20
Consistently use Yes instead of true for makefile variables

mstorsjo:
All similar variables use Yes so far; if we want to change what
strings we use for booleans in makefiles, the change should be
consistent.

Review at https://rbcommons.com/s/OpenH264/r/1663/.

### DIFF
--- a/build/arch.mk
+++ b/build/arch.mk
@@ -1,10 +1,10 @@
 #for x86
-HAVE_AVX2 := true
+HAVE_AVX2 := Yes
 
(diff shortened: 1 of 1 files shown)


mstorsjo:
Would anybody care to merge this? It was approved on reviewboard already.

HaiboZhu:
About this change, it is OK for code review but it will affect the build pipeline of some products so we want to merge it later. I am sorry about it.

PR #2558, opened 2016-09-13
remove too many log for IDR loss under EC is off

huili2:
when EC is off, the log "CAN NOT find..." exists for succeeding slices. Modify so that only the first input slice would print this log.

code review at:
https://rbcommons.com/s/OpenH264/r/1699/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1699/

----
**testing_done**:  
**description**:  <p>remove too many log for IDR loss under EC is off<br />when EC is off, subsequent packet decoding will always print "CAN NOT find IDR...". Only enable printing logs by detecting the first data as new sequence.</p>

----
 reviewer Karina Li (ruil2) 
 time 2016-09-13T04:01:57+00:00 
2016-09-13T04:01:57+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1699/

PR #2560, opened 2016-09-19
fix list queue for MT & add return check

ycqian:
1、remove malloc in class constructor.
2、fix bug in WelsList ExpandList function.
3、add some return check.


ruil2:
please merge these commits to one to remove the failed commit.

PR #2562, opened 2016-09-20
fix macro missing build issue for avx2 in UT

huili2:
missing some macros for building UT solution related to AVX2 asm.
review see:
https://rbcommons.com/s/OpenH264/r/1704/

### DIFF
--- a/test/encoder/EncUT_DecodeMbAux.cpp
+++ b/test/encoder/EncUT_DecodeMbAux.cpp
@@ -239,10 +239,12 @@ TEST (DecodeMbAuxTest, WelsIDctT4Rec_mmx) {
 TEST (DecodeMbAuxTest, WelsIDctT4Rec_sse2) {
   TestIDctT4Rec<int16_t> (WelsIDctT4Rec_sse2);
 }
+#if defined(HAVE_AVX2)
 TEST (DecodeMbAuxTest, WelsIDctT4Rec_avx2) {
   if (WelsCPUFeatureDetect (0) & WELS_CPU_AVX2)
     TestIDctT4Rec<int16_t> (WelsIDctT4Rec_avx2);
 }
+#endif
 #endif
 template<typename clip_t>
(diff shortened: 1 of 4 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1704/

----
**testing_done**:  
**description**:  fix macro missing build issue for avx2 in UT

----
 reviewer Guangwei Wang (Guangwei) 
 time 2016-09-20T01:45:16+00:00 
2016-09-20T01:45:16+00:00 Guangwei Wang (Guangwei):
>Ship It!
Ship It!
Ship it!


----
 reviewer Karina Li (ruil2) 
 time 2016-09-20T03:14:36+00:00 
2016-09-20T03:14:36+00:00 Karina Li (ruil2):
 on **test/encoder/EncUT_DecodeMbAux.cpp line345**: 
 id="comment451671"><p>don't understand why HAVE_AVX2 isn't turned on in default.</p>
don't understand why HAVE_AVX2 isn't turned on in default.

2016-09-20T05:42:02+00:00 Guangwei Wang (Guangwei):
><p>Not necessarily true. In makefile, HAVE_AVX2 is enabled by default.<br />Anyway, it has nothing to do with this fix, which is to fix Windows UT build failure.</p>
I think the reason is:<br />1) keep the style consistent with our other macros. e.g. HAVE_ASM, HAVE_NEON, HAVE_X86 ...<br />2) basic code is a pure C/C++ code and is independent on any platform, our assembly code is a optimization for the basic code,so I think this should be not a default options for code.<br />3) althougt this macros HAVE_AVX2/HAVE_ASM... is not default for code, but this is default options for our complile options.It means the library compiled by default makefile will include the assembly code.
according your comments,now AVX2 UT code isn't tested. right?
Not necessarily true. In makefile, HAVE_AVX2 is enabled by default.<br />Anyway, it has nothing to do with this fix, which is to fix Windows UT build failure.


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1704/

PR #2566, opened 2016-09-22
mingw and msys2 build fix

BilyakA:
exported SHAREDLIBSUFFIXFULLVER variable to mingw-nt build

MSYS2 used to fail build process with `ln: failed to create symbolic link 'libopenh264.': No such file or directory` error.


GuangweiWang:
thanks for fix.

build/platform-mingw_nt.mk
```diff
@@ -1,7 +1,7 @@
 include $(SRC_PATH)build/x86-common.mk
 SHAREDLIB_DIR = $(PREFIX)/bin
 SHAREDLIBSUFFIX = dll
-SHAREDLIBSUFFIXVER=$(SHAREDLIBSUFFIX)
+SHAREDLIBSUFFIXFULLVER=$(SHAREDLIBSUFFIX)
```
GuangweiWang: please also export this variable together:
SHAREDLIBSUFFIXMAJORVER=$(SHAREDLIBSUFFIX)

BilyakA:
Added SHAREDLIBSUFFIXMAJORVER export to mingw-nt build.
I haven't found any changes in build process even with DEBUGSYMBOLS=True flag, but for future changes (for example: lib ver in filename on windows platforms) it may be useful.

GuangweiWang:
@BilyakA yes, you are right.
thanks for the fix!

GuangweiWang:
ship it

PR #2583, opened 2016-10-24
EHN: Make the console program importable by other program

leochan2009:
Dear OpenH264 community,

I am currently working on video compression development.  
The console program is very helpful, which is able to read configuration file. 
However, it is not possible to import the console program into other program.
So, i refactored the code, separate the main function part from the code. 
Hope this minor change could be adopted. 
Or if you have better solution, i would be more glad to know.

Best Regards,
Longquan


ruil2:
the console program only is a sample test program. You can do this change according to your application on your side. Don't suggest to merge into master branch.

PR #2594, opened 2016-11-11
fix parseonly issues when bs is very large

huili2:
when bs is too large that original BS memory is not enough, parse only functions will cause access violation.
Also when memory failed, reset the decoder.
see:
https://rbcommons.com/s/OpenH264/r/1728/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1728/

----
**testing_done**:  
**description**:  <p>fix parseonly issues when bs is very large<br />1. when sps+pps+nal data size is larger than given memory, add protection.<br />2. when failing, add reset function to prevent subsequent issues.<br />3. fix bugs in ExpandBsBuffer() to also expand parseonly required buffer.</p>

----
 reviewer Guangwei Wang (Guangwei) 
 time 2016-11-09T05:23:26+00:00 
2016-11-09T05:23:26+00:00 Guangwei Wang (Guangwei):
><p>it is ok for me</p>
it is ok for me
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1728/

PR #2607, opened 2016-11-27
fix wrong source pic(encoder) for multi-layers

PeterXu:
When encoding 4 layers(90/180/360/720p), the layer of 180p-T1 may be wrong.


ruil2:
it doesn't match the design that downsampling source will use the closest layer no matter that the temporal layer is valid or not.  I will submit a pull request, you can have a look.

ruil2:
please refer to https://github.com/cisco/openh264/pull/2608

PR #2617, opened 2016-12-07
Revert "remove too many log for IDR loss under EC is off"

Marahman1:
Reverts cisco/openh264#2558


ruil2:
why did this revert?

huili2:
No update till now. Close it.

PR #2633, opened 2017-01-02
Fixes for the pkg-config files

mstorsjo:
Fix the version field which is broken since September. Include libraries that are needed when using the static library on linux and BSD. Remove an unnecessary linker flag on Darwin.


GuangweiWang:
ship it!

PR #2647, opened 2017-01-16
Cygwin x86_64 support.

moriyoshi:
## Description

* This patch adds support for builds on Cygwin x86_64 platform.
* i386 builds aren't supported just as MinGW32.

## Notice

* I believe this patch does not contain any materials subject to the intellectual property laws as it isn't an independent creative work.


GuangweiWang:
ship it!

PR #2680, opened 2017-03-09
fixed RC UT faiked issue: verify with frame rate

shihuade:


huili2:
why comparing bitrate with framerate?

GuangweiWang:
the same question as Wayne, looks like the sInfo.iBitrate just a random number less 0.

PR #2689, opened 2017-03-20
Make SDKROOT immediately assigned

mstorsjo:
This avoids re-expanding and rerunning the xcrun command for each time
the SDKROOT variable is used (i.e. once per file built).


GuangweiWang:
ship

PR #2692, opened 2017-03-20
Readd the DecodeCrashTestAPI specific version of prepareEncDecParam

mstorsjo:
This was unexplainedly removed in d1c0a93, readd it with the
necessary adaptations.

This fixes valgrind errors about use of uninitialized data in
DecodeCrashTestAPI.DecoderCrashTest.


huili2:
I removed this as I remembered a base class function has exactly the same implementation. Not sure how this happens. Will check tomorrow.

mstorsjo:
Thanks for the pointer - then I found a better way to fix the issue, see #2694. I'll close this one then.

PR #2721, opened 2017-04-11
Fix arm downsampler to add horizontally first

mstorsjo:
When rounding inbetween each step, the order of the additions matter.

This fixes the testsuite when running on arm and arm64, after
a56b927135a61c4363a21d6e779f611de6304932.


GuangweiWang:
hi @mstorsjo , thanks a lot for your fix. 
Actually, downsampler intends to do vertically first, then horizontally. Just the C function has not been changed.

GuangweiWang:
I think we can merge this PR.
ship!

PR #2727, opened 2017-04-25
Fix warnings with MSVC in EncUT_SliceBufferReallocate.cpp

mstorsjo:


test/encoder/EncUT_SliceBufferReallocate.cpp
```diff
@@ -211,7 +211,7 @@ void CSliceBufferReallocatTest::InitParam() {
     pCtx->pSvcParam->iMultipleThreadIdc = (pCtx->pSvcParam->iMultipleThreadIdc <= 1 )? 2 : pCtx->pSvcParam->iMultipleThreadIdc;
     pCtx->iActiveThreadsNum = pCtx->pSvcParam->iMultipleThreadIdc;
     pCtx->pSvcParam->iSpatialLayerNum = 1;
-    pCtx->pSvcParam->bSimulcastAVC = (bool)rand() % 2;
```
GuangweiWang: is this a new warning or a old one?
why the force type conversion does not work on MSVC? 
thanks.

mstorsjo: It's a new warning, it appeared a few days ago. The warnings this commit fixes are the following:

test/encoder/EncUT_SliceBufferReallocate.cpp(214) : warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)
test/encoder/EncUT_SliceBufferReallocate.cpp(214) : warning C4804: '%' : unsafe use of type 'bool' in operation
test/encoder/EncUT_SliceBufferReallocate.cpp(214) : warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)
test/encoder/EncUT_SliceBufferReallocate.cpp(324) : warning C4244: 'initializing' : conversion from 'double' to 'int32_t', possible loss of data
test/encoder/EncUT_SliceBufferReallocate.cpp(551) : warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)
test/encoder/EncUT_SliceBufferReallocate.cpp(649) : warning C4800: 'int' : forcing value to bool 'true' or 'false' (performance warning)

I updated the commit message to include this.

The current code with the bool cast is also wrong, since it is applied before the modulo, that is, the current code does this: bool(rand()) % 2, while it probably meant to do bool(rand() % 2). But even in that case you'd still get the warning about forced conversion from int to bool. It only stops warning if you produce the bool with a comparison operator.

GuangweiWang: thanks

PR #2729, opened 2017-04-26
[Encoder] fix potential deadlock prob in multi-thread

sijchen:
https://rbcommons.com/s/OpenH264/r/1815/

previously we are asked to avoid no-named sem on Apple, so we use pcond to replace it. But simple replacing is not a complete solution since the usage of sem and pcond are different.
So we changed to current solution.

### DIFF
--- a/codec/common/inc/WelsThread.h
+++ b/codec/common/inc/WelsThread.h
@@ -82,7 +82,7 @@ class CWelsThread {
   }
 
   void SignalThread() {
-    WelsEventSignal (&m_hEvent);
+    WelsEventSignal (&m_hEvent, &m_hMutex, &m_iConVar);
   }
 
  private:
@@ -91,6 +91,7 @@ class CWelsThread {
   CWelsLock           m_cLockStatus;
   bool                m_bRunning;
   bool                m_bEndFlag;
+  int                 m_iConVar;
 
   DISALLOW_COPY_AND_ASSIGN (CWelsThread);
 };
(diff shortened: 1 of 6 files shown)


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1815/

----
**testing_done**:  
**description**:  <p>fix potential deadlock prob in multi-thread</p><p>fixmt1</p>

----
 reviewer Karina Li (ruil2) 
 time 2017-04-24T09:32:11+00:00 
2017-04-24T09:32:11+00:00 Karina Li (ruil2):
 on **codec/common/src/WelsThreadLib.cpp line343**: 
 id="comment534260"><p>Is it neccessary to add these two lines on other platform?</p>
Is it neccessary to add these two lines on other platform?

2017-04-25T00:30:54+00:00 Sijia Chen (sijchen):
><p>yes, since we move the dealing with the consition into this function. otherwise "*iCondition" will not be updated correctly.</p><p>see ln202 in codec/encoder/core/src/wels_task_management.cpp</p>
yes, since we move the dealing with the consition into this function. otherwise "*iCondition" will not be updated correctly.
see ln202 in codec/encoder/core/src/wels_task_management.cpp


----
 reviewer Karina Li (ruil2) 
 time 2017-04-25T05:03:31+00:00 
2017-04-25T05:03:31+00:00 Karina Li (ruil2):
>Ship It!
Ship It!
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1815/

PR #2738, opened 2017-05-02
    fix static block detection issue

ruil2:
reviewed at:https://rbcommons.com/s/OpenH264/r/1824/


sijchen:
Archiving from https://rbcommons.com/s/OpenH264/r/1824/

----
**testing_done**:  
**description**:  <p>fix static block detection issue and also fix scrolling potential issue.<br />1. in some cases, luma block is same, but chroma blocks are not same, so the MB can't be skipped.<br />2. when uses screen mode and disalbe ltr , the original blocks can't be got direclty from current structure, so use<br />   reference frame to detect whether the MB can be skipped or not.</p>

----
 reviewer wayne liu (huili2) 
 time 2017-05-02T02:10:27+00:00 
2017-05-02T02:10:27+00:00 wayne liu (huili2):
><p>seems OK to me.</p>
seems OK to me.
Ship it!


----

----

Finish archiving from https://rbcommons.com/s/OpenH264/r/1824/

PR #2847, opened 2017-09-22
bug fix on slice size calculation

Tee0125:
Since range of 'i' is 0 ~ iFileSize-1, 'iBufPos +i' can access outside of CPB when iBufPos is bigger than 0.

### DIFF
--- a/codec/console/dec/src/h264dec.cpp
+++ b/codec/console/dec/src/h264dec.cpp
@@ -181,13 +181,18 @@ void H264DecodeInstance (ISVCDecoder* pDecoder, const char* kpH264FileName, cons
         return;
       iSliceSize = static_cast<int32_t> (pInfo[2]);
     } else {
-      for (i = 0; i < iFileSize; i++) {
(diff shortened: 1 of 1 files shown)


huili2:
@Tee0125 In line 157, 
pBuf = new uint8_t[iFileSize + 4];
it has been protected by additional 4 bytes allocation. 
Thanks all the same for your contribution and look forward to more ideas.

Tee0125:
With 4 bytes padding, room checking can be eliminated but problem still exist

at least range of i should be modifed to 0, iFileSize-iBufPos-1

huili2:
@Tee0125 If you read through the above codes, you will find actually the padding 4 bytes are exactly "0001" and will guarantee the condition in line 185 to be true when iBufPos increases until "iBufPos+i" == iFileSize; then iBufPos += i to make iBufPos == iFileSize; soon line 172 will hit and "break". 
All above is to ensure iBufPos will never be bigger than iFileSize. This is a run-level protection with fewer codes, rather than static programmatic check.

Winly-40:
bug fix on slice size calculation

ruil2:
submit other pull request for this requirement.

PR #2871, opened 2017-11-22
Openh264v1.7.1

manticore3:


sijchen:
doesn't seem a reasonable PR to me. will close if no further updates seen

PR #2873, opened 2017-11-29
Allow the strip tool to be set by the caller

pedronavf:
When cross compiling sometimes the strip utility which is being used by default, which comes from the path, doesn't understand the architecture of the library it needs to strip, resulting in a failed build. Allow STRIP to be same in the same fashion as CC and CXX.


GuangweiWang:
ship

PR #2877, opened 2017-12-06
Fix Clang compiler error

evivrus510:
Clang compiler does not recognize .func and .endFunc.
Removed them.

### DIFF
--- a/codec/common/arm/arm_arch_common_macro.S
+++ b/codec/common/arm/arm_arch_common_macro.S
@@ -60,11 +60,17 @@ mov pc, lr
 .arm
 .global \funcName
 .type \funcName, %function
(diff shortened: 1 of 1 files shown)


GuangweiWang:
The .func directive can be used to tell the assembler that we're starting a function. This is used for debugging information. Any impact if moving the ".func" directives under other compiler or platform?

evivrus510:
Hi. I made a new change so that ".func" and ".endFunc" are only omitted when the compiler is Clang. Please take a look.

jbeich:
aarch64 is [still broken](https://ptpb.pw/bzZp) with Clang. Can you apply the same change against `codec/common/arm64/arm_arch64_common_macro.S`?

PR #2878, opened 2017-12-06
Enable NEON for ChromeOS devices

evivrus510:
Chrome uses openh264 to encode video frames for MediaRecorder APIs.
However, on ChromeOS, the encoding procedure is run in renderer process,
which can not access /proc/cpuinfo, making openH264 think that NEON is
not available.

Since all arm-based ChromeOS devices have NEON, it can be marked as
enabled anyway.


GuangweiWang:
ship

PR #2935, opened 2018-03-20
Fix WelsThreadLib.cpp to build under Fuchsia.

drwez:
Fuchsia's POSIX compatibility layer does not provide the two pthread calls used here.


alvestrand:
Pinging @fluffy to help with assignment

PR #2955, opened 2018-04-16
Add meson build definitions

MathieuDuponchelle:
Hi,

Please find attached a set of patches that adds support for building openh264 with the Meson build system.

I understand that from a maintainer's point of view these kind of unsolicited "add support for a different build system" patches are not always welcome, not least because they might add additional maintenance burden, so apologies in advance for that. I would be happy to help maintain them going forward, along with other GNOME folks.

The Meson build system is being adopted by projects such as GNOME, GTK or GStreamer.

It would be fantastic to have support for building openh264 with Meson in upstream openh264, as it would facilitate builds of GStreamer via Meson's subproject support.

Cross-compilation is of course also supported, but I have not tested that yet. Meson provides out-of-the-box support for building native binaries and cross binaries in the same build, which seems like something that could be useful for openh264.

To try the meson build:

$ meson --prefix=/tmp/prefix builddir
$ ninja -C builddir
$ ninja -C builddir test
$ ninja -C builddir install

You can install a recent Meson version with pip3 if your system does not have a recent one. You can also run Meson straight from a git checkout, since it's just python code.

Commit message follows:

```
This is not a complete port, but a good first step towards
using a simpler build system.

In their current state, the build definitions work and
have been tested on Linux and Windows (both against MSVC
and mingw). The build will fail explicitly on all other platforms.

No configuration options are currently exposed.

The test suite has been ported, and the README updated.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,17 @@ compiler:
 
 before_install:
   - sudo apt-get update -qq
-  - sudo apt-get install -qq nasm g++-multilib gcc-multilib libc6-dev-i386
+  - sudo apt-get install -qq nasm g++-multilib gcc-multilib libc6-dev-i386 python3-pip unzip
+  - sudo python3 -m pip install meson==0.44.1
+  - wget https://github.com/ninja-build/ninja/releases/download/v1.8.2/ninja-linux.zip
+  - unzip ninja-linux.zip
+  - export PATH=$PATH:$PWD
 
 install:
   - make gmp-bootstrap
   - make gtest-bootstrap
+  - meson builddir
+  - ninja -C builddir -v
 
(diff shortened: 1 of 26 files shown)


ruil2:
Could you add Meson build testing on travis CI? thanks!

MathieuDuponchelle:
> Could you add Meson build testing on travis CI? thanks!

Done, as said in the commit message this doesn't run the binary comparison tests, mostly because this is implemented in custom shell scripts which I'm a bit afraid to break, ideally these tests could be folded into the gtest test suite, but that's another topic :)

ruil2:
@MathieuDuponchelle  thanks for your contribution

PR #2965, opened 2018-04-29
Makefile: Use tr in a POSIX-conform way

strfry:
I stumbled upon this issue when trying to build on Alpine Linux, where busybox provides tools like 'tr'.
Busybox' tr doesn't interpret the :digit: character class, and instead literally removes the characters :digit from the input, resulting in the OS string "lnux"

0-9 should be a compatible and more standard-conform way.


GuangweiWang:
LGTM

PR #2971, opened 2018-05-15
Clip fFrameRate locally in EncoderInterfaceTest.EncoderOptionSetTest

mstorsjo:
For the cases where we pass an out of range framerate
to the encoder in the test, we already did test that the
encoder internally clipped the value to 60. Clip the local
variable as well, to make sure that later code in the test
does the right decisions based on fFrameRate.

This fixes a case where the check
    static_cast<float> (pParamExt->sSpatialLayers[sInfo.iLayer].iSpatialBitrate) <= fFrameRate
would indicate a failure, when the surrounding test used a
value of fFrameRate different than the encoder used internally.


mstorsjo:
Ping

PR #2977, opened 2018-06-16
Fix for typos in example of coding & decoding

pl0q1n:
It would be nice if someone could fix these typos in the wiki too.


sijchen:
LGTM

sijchen:
LGTM

PR #2980, opened 2018-06-20
Update WelsThreadLib.cpp

ashikns:
cppwinrt headers are included in the windows sdk from 17134 onwards. This gets rid of the C++/CX code in this class and hence can be compiled without the ZW flag, which is very helpful when integrating with other build systems. My personal use case here is building webrtc for uwp.

### DIFF
--- a/codec/common/src/WelsThreadLib.cpp
+++ b/codec/common/src/WelsThreadLib.cpp
@@ -74,9 +74,12 @@
 #ifdef WINAPI_FAMILY
 #if !WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
 #define WP80
-using namespace Platform;
-using namespace Windows::Foundation;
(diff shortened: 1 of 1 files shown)


mstorsjo:
I do see the value in removing the usage of the ZW build flag. However, I think a much less disruptive change would be to just skip using the WinRT threadpool altogether and use the same CreateThread API as for normal windows here. The CreateThread API was disallowed for phone/store apps for 8.0, but was allowed in 8.1 again.

To compare alternatives:
- This patch requires building with the very latest windows SDK
- Skipping the threadpool API would only disable building with MSVC 2012 targeting WP 8.0, still supporting 8.1.

So out of those two, I would very much prefer the latter.

mstorsjo:
Would you mind closing this PR, so there's no confusion on what to merge once someone is willing to take action?

Issue #3009, opened 2018-08-06
Legal(?) and size issues with bundled test clips


mstorsjo:
Ping @ethanhugg

The recently added support for b-frames also came with a few test samples, which turn out to be clips from actual movie trailers. Is this ok copyright wise for this project?

And even if they'd be ok license wise, these clips are way way too big for the testsuite. When running the whole testsuite in valgrind (in order to catch use of uninitialized data, reads/writes out of bounds etc, including in assembly which address sanitizer doesn't catch), these 4 test clips more than doubled the runtime of the testsuite, from around 30 minutes before, to 70 minutes now, with these 4 clips taking over 40 minutes to decode when run in valgrind.

Either the clips must be reduced, or I have to reconsider running regular testing of this project in valgrind.

fluffy:
I think we would prefer to use test clips which were well accepted test data that was  commonly used in to develop video codecs. That type of data has copyright that clearly allows us to use it and is also well known what it should look like to video experts so it tends to be easier to catch mistakes in.     How long are these clips right now? I suspect that there is no real gain to longer clips and to be able to speed up testing, we should probably use clips that are long enough to test the system but as short as possible.

mstorsjo:
The current tests for high profile are 200, 300, 300 and 580 frames long, but they are in 720p and 1080p, and weigh in at about 40 MB - imo total overkill for this test case. (And then the tests don't even cover CAVLC-encoded B-slices, which the pull request didn't contain support for.)

sijchen:
@xiaotianshi2 we will need to remove the test-clips for B-frames if the copy-right problem is not clear. Thanks.

xiaotiansf:
@sijchen If that's the issue, it had better to remove the trailer test cases.

GuangweiWang:
@xiaotiansf 
I have removed two test samples on the [PR#3010](https://github.com/cisco/openh264/pull/3010) 
can you help to check the other two files?

xiaotiansf:
@GuangweiWang 

I am sure these two files have no licence issue.
res/HighProfile_B_Frame_1920x1080p_2397fps.h264
res/HighProfile_B_Frame_1920x1080p_30fps.h264

mstorsjo:
@xiaotiansf License issues aside, can you produce samples with similar code coverage but shorter clips and much lower resolution? It doesn't need to be HD content just to test decoding of b-frames.

xiaotiansf:
@mstorsjo 
[...]

PR #3010, opened 2018-08-07
delete test samples which may clip from actual movie trailers

GuangweiWang:
as [issue#3009](https://github.com/cisco/openh264/issues/3009) mentioned.

@xiaotianshi2 I have removed this two move trailers: 
res/HighProfile_B_Frame_1280x544_2397p.h264
res/HighProfile_B_Frame_1280x720_2397p.h264

can you help to double check this two files if have the license issue or not? thanks.
res/HighProfile_B_Frame_1920x1080p_2397fps.h264
res/HighProfile_B_Frame_1920x1080p_30fps.h264


mstorsjo:
Not sure about license issues, but they are by far too large for the test case. They should be as small as possible while still covering the significant parts of the decoder. Imagine running the testsuite in valgrind, which makes things 100x slower. Imagine running the testsuite on ARM, in valgrind.

PR #3011, opened 2018-08-12
Fix Bugzillia reported Bug-1479669, Bug-1479670 and Bug-1479831. fix null pointers for some test cases when specific error conditions.

xiaotiansf:


AllenChong:
pull request

xiaotiansf:
@huili2 You are right. I missed this one and try again with my (just) checkin.

codec/decoder/core/src/cabac_decoder.cpp
```diff
@@ -145,6 +145,9 @@ int32_t DecodeBinCabac (PWelsCabacDecEngine pDecEngine, PWelsCabacCtx pBinCtx, u
   int32_t iRenorm = 1;
   uint32_t uiRangeLPS = g_kuiCabacRangeLps[uiState][ (uiRange >> 6) & 0x03];
   uiRange -= uiRangeLPS;
+  if (pDecEngine->iBitsLeft < 0) {
```
huili2: it is not appropriate to add judgment in this kernel function in terms of efficiency. Could you please find a better place to prevent branch here?

xiaotiansf:
Good point and let me see...

xiaotiansf:
That does seem to be quite challenging to not to add   
 if (pDecEngine->iBitsLeft < 0) {
    return GENERATE_ERROR_NO (ERR_LEVEL_MB_DATA, ERR_CABAC_NO_BS_TO_READ);
  }

according to call stack and I'll have to think about it.

xiaotiansf:
@huili2 
What do you think of the following code?  Will that be acceptable?

if (uiOffset >= (uiRange << (uint32_t)pDecEngine->iBitsLeft)) { //LPS
    uiOffset -= (uiRange << (uint32_t)pDecEngine->iBitsLeft);

huili2:
I was just thinking in what condition it would be negative. If it is the case, there should be somewhere error before stepping into this function...

xiaotiansf:
Right, I did check the call stacks to have tried to minimize the checking condition from topper of the call stacks. There is no chance to change in the toppers at all. I think C-style cast (uint32_t) seems to be the best solution without performance penalty and without changing the error condition later.

xiaotiansf:
This is resolved by adding WELS_READ_VERIFY 
at line 83 WELS_READ_VERIFY (DecodeTerminateCabac (pCabacDecEngine, uiCode)); of file parse_mb_syn_cabac.cpp. It no longer needs to add the judgement back in cabac_decoder.cpp:)

xiaotiansf:
codec/decoder/core/src/mv_pred.cpp:507:35 is also fixed by replacing with ST/LD.

xiaotiansf:
So you can try again now.

huili2:
tested, and seems OK to me now~

PR #3012, opened 2018-08-15
test CAVLC

chezhan3:


mstorsjo:
Could you similarly add test files for B-frames with CABAC, to replace the huge existing ones?

PR #3014, opened 2018-08-21
fix one more Bug-1485134 from Bugzilla reported. Add WELS_B_MB_REC_VE…

xiaotiansf:
…RIFY to verify if ref picture is lost

### DIFF
--- a/codec/decoder/core/inc/rec_mb.h
+++ b/codec/decoder/core/inc/rec_mb.h
@@ -48,6 +48,13 @@
 #include "decoder_context.h"
 
 namespace WelsDec {
+
+#define WELS_B_MB_REC_VERIFY(uiRet) do{ \
+  uint32_t uiRetTmp = (uint32_t)uiRet; \
+  if( uiRetTmp != ERR_NONE ) \
+    return uiRetTmp; \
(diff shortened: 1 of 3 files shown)


huili2:
It still reports other error like:
codec/decoder/core/src/deblocking.cpp:759:5: runtime error: index 14 out of bounds for type 'const int8_t [4]'

please check again.

xiaotiansf:
Sure, I'll fix it and I am  currently tied with other work right now but will try to fix as soon as I can.

xiaotiansf:
@huili2 Do you have Bugzilla bug number for the error: in codec/decoder/core/src/deblocking.cpp:759:5?

huili2:
exactly the same one as 1485134.

xiaotiansf:
That's strange and I don't see the run time error codec/decoder/core/src/deblocking.cpp:759:5 anymore after my latest PR #3014.

xiaotiansf:
I mean I don't see the run time error anymore with the attached sample stream in Bug1485134.

huili2:
Since you are not able to reproduce it (maybe my environment is different), let's bypass it and go on.

PR #3018, opened 2018-09-03
Avoid using C++/CX code for threads for Windows Phone/Windows Store/UWP

mstorsjo:
The CreateThread function was only unavailable in Windows (Phone) 8.0,
but became available again in Windows Phone 8.1 and for Store apps
in Windows 8.1.

This avoids building source files with the /ZW flag for C++/CX mode,
simplifying integration with other projects and build systems.

This is a better suggestion instead of #2980 by @ashikns.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,8 @@ If building for Windows Phone with MSVC 2013, there's no included bat file that 
 
 This is only necessary for building the DLL; the static library can be built without setting this.
 
+Note, only Windows Phone 8.1 or newer is supported, 8.0 is no longer supported.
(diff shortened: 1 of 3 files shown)


mstorsjo:
@ashikns - Does this look good to you?

ashikns:
Sorry for the late reply. Yes it looks good to me. In fact some of the the other upstream code I'm using has made the same change and has worked well for me.

mstorsjo:
Ping @huili2 @GuangweiWang or someone else who can merge this.

PR #3029, opened 2018-10-12
Remove static initializer in WelsThreadPool

tanderson-google:
This change fixes a static initializer that was introduced in the Chromium build
[1] on Mac and Linux.  For Mac in particular, Chromium does not allow any SIs.

[1] https://bugs.chromium.org/p/chromium/issues/detail?id=893594


mstorsjo:
For what it's worth, this looks acceptable to me - ping @huili2 @GuangweiWang or whoever can merge?

fluffy:
This looks good to me

PR #3031, opened 2018-10-17
add clang support for arm/arm64/x86 when NDK version is greater then 17

GuangweiWang:
as the title.

### DIFF
--- a/build/platform-android.mk
+++ b/build/platform-android.mk
@@ -47,6 +47,38 @@ CXXFLAGS += -fno-rtti -fno-exceptions
 LDFLAGS += --sysroot=$(SYSROOT)
 SHLDFLAGS = -Wl,--no-undefined -Wl,-z,relro -Wl,-z,now -Wl,-soname,lib$(PROJECT_NAME).so
 
+ifeq ($(NDK_TOOLCHAIN_VERSION), clang)
+  HOST_OS = $(shell uname -s | tr A-Z a-z)
+  CC = $(NDKROOT)/toolchains/llvm/prebuilt/$(HOST_OS)-x86_64/bin/clang
+  CXX = $(NDKROOT)/toolchains/llvm/prebuilt/$(HOST_OS)-x86_64/bin/clang++
+  AR = $(NDKROOT)/toolchains/llvm/prebuilt/$(HOST_OS)-x86_64/bin/llvm-ar
+  ifeq ($(ARCH), arm)
+  CFLAGS += -target armv7-none-linux-androideabi
+  CFLAGS += --sysroot=$(NDKROOT)/sysroot -I$(NDKROOT)/sysroot/usr/include/arm-linux-androideabi/
+  LDFLAGS += -target armv7a-none-linux-androideabi -isysroot $(SYSROOT)
+  LDFLAGS += -B $(NDKROOT)/toolchains/arm-linux-androideabi-4.9/prebuilt/$(HOST_OS)-x86_64/bin/arm-linux-androideabi-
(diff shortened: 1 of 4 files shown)


build/platform-android.mk
```diff
@@ -47,6 +47,38 @@ CXXFLAGS += -fno-rtti -fno-exceptions
...
+  CXX = $(NDKROOT)/toolchains/llvm/prebuilt/$(HOST_OS)-x86_64/bin/clang++
+  AR = $(NDKROOT)/toolchains/llvm/prebuilt/$(HOST_OS)-x86_64/bin/llvm-ar
+  ifeq ($(ARCH), arm)
+  CFLAGS += -target armv7-none-linux-androideabi
+  CFLAGS += --sysroot=$(NDKROOT)/sysroot -I$(NDKROOT)/sysroot/usr/include/arm-linux-androideabi/
+  LDFLAGS += -target armv7a-none-linux-androideabi -isysroot $(SYSROOT)
+  LDFLAGS += -B $(NDKROOT)/toolchains/arm-linux-androideabi-4.9/prebuilt/$(HOST_OS)-x86_64/bin/arm-linux-androideabi-
+  LDFLAGS += -L$(NDKROOT)/toolchains/arm-linux-androideabi-4.9/prebuilt/$(HOST_OS)-x86_64/lib/gcc/arm-linux-androideabi/4.9.x
+  else ifeq ($(ARCH), arm64)
+  CFLAGS += -target aarch64-none-linux-androideabi
```
mstorsjo: This shouldn't be `androideabi` but plain `android` here

build/platform-android.mk
```diff
@@ -47,6 +47,38 @@ CXXFLAGS += -fno-rtti -fno-exceptions
...
+  HOST_OS = $(shell uname -s | tr A-Z a-z)
+  CC = $(NDKROOT)/toolchains/llvm/prebuilt/$(HOST_OS)-x86_64/bin/clang
+  CXX = $(NDKROOT)/toolchains/llvm/prebuilt/$(HOST_OS)-x86_64/bin/clang++
+  AR = $(NDKROOT)/toolchains/llvm/prebuilt/$(HOST_OS)-x86_64/bin/llvm-ar
+  ifeq ($(ARCH), arm)
+  CFLAGS += -target armv7-none-linux-androideabi
+  CFLAGS += --sysroot=$(NDKROOT)/sysroot -I$(NDKROOT)/sysroot/usr/include/arm-linux-androideabi/
+  LDFLAGS += -target armv7a-none-linux-androideabi -isysroot $(SYSROOT)
+  LDFLAGS += -B $(NDKROOT)/toolchains/arm-linux-androideabi-4.9/prebuilt/$(HOST_OS)-x86_64/bin/arm-linux-androideabi-
+  LDFLAGS += -L$(NDKROOT)/toolchains/arm-linux-androideabi-4.9/prebuilt/$(HOST_OS)-x86_64/lib/gcc/arm-linux-androideabi/4.9.x
```
mstorsjo: I'm not a huge fan of hardcoding all these paths like this, especially as they are almost duplicated between all 4 architectures as well. Further up in this file, we call build-local.mk with DUMP_TOOLCHAIN_PREFIX to query the build system about what paths it is using; is it possible to do something like that to extract more of these paths that are needed?

GuangweiWang: @mstorsjo thanks your suggestion.

build/platform-android.mk
GuangweiWang: will fix this. thanks!

GuangweiWang:
@mstorsjo 
I create a new [PR#3032](https://github.com/cisco/openh264/pull/3032) for this change.

PR #3032, opened 2018-10-19
add clang support for arm/arm64/x86 when NDK version is greater then 17

GuangweiWang:
refine [PR#3031](https://github.com/cisco/openh264/pull/3031)

### DIFF
--- a/build/platform-android.mk
+++ b/build/platform-android.mk
@@ -5,7 +5,7 @@ SHAREDLIBSUFFIX = so
 SHAREDLIBSUFFIXFULLVER=$(SHAREDLIBSUFFIX)
 SHAREDLIBSUFFIXMAJORVER=$(SHAREDLIBSUFFIX)
 SHLDFLAGS =
-NDKLEVEL = $(shell echo $(TARGET) | grep -E -o "[0-9]+")
+NDKLEVEL = 12
 ifeq ($(ARCH), arm)
   ifneq ($(APP_ABI), armeabi)
     CFLAGS += -march=armv7-a -mfloat-abi=softfp
@@ -64,6 +64,8 @@ ifeq ($(NDK_TOOLCHAIN_VERSION), clang)
     TARGET_NAME = i686-none-linux-android
   else ifeq ($(ARCH), x86_64)
     TARGET_NAME = x86_64-none-linux-android
+  else
+    $(error "does not support this arch now!")
   endif
 
   CFLAGS += -target $(TARGET_NAME)
--- a/build/platform-android.mk
+++ b/build/platform-android.mk
@@ -5,7 +5,7 @@ SHAREDLIBSUFFIX = so
 SHAREDLIBSUFFIXFULLVER=$(SHAREDLIBSUFFIX)
 SHAREDLIBSUFFIXMAJORVER=$(SHAREDLIBSUFFIX)
 SHLDFLAGS =
-NDKLEVEL = 12
+NDKLEVEL = $(shell echo $(TARGET) | grep -E -o "[0-9]+")
 ifeq ($(ARCH), arm)
   ifneq ($(APP_ABI), armeabi)
     CFLAGS += -march=armv7-a -mfloat-abi=softfp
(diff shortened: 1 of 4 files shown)


build/platform-android.mk
```diff
@@ -37,16 +37,23 @@ $(error TARGET is not set)
...
 TOOLCHAINPREFIX = $(shell NDK_PROJECT_PATH=$(SRC_PATH)/codec/build/android/dec make --no-print-dir -f $(NDKROOT)/build/core/build-local.mk DUMP_TOOLCHAIN_PREFIX APP_ABI=$(APP_ABI))
+TOOLCHAIN_NAME = $(shell NDK_PROJECT_PATH=$(SRC_PATH)/codec/build/android/dec make --no-print-dir -f $(NDKROOT)/build/core/build-local.mk DUMP_TOOLCHAIN_NAME APP_ABI=$(APP_ABI))
+GCC_TOOLCHAIN_PATH = $(shell dirname $(TOOLCHAINPREFIX) | xargs dirname )
 
 SYSROOT = $(NDKROOT)/platforms/android-$(NDKLEVEL)/arch-$(ARCH)
 CXX = $(TOOLCHAINPREFIX)g++
 CC = $(TOOLCHAINPREFIX)gcc
 AR = $(TOOLCHAINPREFIX)ar
 CFLAGS += -DANDROID_NDK -fpic --sysroot=$(SYSROOT) -MMD -MP
+CFLAGS += -isystem $(NDKROOT)/sysroot/usr/include -isystem $(NDKROOT)/sysroot/usr/include/$(TOOLCHAIN_NAME)
```
mstorsjo: Should we maybe only add this if `NDK_TOOLCHAIN_VERSION` is `clang`? On the other hand, I think it might be harmless to add on older NDK versions even if this directory doesn't exist.

As for the second directory added, I don't see any `sysroot/usr/include/clang` directory anywhere in my NDK - does that really exist in some version, or is this a mistake?

GuangweiWang: thanks for your comments.
for your questions:
1. I think it is harmless to add it for older NDK version. (when NDK version is greater then r14, all the header files have been moved to this two path: a. $NDK/sysroot/usr/include b.$NDK/sysroot/usr/include/$(TOOLCHAIN_NAME))
2. $(TOOLCHAIN_NAME) is not equal to $(TOOLCHAIN_VERSION) for NDK. the value of $(TOOLCHAIN_NAME) is:
arm : arm-Linux-androideabi
arm64: aarch64-linux-android
x86: i686-linux-android
x86_64: x86_64-linux-android

mstorsjo: Ah, I see - I didn't notice the difference between `$(NDK_TOOLCHAIN_VERSION)` and `$(TOOLCHAIN_NAME)` here. So these issues I raised are ok then.

So I presume that `$(NDK_TOOLCHAIN_VERSION)` is something that the user needs to manually set by passing `NDK_TOOLCHAIN_VERSION=clang` when compiling, right? That's probably ok, but is there possibly any way to find out what is the default for that also in the same way as we query the other values? If not this is fine as is.

mstorsjo: Ah, I see - I didn't notice the difference between `$(NDK_TOOLCHAIN_VERSION)` and `$(TOOLCHAIN_NAME)` here. So these issues I raised are ok then.

So I presume that `$(NDK_TOOLCHAIN_VERSION)` is something that the user needs to manually set by passing `NDK_TOOLCHAIN_VERSION=clang` when compiling, right? That's probably ok, but is there possibly any way to find out what is the default for that also in the same way as we query the other values? If not this is fine as is.

build/platform-android.mk
```diff
@@ -37,16 +37,23 @@ $(error TARGET is not set)
 endif
 
 TOOLCHAINPREFIX = $(shell NDK_PROJECT_PATH=$(SRC_PATH)/codec/build/android/dec make --no-print-dir -f $(NDKROOT)/build/core/build-local.mk DUMP_TOOLCHAIN_PREFIX APP_ABI=$(APP_ABI))
+TOOLCHAIN_NAME = $(shell NDK_PROJECT_PATH=$(SRC_PATH)/codec/build/android/dec make --no-print-dir -f $(NDKROOT)/build/core/build-local.mk DUMP_TOOLCHAIN_NAME APP_ABI=$(APP_ABI))
+GCC_TOOLCHAIN_PATH = $(shell dirname $(TOOLCHAINPREFIX) | xargs dirname )
```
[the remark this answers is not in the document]
GuangweiWang: I think there is a little difference between this two.
e.g. cur_path="/home/foo/bar/"

$(shell dirname ($cur_path)) => "/home/foo"
$(dir $(cur_path))                   => "/home/foo/bar"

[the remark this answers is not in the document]
mstorsjo: Ok, apparently the make `dir` function doesn't work in the same way then - I guess this is fine then.

build/platform-android.mk
GuangweiWang: yes. the user should manually pass NDK_TOOLCHAIN_VERSION=clang when compiling. 
of course, we can query this value and set it as the default value. But I think this is not a good choice for some versions. e.g. for NDK-r14, the default value of NDK_TOOLCHAIN_VERSION is clang, but they still use GCC toolchain as the default tool. Maybe we can take this choice for higher NDK version(e.g. NDK version greater than r17), but I think it is not a good choice for right now.

mstorsjo: Ok, fair enough - then I'm ok with this change as it is right now. Thanks for taking the time to explain your choices!

GuangweiWang: Thank you for taking the time to review this code.

build/platform-android.mk
```diff
@@ -45,13 +45,29 @@ CXX = $(TOOLCHAINPREFIX)g++
...
+  CXX = $(LLVM_INSTALL_DIR)/clang++
+
+  ifeq ($(ARCH), arm)
+    TARGET_NAME = armv7-none-linux-androideabi
+  else ifeq ($(ARCH), arm64)
+    TARGET_NAME = aarch64-none-linux-android
+  else ifeq ($(ARCH), x86)
+    TARGET_NAME = i686-none-linux-android
+  else ifeq ($(ARCH), x86_64)
+    TARGET_NAME = x86_64-none-linux-android
```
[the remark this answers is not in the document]
GuangweiWang: will add.

build/platform-android.mk
```diff
@@ -5,7 +5,7 @@ SHAREDLIBSUFFIX = so
 SHAREDLIBSUFFIXFULLVER=$(SHAREDLIBSUFFIX)
 SHAREDLIBSUFFIXMAJORVER=$(SHAREDLIBSUFFIX)
 SHLDFLAGS =
-NDKLEVEL = 12
+NDKLEVEL = $(shell echo $(TARGET) | grep -E -o "[0-9]+")
```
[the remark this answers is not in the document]
GuangweiWang: yes, the user normally only pass e.g. TARGET=android-21 but does not pass the NDKLEVEL. 
1. default NDKLEVEL=12 maybe too old.
2. and I think pass TARGET=andorid-21 but actually use NDKLEVEL=12 is bit strage.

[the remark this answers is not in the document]
mstorsjo: Well there might be one reason for it; currently TARGET is used as the target for the java code. For the java side of android apps, you generally should set the target SDK version to the highest available, while you set the minimum SDK version to the minimum you ever want it to run on. For native code, you can only set one version, the minimum. So therefore it might make sense to keep them separate since you generally should not build with the NDKLEVEL set to the highest.

I guess this tries to fix the fact that newer NDK versions don't contain all the old platform versions any longer? In that case I think it might be better to just bump up the default value to whichever is the minimum supported by current NDK versions.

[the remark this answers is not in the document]
GuangweiWang: ok, let the NDKLEVEL is not changed. will fix this.

mstorsjo:
This looks ok to me now.

GuangweiWang:
@mstorsjo 
thanks a lot.

GuangweiWang:
@mstorsjo I merged this three commit to one and open a new [PR#3033](https://github.com/cisco/openh264/pull/3033) for it.
This PR will be closed later. thanks.

PR #3034, opened 2018-10-30
Bug: fix bug that under some lossy conditions ltr won't update

zjudongze:


codec/encoder/core/src/ref_list_mgr_svc.cpp
```diff
@@ -159,9 +159,8 @@ static inline void DeleteInvalidLTR (sWelsEncCtx* pCtx) {
   SLogContext* pLogCtx = & (pCtx->sLogCtx);
 
   for (i = 0; i < LONG_TERM_REF_NUM; i++) {
```
ruil2: which case will trigger this issue?

codec/encoder/core/src/ref_list_mgr_svc.cpp
```diff
@@ -545,13 +542,12 @@ int32_t FilterLTRRecoveryRequest (sWelsEncCtx* pCtx, SLTRRecoverRequest* pLTRRec
...
 
         pLtr->bReceivedT0LostFlag = true;
-        pLtr->iLastCorFrameNumDec = pRequest->iLastCorrectFrameNum;
-        pLtr->iCurFrameNumInDec = pRequest->iCurrentFrameNum;
         WelsLog (& (pCtx->sLogCtx), WELS_LOG_INFO,
                  "Receive valid LTR recovery pRequest,feedback_type = %d ,uiIdrPicId = %d , current_frame_num = %d , last correct frame num = %d"
                  , pRequest->uiFeedbackType, pRequest->uiIDRPicId, pRequest->iCurrentFrameNum, pRequest->iLastCorrectFrameNum);
       }
-
+      pLtr->iLastCorFrameNumDec = pRequest->iLastCorrectFrameNum;
```
ruil2: why did you need to change the update logic?

izabfixit4u:
I'm very new to this. For some reason my beloved Firefox browser is causing me all kinds of problems it keeps blocking me out as well I'm cracking up because you don't see it till you have finished and realise there's no data x

PR #3046, opened 2018-11-26
enable stack protection

GuangweiWang:
as the title.


fluffy:
What performance impact would this have ?

PR #3049, opened 2018-11-27
spell error

elesos:


huili2:
@elesos thanks for your contribution on openh264. Could you please make similar changes into one PR for simplicity?

PR #3061, opened 2018-12-18
Add clean CAVLC B-frame decoder support.

xiaotianshicomcast:


codec/decoder/core/inc/decode_slice.h
```diff
@@ -1,99 +1,103 @@
...
-void WelsBlockZero16x16_c (int16_t* block, int32_t stride);
-void WelsBlockZero8x8_c (int16_t* block, int32_t stride);
-
-} // namespace WelsDec
-
-#endif //WELS_DECODE_SLICE_H__
-
-
+/*!
+ * \copy
```
thaytan: This header is almost all whitespace change because of switching to Windows line endings. I'm not sure why that's happened just to this file and parse_mb_syn_cavlc.h and not others.

codec/decoder/core/src/mv_pred.cpp
```diff
@@ -320,6 +320,10 @@ int32_t GetColocatedMb (PWelsDecoderContext pCtx, MbType& mbType, SubMbType& sub
   }
 
   MbType coloc_mbType = colocPic->pMbType[iMbXy];
+  if (coloc_mbType == MB_TYPE_SKIP) {
+    //This indicates the colocated MB is P SKIP MB
+    coloc_mbType |= MB_TYPE_16x16 | MB_TYPE_P0L0 | MB_TYPE_P1L0;
+  }
```
thaytan: This change doesn't seems like a general fix, not CAVLC specific. It should be in a separate commit?

codec/decoder/plus/src/welsDecoderExt.cpp
```diff
@@ -744,7 +744,7 @@ void CWelsDecoder::OutputStatisticsLog (SDecoderStatistics& sDecoderStatistics)
 
 DECODING_STATE CWelsDecoder::ReorderPicturesInDisplay (unsigned char** ppDst, SBufferInfo* pDstInfo) {
   if (pDstInfo->iBufferStatus == 1 && m_pDecContext->pSps->uiProfileIdc != 66
-      && m_pDecContext->pPps->bEntropyCodingModeFlag) {
+      /*&& m_pDecContext->pPps->bEntropyCodingModeFlag*/) {
```
thaytan: I think it would be better to delete the bEntropyCodingModeFlag check entirely. It's just clutter now.

sijchen:
close this for conflicts.

PR #3064, opened 2018-12-19
Add clean CAVLC B-Frame decoder support without whitespace change and…

xiaotianshicomcast:
… general fix.


mstorsjo:
Thanks for getting this done!

With this done, I'm still seeing issues with reordering on lots of clips. One test clip is e.g. 
https://download.blender.org/peach/bigbuckbunny_movies/big_buck_bunny_480p_h264.mov, which displays the same phenomenon as on many other samples - frames are returned in the wrong order with non-increasing timestamps. The reordering does however work on one small sample though, so the libavcodec wrapper that I'm using should be working in general.

thaytan:
I've been looking at the output re-ordering issue. It's a problem with some clips, but I'm not sure what the trigger is yet. We should track that in a separate issue

PR #3065, opened 2018-12-19
Bugzilla 1479656, 1485229 and 1486223 fixes

xiaotianshicomcast:


mstorsjo:
This pull request also contains a version of the CAVLC B-frame support. As that is merged now, perhaps it's ok though, but it would be nicer to have it on a new plain branch without any extra commits.

huili2:
This has duplicated code as other ones? Need close this one or not?

xiaotiansf:
@sijchen I created a separate #PR3085 which only includes the Bugzilla fixes. So this #PR 3065 can be removed.

huili2:
close this one as it is done by other PRs.

PR #3068, opened 2018-12-20
Add user level malloc/free support

kevincoyle2018:
Added a way via ISVCDecoder::SetOption to register user provided malloc/free.   The signature differs from void(*)(size_t) since it looks like current practice at this API level is to not use size_t and other typedefs.  Additionally, internal malloc equivalents WelsMalloc/WelsMallocz take a uint32_t instead of size_t so the signature is consistent across levels.

This allows application writers to hook into memory allocation without modifying prebuilt binaries.


kevincoyle2018:
Addressing this via a different API now

PR #3080, opened 2019-01-03
Update and fix temporal direct prediction and fix multi-reference pic…

xiaotiansf:
…ure re-ordering. Pending to fix possible decoded B-frame overwritten.


huili2:
not quite code review done, but mostly seems OK to me.

PR #3085, opened 2019-01-08
Bugzilla 1479656, 1485229 and 1486223 fixes without mixing with other…

xiaotiansf:
… update.


xiaotiansf:
@huili2 The removed lines 213 "else {..." does look correct. But I realized that the same code must also be inserted right after line (now) 303, which make the code inefficient. Since this condition only happens at a new sequence beginning, It is much easier and better to do it in lines 829-832 in "decoder.cpp". That's why.

PR #3091, opened 2019-01-30
Additional fix for Bugzilla reported bug-1522172. Do not try to contin…

xiaotiansf:
…ue B-frame temporal direct prediction when previously encountered a concealed errors.


codec/decoder/core/src/decode_slice.cpp
```diff
@@ -2371,6 +2371,12 @@ int32_t WelsDecodeMbCavlcBSlice (PWelsDecoderContext pCtx, PNalUnit pNalCur, uin
     memset (pCurLayer->pRefIndex[LIST_1][iMbXy], 0, sizeof (int8_t) * 16);
     pCtx->bMbRefConcealed = pCtx->bRPLRError || pCtx->bMbRefConcealed || ! (ppRefPicL0[0] && ppRefPicL0[0]->bIsComplete)
                             || ! (ppRefPicL1[0] && ppRefPicL1[0]->bIsComplete);
+
+    if (pCtx->bMbRefConcealed) {
+      SLogContext* pLogCtx = & (pCtx->sLogCtx);
+      WelsLog (pLogCtx, WELS_LOG_ERROR, "Ref Picture for B-Slice is lost, B-Slice decoding cannot be continued!");
```
huili2: WELS_LOG_ERROR may be flooding if actual ref pic is lost. Can consider to remove the log in the future.

PR #3092, opened 2019-01-30
External allocator and configurable bitstream buffer

kevincoyle2018:
Update SDecodingParam to accept a pointer to an external memory allocator.

Also allow the bitstream buffer to be configurable instead of being fixed to 3 MB.

### DIFF
--- a/codec/common/inc/memory_align.h
+++ b/codec/common/inc/memory_align.h
@@ -37,7 +37,7 @@
 #include "codec_app_def.h"
 
 // NOTE: please do not clean below lines even comment, turn on for potential memory leak verify and memory usage monitor etc.
-#define MEMORY_CHECK
+//#define MEMORY_CHECK
 #define MEMORY_MONITOR
 #ifdef MEMORY_CHECK
 #ifndef MEMORY_MONITOR
@@ -49,7 +49,6 @@
 #ifdef MEMORY_CHECK
 #include <stdio.h>
 #endif//MEMORY_CHECK
-
 
 namespace WelsCommon {
 
--- a/codec/common/inc/memory_align.h
+++ b/codec/common/inc/memory_align.h
@@ -34,9 +34,10 @@
 #define WELS_COMMON_MEMORY_ALIGN_H__
 
 #include "typedefs.h"
+#include "codec_app_def.h"
 
 // NOTE: please do not clean below lines even comment, turn on for potential memory leak verify and memory usage monitor etc.
-//#define MEMORY_CHECK
+#define MEMORY_CHECK
 #define MEMORY_MONITOR
 #ifdef MEMORY_CHECK
 #ifndef MEMORY_MONITOR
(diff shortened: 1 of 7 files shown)


codec/common/inc/memory_align.h
```diff
@@ -34,9 +34,10 @@
 #define WELS_COMMON_MEMORY_ALIGN_H__
 
 #include "typedefs.h"
+#include "codec_app_def.h"
 
 // NOTE: please do not clean below lines even comment, turn on for potential memory leak verify and memory usage monitor etc.
-//#define MEMORY_CHECK
+#define MEMORY_CHECK
```
thaytan: I don't think you meant to leave this on here.

codec/console/dec/src/h264dec.cpp
```diff
@@ -479,8 +479,6 @@ int32_t main (int32_t iArgC, char* pArgV[]) {
   }
 
 
-
-
```
thaytan: Unnecessary whitespace changes here in h264dec.cpp

codec/decoder/plus/src/welsDecoderExt.cpp
```diff
@@ -234,19 +234,19 @@ void CWelsDecoder::UninitDecoder (void) {
...
+    pSMemAllocator->WelsFree (m_pDecContext, "m_pDecContext");
     m_pDecContext = NULL;
   }
+
+  if (NULL != pSMemAllocator) {
+    WelsLog (&m_pWelsTrace->m_sLogCtx, WELS_LOG_INFO,
+             "CWelsDecoder::UninitDecoder(), verify memory usage (%d bytes) after free..",
+             pSMemAllocator->WelsGetMemoryUsage());
+    delete pSMemAllocator;
+  }
```
thaytan: You should set m_pDecContext->pMemAlign back to NULL here, as UninitDecoder() can be called multiple times potentially (if InitDecoder fails for some reason). You therefore also can't assume that pSMemAllocator is non-NULL on entry.

kevincoyle2018:
External allocation allows access to heaps not part of system memory.  For example, some silicon vendors may partition physical memory where X is available via standard Linux calls and reserve the rest for things like hardware video decoder buffers and so on.  However, since this partitioning occurs at device deployment time (think a set top box or TV) and is not adjustable dynamically, this memory may go to waste.  These vendors would have an API to allocate/deallocate/map that memory.

Additionally, it allows introspection into memory usage instead of relying on a developer to use a specialized tool (build of h264dec with MEMORY_CHECK).  For example, the allocator may be queriable via some JS bridge allowing a JS developer to determine quickly that memory usage is  higher than some planned budget and empower him to speak with an in-house content creator.

codec/console/dec/src/h264dec.cpp
kevincoyle2018: Thanks, restored to original state.

PR #3093, opened 2019-02-05
Fix Bugzilla reported bug-1525097.

xiaotiansf:


codec/decoder/core/src/decode_slice.cpp
```diff
@@ -2375,7 +2375,7 @@ int32_t WelsDecodeMbCavlcBSlice (PWelsDecoderContext pCtx, PNalUnit pNalCur, uin
     if (pCtx->bMbRefConcealed) {
       SLogContext* pLogCtx = & (pCtx->sLogCtx);
       WelsLog (pLogCtx, WELS_LOG_ERROR, "Ref Picture for B-Slice is lost, B-Slice decoding cannot be continued!");
-      return GENERATE_ERROR_NO (ERR_LEVEL_SLICE_DATA, ERR_INFO_REFERENCE_PIC_LOST);
+      return GENERATE_ERROR_NO (ERR_LEVEL_SLICE_DATA, ERR_INFO_REFERENCE_PIC_LOST) | dsOutOfMemory;
```
huili2: This kind of error differs from dsOutOfMemory. You should not add it here. If required, consider changing ctx->errorcode to it on higher level.

codec/decoder/plus/src/welsDecoderExt.cpp
```diff
@@ -581,6 +581,12 @@ DECODING_STATE CWelsDecoder::DecodeFrame2 (const unsigned char* kpSrc,
 
     eNalType = m_pDecContext->sCurNalHead.eNalUnitType;
 
+    if ((m_pDecContext->iErrorCode & dsRefLost) && m_pDecContext->eSliceType == B_SLICE) {
+      if (ResetDecoder())
```
sijchen: could you keep the coding style consistent by adding {} in this conditional? Thx

xiaotiansf:
Fix Bugzilla reported bug 1522172_2 (second test case) and 1525089.

PR #3094, opened 2019-02-05
meson: Fix generation of import library on Windows

nirbheek:
Add the `.def` file when generating the DLL so that the required symbols are exported when building with MSVC and an import library is generated.

Without this, you can't use the openh264 DLL generated with the meson build files.


MathieuDuponchelle:
lgtm

nirbheek:
I had to force-push multiple times because the CI kept erroring out while setting up the image.

PR #3095, opened 2019-02-07
Remove check for working compiler in NDK

dminor:
For Firefox, we use our own build of clang for Android builds. The version of clang shipped in the Android NDK will not run on our build machines because it uses an incompatible version of glibc. Removing this check allows us to compile the plugin on our build machines. If there is no working compiler available, we'll just hit errors further along in the build, so I don't think there's any harm in removing this check.

### DIFF
--- a/build/platform-android.mk
+++ b/build/platform-android.mk
@@ -73,6 +73,14 @@ ifeq ($(NDK_TOOLCHAIN_VERSION), clang)
   LDFLAGS += -Wl,--exclude-libs,libgcc.a -Wl,--exclude-libs,libunwind.a
 endif
 
+ifneq ($(findstring /,$(CXX)),$(findstring \,$(CXX)))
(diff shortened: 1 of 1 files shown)


mstorsjo:
The reason for this check is that history proved that it isn't totally easy for users to point correctly to the NDK, and the error messages you get if it didn't pick up the compiler correctly are rather obscure.

Does it sounds sensible to check for slashes in `$(CXX)` and only run this check if found? If no slashes are found, it's presumed that `$(CXX)` is a tool that should be found in `$PATH` by the shell.

That should be doable by enclosing the existing check with `ifneq ($(findstring /,$(CXX)$(findstring \,$(CXX),)`.

mstorsjo:
If it proves too hard/messy though I won't object to removing the check.

dminor:
That makes sense, I will give that a try. Thank you!

PR #3098, opened 2019-02-09
Fix Bugzilla reported bug 1526408.

xiaotiansf:


xiaotiansf:
This seems also fixed Bugzilla reported bug 1526427.

PR #3100, opened 2019-02-13
Fix Issues 13030 and 13039 in oss-fuzz.

xiaotiansf:


codec/decoder/core/src/decode_slice.cpp
```diff
@@ -56,6 +56,22 @@ extern void FreePicture (PPicture pPic, CMemoryAlign* pMa);
 
 extern PPicture AllocPicture (PWelsDecoderContext pCtx, const int32_t kiPicWidth, const int32_t kiPicHeight);
 
+static bool CheckRefPics (const PWelsDecoderContext& pCtx) {
+  int32_t listCount = 1;
```
huili2: So it is only for P frame decoding? Why it occurs now since P frame is always on?

codec/decoder/core/src/decode_slice.cpp
```diff
@@ -318,7 +334,10 @@ int32_t WelsTargetMbConstruction (PWelsDecoderContext pCtx) {
     WelsMbIntraPredictionConstruction (pCtx, pCurLayer, 1);
   } else if (IS_INTER (pCurLayer->pMbType[pCurLayer->iMbXyIndex])) { //InterMB
     if (0 == pCurLayer->pCbp[pCurLayer->iMbXyIndex]) { //uiCbp==0 include SKIP
-      WelsMbInterPrediction (pCtx, pCurLayer);
+      if (!CheckRefPics (pCtx)) {
```
huili2: Is this function necessary or too late to check here? Suppose it should be checked in parse process rather than in construction process.

xiaotiansf: Not quite sure. It is a strange test case. With P Slice, pCtx->sRefPic.uiRefCount > 0 while pCtx->pRefPic.pRefList[LIST_0] is empty.

xiaotiansf: It is weird that pCtx->pRefPic.pRefList[LIST_1][0] is not null (but uiRefCount[LIST_1] = 0) while pCtx->pRefPic.pRefList[LIST_0][0] is null (uiRefCount[LIST_0] = 1) for the P Slice for this testcase.

huili2: For me, it seems not a perfect solution here. I think there should be somewhere to place better changes. I can agree for now. But need dig further.

PR #3104, opened 2019-03-01
enable stack protector

GuangweiWang:
as the title.


rdp:
I believe this is causing  undefined symbol __stack_chk_fail' on mingw https://github.com/rdp/ffmpeg-windows-build-helpers/issues/380#issuecomment-473605252

PR #3113, opened 2019-03-15
deblocking: Disambiguate reference pictures

thaytan:
Macroblocks can refer to the same reference picture using
    different indices when the same picture appears in multiple
    slots in the current reference picture lists. That leads to
    calculating the wrong loop filter strength for edge processing,
    and small errors in the conformance suite on (for example)
    CVWP5_TOSHIBA_E.264 and others.
    
    Modify the test-suite and checksums for affected test files
    where this situation occurs.

Note that the 6 test files in question still don't produce the same pixel output as the reference decoder, so there's more work to do.


huili2:
Many thanks for your contribution! Longing for your future work on solving these test files.

PR #3117, opened 2019-03-29
Fix oss-fuzz reported bug 13941 Heap-use-after-free in WelsDec::MapCo…

xiaotiansf:
…lToList0.


xiaotiansf:
The fix for the bug 13941 also fixed the issue 13002  (which was reported on Feb. 8) but I never get informed.

PR #3118, opened 2019-04-03
Fix oss-fuzzer 14035. Reset all B-frame's reference's references when…

xiaotiansf:
… IDR is lost to prevent temporal prediction from trying to access lost references.


codec/decoder/core/src/manage_dec_ref.cpp
```diff
@@ -126,6 +126,14 @@ static int32_t WelsCheckAndRecoverForFutureDecoding (PWelsDecoderContext pCtx) {
         pRef->bIsComplete = false; // Set complete flag to false for lost IDR ref picture
         pRef->iSpsId = pCtx->pSps->iSpsId;
         pRef->iPpsId = pCtx->pPps->iPpsId;
+        if (pCtx->eSliceType == B_SLICE) {
+          //reset reference's references when IDR is lost
+          for (int32_t list = LIST_0; list < LIST_A; ++list) {
+            for (int32_t i = 0; i < 17; ++i) {
```
sijchen: could you please avoid using magic numbers in codes? Thx

PR #3123, opened 2019-04-20
Fix oss-fuzz newly reported bug 14267. reset reference's references a…

xiaotiansf:
…fter call DecreasePicBuff when new sequence arrives


huili2:
Please pay attention to the commit number next time.

PR #3150, opened 2019-07-10
Submit two patches for loongson platform.

gxw-loongson:
patch 1:  add initialization of mmi for m_pfSad.
patch 2:  when testing the latest code on loongaon 3a3000,  the performance of encoder/decoder is poor. The reason is unaligned access in BACKUP_REG. Fix it in this patch.


gxw-loongson:
@GuangweiWang，could you help me review the patch, thanks.

PR #3164, opened 2019-08-28
* Added support for iOS with arm64e

agah-pt:


GuangweiWang:
@agah-pt why this travis testing is still pending?

GuangweiWang:
I had understood that arm64e is corresponding to the ARMV8.3. is it compatible with arm64?

PR #3177, opened 2019-10-11
All 4-commits of multi-thread decoding support

xiaotiansf:
This is the first commit of re-organized commits for multi-thread decoding support.
It contains re-structured decoder_context and others for thread decoding support.


huili2:
I prefer this kind of PRs :), and will be happier to see a single commit rather than 4 commits with some "merge" commits into them.

xiaotiansf:
I see. So next PR for next commit will have to wait for the current commit (PR) to be merged, right?

huili2:
If they are independent, we do not need to wait. Otherwise, I hope so....

xiaotiansf:
next commit will always depend on its preceeding commit.

xiaotiansf:
@huili2 Let me know as soon as you have reviewed the current commit. I have made next two commits ready to submit. Thx.

huili2:
Yes, I'm looking at the code diffs, and will let you know in time.

xiaotiansf:
Thanks. 
I also have gotten the 4th commit (which is the final commit) ready for you.
preliminary test shows that the decoding speed is nx where n (thread count) is >= 2. Performance of for decoding is measured with the time spent on API "DecodeFrameNoDelay". The result is consistent on OSX, Linux and Windows. Note that no significant performance improvement is observed if thread count n > 3.

huili2:
Seems OK. I still suggest you combine all your 4 PR into one as 3 out of them are "merge" commits.
btw: If possible, could you please add your multi-thread modules into common folder? There's already some kind of multi-thread modules for common usage, not only for decoder. That can be used in future.

xiaotiansf:
Sounds good. As for multi-thread modules, I have not made the decoding to use the event defined in WelsThreadLib.h correctly. Yes, eventually I'll completely merge them into WelsThreadLib.h in common.

huili2:
@xiaotiansf  sorry seems I did not explain "combine all your 4 PR" from your behavior. I mean, previously you had 4 commits in this PR, and I hoped you can make a single comit for clear purpose. Then after this PR is merged, you can raise new PR one by one as mentioned before.
However, you now merged new 4 commits into this one, making this PR more complex as before. That's becoming the same as #3174 did.

xiaotiansf:
@huili2 I see and I misunderstood your intention. Let me re-do commits then.

xiaotiansf:
close this one due to mistake and refer #3179.

PR #3179, opened 2019-10-16
Multi-thread support

xiaotianshi2:
commit 1: re-structure SWelsDecoderContext and others and prepare for the support of multi-threading support.


huili2:
seems OK for me, longing for future UT.

PR #3180, opened 2019-10-17
Detect Loongson-MMI at runtime

FlyGoat:
Just like what Arm did.

### DIFF
--- a/build/arch.mk
+++ b/build/arch.mk
@@ -30,14 +30,16 @@ CFLAGS += -DHAVE_NEON_AARCH64
 endif
 endif
 
-#for loongson
+#for mips
 ifneq ($(filter mips mips64, $(ARCH)),)
 ifeq ($(USE_ASM), Yes)
 ASM_ARCH = mips
 ASMFLAGS += -I$(SRC_PATH)codec/common/mips/
-LOONGSON3A = $(shell g++ -dM -E - < /dev/null | grep '_MIPS_TUNE ' | cut -f 3 -d " ")
-ifeq ($(LOONGSON3A), "loongson3a")
+#only enable mmi for 64-bit system
+ifneq ($(filter mips64, $(ARCH)),)
 CFLAGS += -DHAVE_MMI
+#tell the assembler to accept mmi instructions
+CFLAGS += -Wa,-mloongson-mmi,-mloongson-ext
 endif
 endif
 endif
--- a/codec/common/src/cpu.cpp
+++ b/codec/common/src/cpu.cpp
@@ -310,7 +310,26 @@ uint32_t WelsCPUFeatureDetect (int32_t* pNumberOfLogicProcessors) {
 #elif defined(mips)
 /* for loongson */
 uint32_t WelsCPUFeatureDetect (int32_t* pNumberOfLogicProcessors) {
-#if defined(HAVE_MMI)
+#if defined(__linux__)
+  int flags = 0;
+  FILE* f = fopen ("/proc/cpuinfo", "r");
+
+  if (!f) {
+    return flags;
+  }
+
+  char buf[200];
+  while (fgets (buf, sizeof (buf), f)) {
+    if (!strncmp (buf, "ASEs implemented", strlen ("ASEs implemented"))) {
+      // mmi code depends on ext also
+      if (strstr (buf, " loongson-mmi ") || strstr (buf, " loongson-ext "))
+        flags |= WELS_CPU_MMI;
+      break;
+    }
+  }
+  fclose (f);
+  return flags;
+#elif defined(HAVE_MMI)
   return WELS_CPU_MMI;
 #else
   return 0;


FlyGoat:
@gxw-loongson Please, do not detect features by model string. Models will change. Only ASE flags are reliable.

@GuangweiWang Could you please review this change?

Thanks.

gxw-loongson:
Thanks for you advice and help.
If you enable mmi for all mips64, if the compiler not support the mmi, it will caused some problems.
The current method is not good(use g++ _MIPS_TUNE), we will fix it in the feature.

FlyGoat:
> Thanks for you advice and help.
> If you enable mmi for all mips64, if the compiler not support the mmi, it will caused some problems.
> The current method is not good(use g++ _MIPS_TUNE), we will fix it in the feature.

According to the binutils documents, all the MIPS assembler can support MMI code, as only assembler is involved in the generation of MMI related instructions, we can just ignore compiler and pass -mloongson-mmi,-mloongson-ext to the assembler.

Verified on GCC9 + binutils 2.32 toolchain configured as mips64r2 target with a Loongson-3A3000 machine.

Ref:
https://sourceware.org/binutils/docs-2.32/as/MIPS-ASE-Instruction-Generation-Overrides.html#MIPS-ASE-Instruction-Generation-Overrides
https://sourceware.org/binutils/docs-2.32/as/MIPS-Options.html#MIPS-Options

yinshiyou:
I don't think it is a good choice to enable MMI when ARCH = mips64. ARCH is got with command 'uname -m' in Makefile, there are no direct relationship between 'uname -m' and 'build tool support MMI or not'.
The most robust way to check build tool support MMI or not is to add a compile test in the make process before set "HAVE_MMI". Anthor way is to check whther '-march=loongson3a' or '-mloongson-mmi' is supported.

FlyGoat:
> I don't think it is a good choice to enable MMI when ARCH = mips64. ARCH is got with command 'uname -m' in Makefile, there are no direct relationship between 'uname -m' and 'build tool support MMI or not'.
> The most robust way to check build tool support MMI or not is to add a compile test in the make process before set "HAVE_MMI". Anthor way is to check whther '-march=loongson3a' or '-mloongson-mmi' is supported.

All the MIPS64 assembler after binutils-2.31 should support  -Wa,-mloongson-mmi,-mloongson-ext at all.
I'd agree that we can have a test for that opinion. But current way is blocking MMI being compiled on generic MIPS64 builds.
Other distros like Debian don't set "loongson3a" as their march or mtune. But they can also benefit from MMI optimization.

yinshiyou:
To add MMI check here is in order to ensure program can be succesfully builded with MMI enabled, nomatter what hardware it is(To ensure MMI can running succesfully on the hardware, runtime check will be needed). The previous way is a conservative approach. This patch will cause build failure when compile don't support '-Wa,-mloongson-mmi,-mloongson-ext' although ARCH = mips64.

FlyGoat:
> To add MMI check here is in order to ensure program can be succesfully builded with MMI enabled, nomatter what hardware it is(To ensure MMI can running succesfully on the hardware, runtime check will be needed). The previous way is a conservative approach. This patch will cause build failure when compile don't support '-Wa,-mloongson-mmi,-mloongson-ext' although ARCH = mips64.

I'd agree with you. But current way is really tricky. We'd better find a better method.

yinshiyou:
> I'd agree with you. But current way is really tricky. We'd better find a better method.
The most robust way to check build tool support MMI or not is to add a compile test in the make process before set "HAVE_MMI".

FlyGoat:
@yinshiyou Not only self-defined. Mainline kernel can already display them in cpuinfo.

I know that 3A4000+ have cpucfg instruction to tell that, but that's not a generic way.

See: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v4.19.80&id=fb93ccde081e39631fd16c03d1c9eb0b4bef7edd

FlyGoat:
@yinshiyou 
FYI:

```
[flygoat@flygoat-ls3a7a ~]$ cat /proc/cpuinfo
system type		: generic-loongson-machine
machine			: loongson,generic
processor		: 0
cpu model		: Loongson-3 V0.13  FPU V0.1
model name		: Loongson-3A R3 (Loongson-3A3000) @ 1450MHz
CPU MHz			: 1450.00
BogoMIPS		: 2887.52
wait instruction	: yes
microsecond timers	: yes
tlb_entries		: 1088
extra interrupt vector	: no
hardware watchpoint	: yes, count: 0, address/irw mask: []
isa			: mips1 mips2 mips3 mips4 mips5 mips32r1 mips32r2 mips64r1 mips64r2
ASEs implemented	: dsp dsp2 vz loongson-mmi loongson-cam loongson-ext loongson-ext2
shadow register sets	: 1
kscratch registers	: 6
package			: 0
core			: 0
VCED exceptions		: not available
VCEI exceptions		: not available

```

yinshiyou:
> @yinshiyou Not only self-defined. Mainline kernel can already display them in cpuinfo.
> 
> I know that 3A4000+ have cpucfg instruction to tell that, but that's not a generic way.
> 
> See: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v4.19.80&id=fb93ccde081e39631fd16c03d1c9eb0b4bef7edd

Thank you for your info. One thing to correct is using user mode instruction to check cpu features is the best and generic way, it's more efficient than probe cpuinfo. 
About this patch, I just suggest to analysis 'model name' instead because there are more versions of kernel can suppose it(version <4.19.80).

FlyGoat:
> > @yinshiyou Not only self-defined. Mainline kernel can already display them in cpuinfo.
> > I know that 3A4000+ have cpucfg instruction to tell that, but that's not a generic way.
> > See: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v4.19.80&id=fb93ccde081e39631fd16c03d1c9eb0b4bef7edd
> 
> Thank you for your info. One thing to correct is using user mode instruction to check cpu features is the best and generic way, it's more efficient than probe cpuinfo.
> About this patch, I just suggest to analysis 'model name' instead because there are more versions of kernel can suppose it(version <4.19.80).

Not really. This kernel patch will be backport to all stable releases.
And. The generic means more general to the MIPS world. Not Loongson only. cpucfg is purely self-defined by Loongson.
You can use model string here temporarily. But that's not a permanent solution. Models will change.

gxw-loongson:
The most accurate and simple method is to read ASEs implemented. But it will failed on most of 
existing loongson machines(low version of kernel). For better compatibility, i suggest use mode name rather than ASEs implemented for runtime detect.

yinshiyou:
> Not really. This kernel patch will be backport to all stable releases.
>And. The generic means more general to the MIPS world. Not Loongson only. cpucfg is purely self-defined by Loongson.
>You can use model string here temporarily. But that's not a permanent solution. Models will change.

1. x86 processor implemented 'cpuid', for your reference, you can check the usage in FFmpeg. I just suggest you to use user mode instruction to get the cpu features, not just use cpucfg on loongson platform, moreover cpucfg is supported from 3A4000. We still have to analysis cpuinfo on the processor which don't support this instruction.
2. MMI is loongson self-defined extention, according to the suggestion from loongson kernel team, I will suggest to use 'model name', at least, it works well on all the current versions.

FlyGoat:
I'd agree with you at that way. But don't forget that generic MIPS64 binary will also run on the Loongson systems. Loongnix is not the only distro.

So let's do like this:
1. All the cpuinfo will display msa, so we should probe msa fully with ASEs domain.
2.  For MMI, we probe them by both cpu model and ASEs, if any of them displayed loongson, then set MMI flag.
3. cpucfg have not accpted by generic MIPS toolchain, we should avoid them in upstream code. Probably we could ask Xu Chenghua about his toolchain upstream plan? Other MIPS64 machine will report Illegal instructions if you try to use cpucfg. So for MIPS world, it's poisonous.

We don't mind what are you doing with Loongson's own release, but MIPS and Loongson should benefit eachother in this way.

PR #3185, opened 2019-10-22
Adjust the mmi/msa detection mode for mips platform.

gxw-loongson:
Using mips-simd-check.sh to test the current compiler support mmi/msa
or not before make. If supported, enable mmi/msa.
According to the model name in /proc/cpuinfo to test the current
cpu support mmi/msa or not for runtime detection.
Now We can use the following make instructions on mips platform:
1. make (automatic detection mmi/msa)
2. make ENABLE_MMI=No (disable mmi)
3. make ENABLE_MSA=No (disable msa)
4. make ENABLE_MMI=No ENABLE_MSA=No (disable mmi and msa)

Change-Id: Ibd348ebc11912d7fca1b548c76838675d69b7c40
Signed-off-by: gxw <guxiwei-hf@loongson.cn>

### DIFF
--- a/codec/decoder/core/src/wels_decoder_thread.cpp
+++ b/codec/decoder/core/src/wels_decoder_thread.cpp
@@ -57,12 +57,6 @@
 #define HW_NCPU_NAME "hw.ncpu"
 #endif
 #endif
-#ifdef ANDROID_NDK
-#include <cpu-features.h>
-#endif
-#ifdef __ANDROID__
-#include <android/api-level.h>
-#endif
 
 #include "wels_decoder_thread.h"
 #include <stdio.h>


gxw-loongson:
@FlyGoat @yinshiyou
please help me review the patch, thx.

build/arch.mk
```diff
@@ -30,14 +30,24 @@ CFLAGS += -DHAVE_NEON_AARCH64
...
 ASM_ARCH = mips
 ASMFLAGS += -I$(SRC_PATH)codec/common/mips/
-LOONGSON3A = $(shell g++ -dM -E - < /dev/null | grep '_MIPS_TUNE ' | cut -f 3 -d " ")
-ifeq ($(LOONGSON3A), "loongson3a")
-CFLAGS += -DHAVE_MMI
+#mmi
+ifeq ($(ENABLE_MMI), Yes)
+ENABLE_MMI = $(shell $(SRC_PATH)build/mips-simd-check.sh $(CC) mmi)
+ifeq ($(ENABLE_MMI), Yes)
+CFLAGS += -march=loongson3a -DHAVE_MMI
```
FlyGoat: -Wa,-mloongson-mmi,-mloongson-ext will be better as it won't generate Loongson instructions in C part of generic MIPS Build.

codec/common/src/cpu.cpp
```diff
@@ -309,12 +309,37 @@ uint32_t WelsCPUFeatureDetect (int32_t* pNumberOfLogicProcessors) {
...
+    FILE* fp = fopen("/proc/cpuinfo", "r");
+    if (!fp)
+        return flags;
+
+    char buf[200];
+    memset(buf, 0, sizeof(buf));
+    while (fgets(buf, sizeof(buf), fp)) {
+        if (!strncmp(buf, "model name", strlen("model name"))) {
+            if (strstr(buf, "3A4000")) {
+                flags |= WELS_CPU_MSA | WELS_CPU_MMI;
```
FlyGoat: We should probe MSA via ASEs domain.  Early Loongson-2K Kernel didn't enable MSA. If we purely rely on model, than it may lead to illegal instructions.

For all the versions of MSA enabled kernel, msa will be displayed in the ASEs Implemented.

codec/common/src/cpu.cpp
```diff
@@ -309,12 +309,37 @@ uint32_t WelsCPUFeatureDetect (int32_t* pNumberOfLogicProcessors) {
...
+            } else if (strstr(buf, "2K1000")) {
+                flags |= WELS_CPU_MSA | WELS_CPU_MMI;
+            } else if (strstr(buf, "Loongson-3A") || strstr(buf, "Loongson-3B")) {
+                flags |= WELS_CPU_MMI;
+            }
+            break;
+        }
+    }
```
FlyGoat: We'd better also check MMI in ASEs implemented here.
Kernels from Lemote didn't display model as Loongson-3A/Loongson-3B but they'll provide loongson-mmi and loongson-ext in ASEs Implemented.

build/arch.mk
gxw-loongson: Lower versions of GCC(like 4.9.3 on loongson) are not support -mloongson-mmi,-mloongson-ext. Maybe we can only add -march=loongson3a on *_mmi.c.

FlyGoat: As far as I know, Loongson had backport this option to GCC 4.9 in Loongnix. And it also mainlined in GCC9.

So can we just drop support for old compilers in upstream code?

codec/common/src/cpu.cpp
gxw-loongson: Thank you for your advice, I will check model name and ASEs for mmi, check ASEs for msa.

yinshiyou:
looks good now.

GuangweiWang:
can you help to merge you commit to one? thanks

build/mips-simd-check.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
GuangweiWang: is this scrip just for check mmi/msa which is the architecture of loongson.

gxw-loongson: yes

Issue #3186, opened 2019-10-22
fuzz issue 18458


sijchen:
https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=18458

INFO: Loaded 1 PC tables (7332 PCs): 7332 [0x8379c10,0x8388130),
--
  | /mnt/scratch0/clusterfuzz/bot/builds/clusterfuzz-builds-i386_openh264_bc9badabd72f3ca917863fe37b13183bc37b8f1a/revisions/decoder_fuzzer: Running 1 inputs 100 time(s) each.
  | Running: /crash-9683c3493f1651af5dd591bd124e267b271b336e
  | =================================================================
  | ==1==ERROR: AddressSanitizer: heap-use-after-free on address 0xf5b01a17 at pc 0x081f0d87 bp 0xfff4f2e8 sp 0xfff4f2e0
  | WRITE of size 1 at 0xf5b01a17 thread T0
  | SCARINESS: 41 (1-byte-write-heap-use-after-free)
  | #0 0x81f0d86 in WelsDec::SetUnRef(WelsDec::SPicture*) openh264/codec/decoder/core/src/manage_dec_ref.cpp:89:50
  | #1 0x81f72e5 in WelsDec::SlidingWindow(WelsDec::TagWelsDecoderContext*, WelsDec::TagRefPic*) openh264/codec/decoder/core/src/manage_dec_ref.cpp:762:9
  | #2 0x81f5f42 in WelsDec::WelsMarkAsRef(WelsDec::TagWelsDecoderContext*, WelsDec::SPicture*) openh264/codec/decoder/core/src/manage_dec_ref.cpp:624:14
  | #3 0x81e0bbf in WelsDec::MarkECFrameAsRef(WelsDec::TagWelsDecoderContext*) openh264/codec/decoder/core/src/error_concealment.cpp:434:18
  | #4 0x81dad76 in WelsDec::CheckAndFinishLastPic(WelsDec::TagWelsDecoderContext*, unsigned char**, TagBufferInfo*) openh264/codec/decoder/core/src/decoder_core.cpp:2803:13
  | #5 0x81b9b7a in WelsDecodeBs openh264/codec/decoder/core/src/decoder.cpp:834:7
  | #6 0x81aaf87 in WelsDec::CWelsDecoder::DecodeFrame2(unsigned char const*, int, unsigned char**, TagBufferInfo*) openh264/codec/decoder/plus/src/welsDecoderExt.cpp:583:3
[...]

PR #3187, opened 2019-10-22
commit-3 for threaded decoding support

xiaotianshi2:
Mainly update DecodeCurrentAccessUnit and associated functions to do threaded parallel frame decoding.
Add additional function PrefetchLastPicForThread and variables for better control of threaded decoding in the case of multi-slice with one slice group contains one slice.


xiaotianshi2:
Temporarily close this PR since it will need to fix issue #3186 first then re-create commit-3 of multi-thread support.

PR #3192, opened 2019-11-01
Fix building for android

mstorsjo:
Remove unused includes from wels_decoder_thread.cpp.

The include path to cpu-features.h is only added to COMMON_INCLUDES,
as only files in codec/common/src included it - before the newly added
wels_decoder_thread.cpp, which didn't really need it.


huili2:
@xiaotianshi2 could you please answer this? Thanks.

xiaotianshi2:
@mstorsjo Thanks for the fix.

PR #3198, opened 2019-11-05
fix fuzz issues 18744, 18746 and 18747 by clearing info of buffered v…

xiaotianshi2:
…ideo frames when decreasing DPB buffers.


xiaotianshi2:
@sijchen @huili2 Could you restart number 4 build machine? Thanks.

PR #3215, opened 2019-12-03
Thread decoding support: Update pNzc (none-zero count) for precise deblocking

xiaotianshi2:
non-zero count vector should also be set for the current MB' neighborMBs. This is achieved by adding pNzc allocation in PICTURE.


xiaotianshi2:
closed due to contains failed build.

PR #3219, opened 2019-12-10
Thread unit repeated test with logging to catch segmentation fault error.

xiaotianshi2:
This is just for test purpose.


xiaotianshi2:
This was for Travis-CI build thread decoding test and is no longer needed.

PR #3223, opened 2019-12-11
Issue 3221 to fix oss fuzz 18747 and 18864

xiaotianshi2:


xiaotianshi2:
closed due to unexpected thread decoding test case segmentation fault error.

PR #3230, opened 2020-01-28
meson: fix libdir in openh264.pc.in

CapOM:
Parse --libdir option passed to meson and reflect
that into the openh264.pc file.

Useful when libopenh264.so is installed in the
directory /usr/lib/x86_64-linux-gnu/. Otherwise
programs that link with openh264 might fail with
error: "/usr/bin/ld: cannot find -lopenh264"


CapOM:
Hi reviewers, please take a look, thx!

mstorsjo:
This broke the pkg-config files when generated with make instead of meson. See #3233 for a fix for that.

PR #3232, opened 2020-02-13
fix osx thread testcase.

xiaotianshi2:


huili2:
Could you please give some details why there are two results here?

xiaotiansf:
Actually, as many as my local tests, it only gives one result for this specific test case. My guess is that somehow, one of the frames at the output is switched in a very rare condition. As I mentioned that it will need to implement frame output callback mechanism to insure frame output order. But that will take time to implement. In any cases, the decoded yuv frames themselves look good.

xiaotiansf:
Run a batch script with at least 100 times of "run_Test.sh" with unique returned hash string for the test case on OSX.
Nevertheless, even there is more than returned has hash strings, it does not mean it is a decoding error because even just one-bit stream decoding mistake will most likely causes subsequent decoding error loggings, which is not the case.  So I think it is a minor issue (with more than hash strings) which has two possibilities (1. deblocking (smooth filtering) somehow results in very little different pixel value on some blocks, which is not visual noticeable. 2. somehow there are chanced a little chance that one or two pictures are not in correct displaying order at output due to there is not picture ready callback API).    

I think it is important to add this optional (since it optional so it is quite safe) threaded-decoding to new release so that it can get more results from customers to help improve and make it more robust.

PR #3233, opened 2020-02-17
Fix libdir path in pkg-config files when built with make

mstorsjo:
Since d6a361410540a4aa037bd4828728a3d08053dd07, the pkg-config
input file expects to have @libdir@ replace in addition to @prefix@.


GuangweiWang:
thanks!

PR #3246, opened 2020-03-05
Using and operation when detecting Loongson mmi instructions

gxw-loongson:
The Loongson 64-bit simd optimizations has used LEXT1(LoongISA general EXTensions 1),
so we must make sure loongson-mmi and loongson-ext exist at the same time.


gxw-loongson:
@FlyGoat @yinshiyou
please help me review the patch, thx.

FlyGoat:
Code LGTM.
Commit message makes me a little bit confusing but should be fine.
Thanks!

PR #3247, opened 2020-03-10
meson: add support for android, ios, macos

ystreet:
Adds support for arm and arm64 in the meson build.

Tested for android, ios, macos by https://gitlab.freedesktop.org/ystreet/cerbero/pipelines/117966

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -36,8 +36,6 @@ encoder_inc = include_directories([
   join_paths('codec', 'encoder', 'plus', 'inc'),
 ])
 
-asm_inc = join_paths(meson.current_source_dir(), 'codec', 'common', 'x86', '')
-
 nasm = find_program('nasm', 'nasm.exe')
 
 system = host_machine.system()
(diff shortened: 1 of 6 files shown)


nirbheek:
Could we please get these patches merged? I have more patches on top of this that add support for Windows ARM/ARM64 :)

meson.build
```diff
@@ -127,18 +156,21 @@ foreach t : ['', '-static']
...
+  pkglibs = cpp_lib
+  if libm_dep.found()
+    pkglibs += ' -lm'
+  endif
+  if pthread_dep.found()
+    pkglibs += ' -lpthread'
+  endif
   if t == '-static'
-    do_install = false
-    pkgconf.set('LIBS', '-lstdc++ -lpthread -lm')
```
GuangweiWang, line 132: why stdc++ is not be specified？

ystreet: iOS and macOS do not have stdc++ anymore and was removed from the toolchain a couple of years ago.  They use libc++.  Android has also deprecated their stdc++ module in favor of libc++ and would be removed in a future NDK release.

ystreet: Also, stc++ is specified by default through the cpp_lib variable ans is only overridden for android, macos and ios

PR #3256, opened 2020-03-26
meson: disable tests automatically if gtest is not found

ystreet:
There's not much point in the tri-state feature option if we don't
correctly use the 'auto' state to disable tests when dependencies are
not found and subproject downloads have been disabled.

### DIFF
--- a/test/meson.build
+++ b/test/meson.build
@@ -1,13 +1,15 @@
-if not get_option('tests').disabled()
-  gtest_main_dep = dependency('gtest', main : true, fallback: ['gtest', 'gtest_main_dep'], required: true)
-  gtest_dep = dependency('gtest', fallback: ['gtest', 'gtest_dep'], required: true)
+build_tests = get_option('tests')
+
+if not build_tests.disabled()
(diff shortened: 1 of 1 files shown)


tp-m:
Would be nice to get this merged.

tp-m:
This looks correct to me and would be nice to get merged.

Quick explanation in case it's unclear why:

Basically the `tests` option can have three values: `auto`, `disabled`, `enabled`.

If the option is `disabled`, then `dependency('gtest', required: get_option('tests'))` will return a `not-found` dependency object (even if gtest is in fact available), so then the tests will be skipped as they should.

If the option is `enabled`, then `dependency('gtest', required: get_option('tests'))` will error out if gtest can't be found.

If the option is `auto`, then `dependency('gtest', required: get_option('tests'))` will return either a 'found' gtest dependency object or a `not-found` object. In this case tests will be built and run if gtest is available, and will be skipped if it's not available. This is a pretty good default for consumers of the library such as GStreamer, and is how it's done pretty much everywhere else as well.

PR #3266, opened 2020-04-10
Pr multi refs encode

fulinjie:
Found some potential issues while enabling multi refs encoding in FFmpeg.
Please help to comment.

### DIFF
--- a/codec/encoder/core/inc/param_svc.h
+++ b/codec/encoder/core/inc/param_svc.h
@@ -365,7 +365,7 @@ typedef struct TagWelsSvcCodingParam: SEncParamExt {
       uiIntraPeriod = ((uiIntraPeriod + uiGopSize - 1) / uiGopSize) * uiGopSize;
 
     if (((pCodingParam.iNumRefFrame != AUTO_REF_PIC_COUNT)
-         && ((pCodingParam.iNumRefFrame > MAX_REF_PIC_COUNT) || (pCodingParam.iNumRefFrame < MIN_REF_PIC_COUNT)))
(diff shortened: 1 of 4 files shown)


fulinjie:
Hi @ruil2 , would you please help to review and offer some comments?

huili2:
the first commit is OK from my side. I'm wondering the "loosing" reason? May it affect the original reference picture settings in hierarchical structures?

fulinjie:
The main reason for loosing this check is to allow automatically adjusting the settings of reference number inside libopenh264 with a warning prompted for user, instead of failing directly.

IMHO this would be more robust/compatible for users if they didn't know the constraints of reference number and keep the encoder working.

For example, if setting -refs 3 in FFmpeg (require reference to 3),
libopenh264 would report warnings like:
iNumRefFrame(3) setting does not support the temporal and LTR setting, will be reset to 4.

After reset the reference, encoder could work well.

Ps, I'm not strongly insisting on this, if it breaks something I didn't noticed.
Feel free to only apply the first commit which is functional, thx.

fulinjie:
ping.

fulinjie:
Hi @huili2, any updates?

huili2:
sorry late for this. I'm OK for this PR.

fulinjie:
Thanks for the review.

mstorsjo:
This change broke 18 tests in EncodeFile/EncoderOutputTest - please update the test references accordingly.

mstorsjo:
Just FWIW, I sent a fix for the test references in #3272.

PR #3272, opened 2020-04-20
Fix test references after pull request #3266

mstorsjo:


huili2:
Thanks for the information. Seems the travis-ci does not start automatically so that previous PR was merged without test. We are investigating on this.

mstorsjo:
Could you apply this while investigating the CI issue, to fix the breakage?

huili2:
done.

PR #3273, opened 2020-04-23
Fix another test reference

mstorsjo:
The DecodeParseAPI.ParseOnly_General test doesn't test all layers
in each test round, but only one, so all of these references weren't
fixed properly at once.


mstorsjo:
Superseded by #3275, closing this one.

PR #3306, opened 2020-07-06
Fix read frame for thread decoding

xiaotianshi2:
update h264dec.cpp's readPicture to make it to read a complete frame at a time for thread-decoding of multi-slice frame.

### DIFF
--- a/codec/console/dec/src/h264dec.cpp
+++ b/codec/console/dec/src/h264dec.cpp
@@ -68,6 +68,40 @@ int    g_iDecodedFrameNum = 0;
 #endif
 //using namespace WelsDec;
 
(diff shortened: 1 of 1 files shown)


huili2:
it is rather weird to implement entropy decoding in console code....

xiaotianshi2:
@huili2 Right, it is.  However, this is just for h264dec test purpose. In real-world (for example ffmpeg AVCodec), the input buffer is always  a frame and it does not need to call this function.

huili2:
thanks. could you please make a clean commit, to remove "merge" one? Thanks.

xiaotianshi2:
Sure and will do so (I was wondering how that "merge" one got in).

xiaotianshi2:
not a clean PR and so close it.

PR #3312, opened 2020-07-15
update readPicture to make it to read one complete frame at a time for the support of thread-decoding of multi-slice frame

xiaotianshi2:
.


huili2:
So how about the PR https://github.com/cisco/openh264/pull/3306? Seems duplicated?

PR #3333, opened 2020-09-10
updated thread-decoding unit tests upon thread-decoder update.

xiaotianshi2:


xiaotianshi2:
Still there are three test cases which have more than one hash values. So this attempt is closed for now.

PR #3334, opened 2020-09-13
update readFrame for thread-decoding test and unit test cases.

xiaotianshi2:
a few test cases have to be excluded since they return multiple hash values on travis-ci build server.

### DIFF
--- a/test/BaseThreadDecoderTest.h
+++ b/test/BaseThreadDecoderTest.h
@@ -40,7 +40,6 @@ class BaseThreadDecoderTest {
   bool ThreadDecodeFile (const char* fileName, Callback* cbk);
 
   bool Open (const char* fileName);
-  bool DecodeNextFrame (Callback* cbk);
   ISVCDecoder* decoder_;
 
  private:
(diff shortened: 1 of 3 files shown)


huili2:
Do you know how "multiple hash values" are introduced? Does it occur in original code or caused by B frames?

xiaotianshi2:
@huili2 Actually I don't know why the results from travis-ci build server can have multiple hash values with these cases. It is quite strange that the returned hash value is unique  when I run hundred times on my local machines (linux, windows and OSX) for these cases.

xiaotianshi2:
@huili2 It does not seem something to do with B-frame. These cases with multiple hash values contain no b-frame.

thaytan:
Returning different hashes on a different machine suggests a few options:
 * Race conditions in multi-thread processing leading to different results - seems unlikely since you've tested on multiple machines
 * Uninitialised memory in use, and the different setup leads to different allocation/use patterns - perhaps when run inside the Travis test harness only?
 * Travis is compiling with different optimisation / SIMD acceleration

xiaotianshi2:
@thaytan Thanks for the comment. Note that there is one test case "res/MR2_MW_A.264" in which, it shows multiple hash values in both local machine and Travis-CI. But I compared dumped motion vectors, there are the same, which is what I expected (because even if just one single bit stream decoding error, it can cause subsequent bit-stream decoding errors and mess up motion vectors). Also I did not notice any kind of artifacts in output yuv in this test case. This may hints that threading somehow affects the stage in pixel reconstruction.  Just don't know how threading affects pixel reconstruction and why it affects just some test cases.

PR #3338, opened 2020-09-17
Fix building for arm/aarch64 with mingw (using clang)

mstorsjo:
@nirbheek If you have feedback on the meson changes, I'd appreciate it. Instead of hardcoding the condition where asm_gen is needed, it'd be neater if there were a construct like `if defined(asm_gen)`. I tried setting `asm_gen = false` or similar things, for cases where it's not needed, but any comparison operator, where `asm_gen` can either be a generator or another type, seems to produce warnings/errors in some case.

### DIFF
--- a/build/platform-mingw_nt.mk
+++ b/build/platform-mingw_nt.mk
@@ -1,4 +1,4 @@
-include $(SRC_PATH)build/x86-common.mk
+include $(SRC_PATH)build/arch.mk
 SHAREDLIB_DIR = $(PREFIX)/bin
 SHAREDLIBSUFFIX = dll
 SHAREDLIBSUFFIXFULLVER=$(SHAREDLIBSUFFIX)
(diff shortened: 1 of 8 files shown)


nirbheek:
You could do `asm_gen = disabler()` and then use `is_disabler()` to check if it's a disabler and skip code based on that. Another way might be to use a separate variable `use_asm_gen = false`. Lastly, you could mirror the arch / argument_syntax conditionals inside the `codec` subdirs so you don't need to check the contents of `asm_gen` at all.

mstorsjo:
Thanks! I updated it and went with a separate variable, `use_asm_gen`, which should be the simplest and clearest approach, even if it isn't maybe the most elegant.

huili2:
strange segmentation fault occurs. Should be the original code issue. Let me post a issue on this.

PR #3342, opened 2020-09-23
arm: Align constants that are used with the vldr instruction

mstorsjo:
The vldr instruction requires the data to be properly aligned.

### DIFF
--- a/codec/decoder/core/arm/intra_pred_neon.S
+++ b/codec/decoder/core/arm/intra_pred_neon.S
@@ -126,6 +126,7 @@ WELS_ASM_FUNC_END
 
 
 
+.align 3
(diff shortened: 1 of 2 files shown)


codec/decoder/core/arm/intra_pred_neon.S
```diff
@@ -126,6 +126,7 @@ WELS_ASM_FUNC_END
 
 
 
+.align 3
```
GuangweiWang, line 129: why the number is 3?

mstorsjo: Because `.align` on arm denotes the alignment as a power of two. This gives 8 byte alignment, which is needed as these are loaded 8 bytes all at once with vldr.

PR #3343, opened 2020-09-25
Use the right thread function signature in WealThreadPoolTest

mstorsjo:
This fixes warnings when built with clang for i686, warning about
casting between cdecl and stdcall - which normally is fatal. As
long as the function actually took no parameters (even though
there technically is an unused one), there was no difference in
stack adjustment between stdcall and cdecl though.

This avoids needing to use casts when passing the function to
WelsThreadCreate across all platforms.


mstorsjo:
Ping

PR #3348, opened 2020-10-11
fix ossfuzz issue: decoder hits infinite loop when iMaxLongTermFrameIdx is invalid (exceeds the maximum).

xiaotiansf:
…


codec/decoder/core/src/decoder_core.cpp
```diff
@@ -543,7 +543,12 @@ int32_t ParseDecRefPicMarking (PWelsDecoderContext pCtx, PBitStringAux pBs, PSli
           WELS_VERIFY_RETURN_IF (-1, bMmco4Exist);
           bMmco4Exist = true;
           WELS_READ_VERIFY (BsGetUe (pBs, &uiCode)); //max_long_term_frame_idx_plus1
-          kpRefMarking->sMmcoRef[iIdx].iMaxLongTermFrameIdx = -1 + uiCode;
```
Cu2y, line 546: `https://github.com/cisco/openh264/pull/3348#issue-501083556`

PR #3351, opened 2020-10-19
meson: Use pkg-config generator and some cleanups

xclaesse:
When using library() instead of shared_library() and static_library,
meson will build shared, static, or both depending on the
value of static_library option.

As far as I know extract_all_objects() was uses as workaround for Meson
bugs fixed a while ago when using not installed static libraries.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -54,7 +54,7 @@ asm_inc = []
 casm_inc = []
 cpp_lib = '-lstdc++'
 
-libm_dep = cpp.find_library('libm', required : false)
+libm_dep = cpp.find_library('m', required : false)
 deps += [libm_dep]
 
 if ['linux', 'android', 'ios', 'darwin'].contains(system)
--- a/meson.build
+++ b/meson.build
@@ -54,9 +54,6 @@ asm_inc = []
 casm_inc = []
 cpp_lib = '-lstdc++'
 
-# TODO: should rely on dependency('threads') instead and change the pkg-config
-# generator below
-pthread_dep = cpp.find_library('pthread', required : false)
 libm_dep = cpp.find_library('libm', required : false)
 deps += [libm_dep]
 
--- a/meson.build
+++ b/meson.build
@@ -184,7 +184,7 @@ api_header_deps = []
 subdir ('codec')
 subdir ('test')
 
-libopenh264_shared = library('openh264',
+libopenh264 = library('openh264',
   link_whole: [libcommon, libprocessing, libencoder, libdecoder],
   install: true,
   soversion: major_version,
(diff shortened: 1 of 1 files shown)


xclaesse:
@tp-m FYI, this is needed to get openh264 static linked into gstreamer-full.

nanonyme:
@xclaesse are you sure that is sensible in general? Only Cisco is allowed to distribute OpenH264 binaries with the patent license so static libraries are problematic for most parties.

xclaesse:
@nanonyme on the contrary, by default openh264 currently installs both shared a static libraries. With this PR by default it will only install shared library because meson defaults to default_library=shared. You have to build with --default-library=both or --default-library=static to get the static library with my change.

xclaesse:
That being said, I didn't check the license, I'm not distributing openh264 I just saw this inconsistency in the build system with my Meson developer hat ;-)

nanonyme:
Right. I was largely referring to that ffmpeg-full use case you mentioned. Expected way is someone builds openh264, Cisco puts it available and end user downloads it directly from Cisco where it's plugged to ffmpeg or whatever. So from that point of view even allowing static linkage seems strange. Maybe there simply should be no static version? Does anyone need it? It seems like a legal footgun to me.

thaytan:
If someone is building OpenH264 for themselves they should already know the H.264 licensing situation. Dynamic or static library build doesn't make a difference - the code license permits either.

If you want to take advantage of Cisco's H.264 licensing you need to get the binary from them. For iOS a static library is appropriate and provided:

https://github.com/cisco/openh264/releases

nanonyme:
@thaytan are you sure iOS usage is appropriate? It does not sound possible to fulfill the requirement there that customer gets OpenH264 through Cisco (not developer but customer).  The patent grant is not generally considered to allow redistribution.

tp-m:
@nanonyme I don't think this is an accurate depiction of the legal situation at all to be honest, it only reflects a particular distro perspective/pov really. But even if it were there are many people that do not distribute the software they build to third parties but simply use it locally. Lots of other people can distribute their own openh264 builds entirely legally too, it just happens that Cisco makes builds available for distros to use. Anyway, I think this consideration is entirely irrelevant to this MR anyway to be honest :)

The MR is a clean-up that embodies meson best practices, and if it works correctly it should be merged.

eli-schwartz:
Fixed by https://github.com/mesonbuild/meson/pull/5936

This requires bumping the minimum version of meson to 0.52 to guarantee that no one tries building it using versions of meson from before the fix.

Other than that... this results in a more idiomatic meson.build with a dozen fewer lines, so it seems like a no-brainer to merge :) the licensing situation seems quite tangential.

eli-schwartz:
@xclaesse Could you bump the minimum meson version?

xclaesse:
> @xclaesse Could you bump the minimum meson version?

done.

@nanonyme I think we answered your concerns, do you think we can go ahead with this PR?

xclaesse:
@nanonyme kind reminder to review this PR please :)

nanonyme:
Sure, do go ahead. I wasn't aware I was a blocker.

xclaesse:
I don't have merge permission, who does?

GuangweiWang:
@xclaesse 
CI build failed. can you help to check?

eli-schwartz:
@GuangweiWang The CI failure is because commit d5dbb59a7f134e0c1df79d4e784a7a3bf00196c5 (and the current state of the github actions workflow) pins an exact version of meson to install in the CI.

This pre-existing PR bumped the *minimum* version of meson from 0.50 to 0.52 and thus meson failed because it detected that the version installed from pip install is too old to handle the project.

I'm not sure I understand the point of pinning an old version here. If you just want to prevent breaking the minimum version, meson warns you if you use too new features (and you can make that into failure by passing --fatal-meson-warnings).

GuangweiWang:
failed reason:
Run meson builddir
The Meson build system
Version: 0.50.1
Source dir: /home/runner/work/openh264/openh264
Build dir: /home/runner/work/openh264/openh264/builddir
Build type: native build

meson.build:1:0: ERROR: Meson version is 0.50.1 but project requires >= 0.52.0.

A full log can be found at /home/runner/work/openh264/openh264/builddir/meson-logs/meson-log.txt
Error: Process completed with exit code 1.

xclaesse:
Rebased this PR and removed the parts that got already merged in separate PRs.

GuangweiWang:
@xclaesse can you help to resolve the conflicts? thanks

xclaesse:
> @xclaesse can you help to resolve the conflicts? thanks

done.

While at it, what's the plan for the next release? Currently the latest release to totally broken when building with Meson unfortunately.

PR #3358, opened 2020-12-09
support CPU detection for avx512 

GuangweiWang:
as the title.


GuangweiWang:
@mstorsjo could you help to review this PR? thanks.

mstorsjo:
> @mstorsjo could you help to review this PR? thanks.

Sorry for the late reply.

I'm not very familiar with the x86 bits at this level, but the code looks acceptable to me.

I'm wondering if it'd make sense to do less in the new function, and just call the existing `WelsCPUId` a couple times, and a separate function like `WelsXGETBV` which could be used both for this and for FMA detection? That'd require less logic in assembly form and keeping most of it high level in C instead.

GuangweiWang:
thank you for your reply. let me try if can use the existing functions.

PR #3371, opened 2021-03-14
fix oss-fuzz reported issue 31439 (integer range out of scope).

xiaotiansf:


xiaotiansf:
closed due to build failure.

PR #3382, opened 2021-05-16
Correct-way to fix picture out-of-order 

xiaotiansf:
when input h.264 stream contains H.264 spec un-complied POC information for any stream without B-frame. Related to issue 3312.


joakim-tjernlund:
hmm, I cannot fix issue 3312, pointer?

xiaotiansf:
@joakim-tjernlund Can you share with me a test case which does not work? I can take a look.

joakim-tjernlund:
> @joakim-tjernlund Can you share with me a test case which does not work? I can take a look.

I don't have any. I was just curious what issue 3312 was. I just found it though :)

joakim-tjernlund:
Maybe @huili2 can comment/merge and then do a new release? The current one is getting old

codec/decoder/plus/src/welsDecoderExt.cpp
```diff
@@ -699,8 +699,15 @@ DECODING_STATE CWelsDecoder::DecodeFrameNoDelay (const unsigned char* kpSrc,
     if (m_sReoderingStatus.iNumOfPicts) {
       WAIT_EVENT (&m_sBufferingEvent, WELS_DEC_THREAD_WAIT_INFINITE);
       RESET_EVENT (&m_sReleaseBufferEvent);
-      ReleaseBufferedReadyPicture (NULL, ppDst, pDstInfo);
-      SET_EVENT (&m_sReleaseBufferEvent);
+      if (!m_sReoderingStatus.bHasBSlice) {
+        if (m_sReoderingStatus.iNumOfPicts > 1) {
```
joakim-tjernlund, line 703: if m_sReoderingStatus.iNumOfPicts <= 1 there is no call to either ReleaseBufferedReadyPictureNoReorder() nor ReleaseBufferedReadyPictureReorder()
Is that intended?

xiaotiansf: Yes. I think it is intended. But I'll double check it and let you know.

xiaotiansf: @joakim-tjernlund You are right. line 703 is better with  if (m_sReoderingStatus.iNumOfPicts >= 1) {. But I guess it is harmless with if (m_sReoderingStatus.iNumOfPicts > 1) {.

joakim-tjernlund: You sure it is harmless? Seem odd to me to skip calling any of ReleaseBufferedReadyPictureNoReorder()/ReleaseBufferedReadyPictureReorder() if m_sReoderingStatus.iNumOfPicts == 1

Issue #3401, opened 2021-07-08
setting sps->direct_8x8_inference_flag doesn't take level into account


vacing:
According to Rec. ITU-T H.264 (10/2016) A3.3.2 A-4, `direct_8x8_inference_flag` should be 1 when level beyond or equal 3, but all set to 0 now.

sijchen:
thank you for the input! Rec. ITU-T H.264 (10/2016) A3.3.2 A-4, is only for Main .... Profile, should we add a profile check in PR#3404 ? thank you

vacing:
more discussion at #3404

PR #3404, opened 2021-07-09
fix bug of sps->direct_8x8_inference_flag setting

vacing:
related to #3401

### DIFF
--- a/codec/encoder/core/src/au_set.cpp
+++ b/codec/encoder/core/src/au_set.cpp
@@ -304,7 +304,11 @@ int32_t WelsWriteSpsSyntax (SWelsSPS* pSps, SBitStringAux* pBitStringAux, int32_
   BsWriteUE (pLocalBitStringAux, pSps->iMbHeight - 1);          // pic_height_in_map_units_minus1
   BsWriteOneBit (pLocalBitStringAux, true/*pSps->bFrameMbsOnlyFlag*/);  // bFrameMbsOnlyFlag
 
-  BsWriteOneBit (pLocalBitStringAux, 0/*pSps->bDirect8x8InferenceFlag*/);       // direct_8x8_inference_flag
+  uint8_t d8x8 = 0;
+  if (pSps->iLevelIdc >= 30)
+    d8x8 = 1;
+  BsWriteOneBit (pLocalBitStringAux, d8x8/*pSps->bDirect8x8InferenceFlag*/);       // direct_8x8_inference_flag
+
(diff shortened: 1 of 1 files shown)


sijchen:
ITU-T H.264 (10/2016) A3.3.2 A-4, is only for Main .... Profile, should we add a profile check in this change ? thank you

vacing:
> ITU-T H.264 (10/2016) A3.3.2 A-4, is only for Main .... Profile, should we add a profile check in this change ? thank you

yes, will fix

vacing:
> change

> direct_8x8_inference_flag is not relevant to the Baseline, Constrained Baseline, Constrained High,
High 10 Intra, High 4:2:2 Intra, High 4:4:4 Intra, and CAVLC 4:4:4 Intra profiles as these profiles do not allow B slice
types, and direct_8x8_inference_flag is equal to 1 for all levels of the Extended profile.

I think we need not to check profile. If they didn't mention, it means they don't care the value; other profiles in annexes(svc mvc mvcd ) obey Table A-4.
BTW, direct_8x8_inference_flag set to 1 for all bitstreams now...

vacing:
> ITU-T H.264 (10/2016) A3.3.2 A-4, is only for Main .... Profile, should we add a profile check in this change ? thank you

> > direct_8x8_inference_flag is not relevant to the Baseline, Constrained Baseline, Constrained High,
> > High 10 Intra, High 4:2:2 Intra, High 4:4:4 Intra, and CAVLC 4:4:4 Intra profiles as these profiles do not allow B slice
> > types, and direct_8x8_inference_flag is equal to 1 for all levels of the Extended profile.
> 
> I think we need not to check profile. If they didn't mention, it means they don't care the value; other profiles in annexes(svc mvc mvcd ) obey Table A-4.
> BTW, direct_8x8_inference_flag set to 1 for all bitstreams now...

sijchen:
thank you for the contribution!

sijchen:
I assume that we will need to update the binary values in UT after this PR merged, however, the Travis hook seems to be broken at this moment, we may update that later after the Travis hook is resumed. cc @huili2

vacing:
> BTW, direct_8x8_inference_flag set to 1 for all bitstreams now...

thanks for merged, here I mean in x264

mstorsjo:
> I assume that we will need to update the binary values in UT after this PR merged, however, the Travis hook seems to be broken at this moment, we may update that later after the Travis hook is resumed. cc @huili2

I made a PR to update those in #3407. Even if Travis isn't active, others may be running the tests externally.

Issue #3399, opened 2021-07-07
not set nal_ref_idc properly of layer 1 when temporal layer equals 4


vacing:
when there has 4 temporal layers: 0, 1, 2, 3
the nal_ref_idc now is :
```
layer 3: 0
layer 2: 1
layer 1: 3
layer 0: 3
```
I think we should set nal_ref_idc of layer 1 to 2 to distinguish layer 0 and layer 1.

sijchen:
as far as I know, according to standard, there is actually no mapping between non-0 values of nal_ref_idc and referencing structure. Such mapping will not be guaranteed in future releases, so we may not want to give illusion saying such mapping can be relied on.

please kindly provide more justification on the change is beneficial.

some other syntax that might be relevant:
D.1.12 Sub-sequence information SEI message syntax
or 
G.7.3.1.1 NAL unit header SVC extension syntax

joakim-tjernlund:
Not doing this just breaks apps that do make this assumption, why not be helpful towards these?

vacing:
> as far as I know, according to standard, there is actually no mapping between non-0 values of nal_ref_idc and referencing structure. Such mapping will not be guaranteed in future releases, so we may not want to give illusion saying such mapping can be relied on.

yes, the h264 standard itself doesn't give an explicit description of how to use the non-zero value of nal_ref_idc.
rfc6184(rfc for RTP, used in RTC, I think that's where openh264 is used) extend it:

"In addition to the specification above, according to this RTP
payload specification, values of NRI indicate the relative
transport priority, as determined by the encoder. MANEs can
use this information to protect more important NAL units better
than they do less important NAL units. The highest transport
priority is 11, followed by 10, and then by 01; finally, 00 is
the lowest.
Informative note: Any non-zero value of NRI is handled
identically in H.264 decoders. Therefore, receivers need
not manipulate the value of NRI when passing NAL units to
the decoder."

> 
> please kindly provide more justification on the change is beneficial.

the main purpose we use nal_ref_idc is to distinguish the importance of different frames, so that we know how to send them to receiver when the Net is not so good. So the more information we can get, the better strategy we can use.
change the nal_ref_idc value doesn't affect the coding process itself, just indicate frame-dependency more clearly, it make no difference to decoder.

> 
> some other syntax that might be relevant:
> D.1.12 Sub-sequence information SEI message syntax
> or
> G.7.3.1.1 NAL unit header SVC extension syntax

[...]

PR #3405, opened 2021-07-11
modify nal_ref_idc from 3 to 2 for level 1 when output 4 temporal layers

vacing:
related to #3399


sijchen:
we can merge the PR for now, but again, this behavior may not be guaranteed in future release if there are other reasonable asks. see discussions in issue#3399

PR #3407, opened 2021-07-13
Fix the encoding test references after 9ed0676e06c1c5b11f6df4a7d50f717fae5cabd9

mstorsjo:


mstorsjo:
Ping! Would you please consider to fix the test references so that `codec_unittest` passes by merging this?

PR #3415, opened 2021-08-30
Makefile, pkg-config: Add INCLUDE_PREFIX variable, use include/openh2…

dpocock:
Makefile, pkg-config:

Add ```INCLUDE_PREFIX``` variable,
use ```$(PREFIX)/include/openh264``` as default,
include correct default in the ```pkg-config``` file

People who use the ```pkg-config``` mechanism will benefit from this immediately and may not have to make any local changes to their own code.

People who hardcoded the paths may need to start using ```pkg-config``` or tweak their paths as they prefer.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ CFLAGS_DEBUG=-g
 BUILDTYPE=Release
 V=Yes
 PREFIX=/usr/local
+INCLUDE_PREFIX=$(PREFIX)/include/openh264
 SHARED=-shared
 OBJ=o
 DESTDIR=
@@ -292,8 +293,8 @@ $(PROJECT_NAME)-static.pc: $(PROJECT_NAME).pc.in
 	@sed -e 's;@prefix@;$(PREFIX);' -e 's;@libdir@;$(PREFIX)/lib;' -e 's;@VERSION@;$(FULL_VERSION);' -e 's;@LIBS@;$(STATIC_LDFLAGS);' -e 's;@LIBS_PRIVATE@;;' < $< > $@
 
 install-headers:
-	mkdir -p $(DESTDIR)$(PREFIX)/include/wels
-	install -m 644 $(SRC_PATH)/codec/api/svc/codec*.h $(DESTDIR)$(PREFIX)/include/wels
+	mkdir -p $(DESTDIR)$(INCLUDE_PREFIX)
+	install -m 644 $(SRC_PATH)/codec/api/svc/codec*.h $(DESTDIR)$(INCLUDE_PREFIX)
 
 install-static-lib: $(LIBPREFIX)$(PROJECT_NAME).$(LIBSUFFIX) install-headers
 	mkdir -p $(DESTDIR)$(PREFIX)/$(LIBDIR_NAME)
--- a/openh264.pc.in
+++ b/openh264.pc.in
@@ -1,6 +1,6 @@
 prefix=@prefix@
 libdir=@libdir@
-includedir=${prefix}/include
+includedir=${prefix}/include/openh264
 
 Name: OpenH264
 Description: OpenH264 is a codec library which supports H.264 encoding and decoding. It is suitable for use in real time applications such as WebRTC.


dpocock:
There are a couple of reasons why this is important:

If people are using ```cisco/openh264``` as a Git submodule, the API files are accessed in a subdirectory without the ```wels/``` prefix, those files [currently live here](https://github.com/cisco/openh264/tree/master/codec/api/svc), notice that ```wels/``` is not part of the path.

The current version of the ```openh264.pc``` file simply lacks the ```wels/``` element and some projects include that as part of ```#include``` while other projects are hacking it into their ```CFLAGS```.  Having a recommended value in the ```.pc``` file encourages conformity

joakim-tjernlund:
This makes sense to me as well.

jrmuizel:
This change broke ffmpeg's detection and use of openh264. Was that intentional?

dpocock:
@jrmuizel thanks for the feedback

I'm guessing the problem you have is one of two things:
1. before this commit, the ```includedir``` in the OpenH264 pkgconfig file didn't have any path element such as ```wels/``` or ```openh264/``` on the end
2. you have the ```wels/``` path element hardcoded in ```configure``` and/or something else

https://people.freedesktop.org/~dbn/pkg-config-guide.html
gives examples with the project name at the end of ```includedir``` in the package config file, e.g. ```Cflags: -I${includedir}/foo``` and that pattern has been followed in this commit.

Projects using a library like that can simply remove ```wels/``` prefixes from the ```configure``` logic and also from those C files that ```#include``` this code.  This is useful if you sometimes want to build with a copy of OpenH264 as a Git submodule, see the comment at the top.

Alternatively, that part of the commit, the [change to openh264.pc.in](https://github.com/cisco/openh264/pull/3415/files#diff-0796f95ab1634277a44e5dc3d1db69b423f23178502fa249bf1c4b079818fa85) could be revised

jrmuizel:
Yes, ffmpeg has `wels` hardcoded: https://github.com/FFmpeg/FFmpeg/blob/129f5ed87e716788191a073791a2446a6099b72f/configure#L6565
https://github.com/FFmpeg/FFmpeg/blob/129f5ed87e716788191a073791a2446a6099b72f/libavcodec/libopenh264.c#L23

How should software like ffmpeg support building against versions of openh264 before and after this commit?

dpocock:
When OpenSSL changed their API it was much more substantial as it required code changes and people developed various hacks with macros to support both new and old.  That was more painful because OpenSSL libraries exist in many Linux distributions.

OpenH264 is very different: because of the patent issue, it is not pre-installed in packages on most systems.  Therefore, it may not justify a lot of effort to support older versions.

jrmuizel:
> OpenH264 is very different: because of the patent issue, it is not pre-installed in packages on most systems. Therefore, it may not justify a lot of effort to support older versions.

Fair enough.

FWIW, I think it would be better to drop the change to `openh264.pc.in`. The header names (e.g `codec_api.h`) are very generic so I think it makes more sense for applications to use `#include <openh264/codec_api.h>` vs `#include <codec_api.h>`

dpocock:
> FWIW, I think it would be better to drop the change to `openh264.pc.in`. The header names (e.g `codec_api.h`) are very generic so I think it makes more sense for applications to use `#include <openh264/codec_api.h>` vs `#include <codec_api.h>`

This is quite interesting and it does challenge the examples from pkgconfig

My own view on that:
- for a library that is being included in every part of ffmpeg (or anything else), e.g. a logging library, that would be a big issue
- but for openh264, most applications only include it into some module in a dedicated subdirectory with a dedicated Makefile where the include path can be configured.  No other module would have the openh264 headers in the include path because the other modules all have different Makefiles

There is another way forward though: maybe we could have a top level ```openh264.h``` file that simply includes all the other public headers

joakim-tjernlund:
Generic names like codec_api.h also risks collision with some other libs headers. openh264 is not a general/useful to many apps that it motivates placing in /usr/include. An extra openh264 dir feels appropriate.

dpocock:
There is no suggestion that codec_api.h should be placed in /usr/include

This commit has not placed codec_api.h in /usr/include

The only question here is which permutation of the path should be in the pkgconfig file: should pkgconfig tell build systems to add ```-I/usr/include/openh264``` to their compiler command line?

I posted a patch for ffmpeg here along with more comments:

http://ffmpeg.org/pipermail/ffmpeg-devel/2022-March/293591.html

joakim-tjernlund:
Oh, my mistake. Sorry about that.

mstorsjo:
> There are a couple of reasons why this is important:
> 
> If people are using `cisco/openh264` as a Git submodule, the API files are accessed in a subdirectory without the `wels/` prefix, those files [currently live here](https://github.com/cisco/openh264/tree/master/codec/api/svc), notice that `wels/` is not part of the path.
> 
> The current version of the `openh264.pc` file simply lacks the `wels/` element and some projects include that as part of `#include` while other projects are hacking it into their `CFLAGS`. Having a recommended value in the `.pc` file encourages conformity

Wouldn't it have been more straightforward, without breaking multiple existing users, by just restructuring the header hierarchy in the repo, to have the headers in a `wels` subdirectory, so that users can consistently use `<wels/codec_api.h>` both when using the code as-is as a submodule and when installed? (And one could have kept the old location with a symlink to avoid breaking existing submodule users too!)

This change has just recently been merged and isn't part of any release (as far as I know?), so we still have time to get this right.

mstorsjo:
If taking a step back - a change like this should be accompanied by a clear declaration - what's the official way of including the library's headers?

Before this change, it was to install the library into a directory, use the pkg-config file and include `<wels/codec_api.h>`. This behaviour has now been broken.

Outside of that, users of the library may have been including `<codec_api.h>` in nonstandard setups by pointing their build tools directory at a non-installed repo checkout of openh264.

This patch broke the existing official usecase and instead made a different, nonstandard usecase the only supported one.

As there are concerns about `codec_api.h` being a very common header name (it's quite plausible that another project also has got a header named that, and thus, including `<codec_api.h>` is brittle and could accidentally load another header from another project, that's a great argument for keeping the directory prefix, i.e. `<somedirectory/codec_api.h>` would be the official recommended way of including it.

Now if we're back at that point, changing the directory from `wels` to `openh264` might make it more understandable, but it's also an entirely avoidable and pointless break.

So if we want a common consistent way of including the library headers from the repo, mimicing the installed layout (with a `wels` subdirectory) would be a straightforward way of achieving that.

mstorsjo:
In #3488 I made a PR to revert this decision.

mstorsjo:
> If anybody wants to search for code using the old header path
> 
> ```
> find . -type f -exec grep -H 'wels/' {} \;
> ```
> 
> and to replace
> 
> ```
> find . -type f -exec sed -i -e 's!wels/!openh264/!g' {} \;
> ```

Really, nobody is arguing that it is hard to change. The point is that there’s no need to change things in the first place. It’s just extra churn for countless of projects using the library, for no benefit.

dpocock:
> The point is that there’s no need to change things in the first place. It’s just extra churn for countless of projects using the library, for no benefit.

Most people would be using this through ffmpeg or GStreamer and patches are available for both of those.  It would be interesting to identify other users.

Things like this do change in libraries from time to time and people just deal with it as part of their release process.

Notice that the pull request combines different changes, each one has different benefits

The rename from wels/ to openh264/ makes code more readable, easier for new developers coming into these projects.  There is now progress [on making packages of the headers for Debian](https://bugs.debian.org/cgi-bin/pkgreport.cgi?owner=bastiangermann%40fishpost.de) and that may mean the headers end up on a lot more systems.  If the headers are installed on a lot of systems then it is better to have them under a directory with an obvious name like ```/usr/include/openh264```

BtbN:
The thing is, with how the patch is right now, you're expected to change from `#include <wels/codec_api.h>`  to just `#include <codec_api.h>`, since include path given in the new pkg-config file already includes the openh264 part.
Those header names are awfully generic to include without any kind of prefixing directory.

At the very least, that needs fixed. But really, I also don't see the point in renaming this for no gain. You can do stuff like that on a major bump that breaks ABI, but not randomly.

dpocock:
> The thing is, with how the patch is right now, you're expected to change from `#include <wels/codec_api.h>` to just `#include <codec_api.h>`, since include path given in the new pkg-config file already includes the openh264 part.

That is not actually true: in most cases, the build system has probably added the higher level directory (e.g. /usr/include) to the include path too so code can use `#include <openh264/codec_api.h>` and it will work

Another enhancement may involve renaming codec_api.h to openh264_api.h

mstorsjo:
> > The point is that there’s no need to change things in the first place. It’s just extra churn for countless of projects using the library, for no benefit.
> 
> Most people would be using this through ffmpeg or GStreamer and patches are available for both of those. It would be interesting to identify other users.

So you basically assume that there’s only 2 projects out there using the API?

> Things like this do change in libraries from time to time and people just deal with it as part of their release process.

Yes, projects have breaking API changes occasionally when it’s justified and needed. Here I still don’t see any justification.


> The rename from wels/ to openh264/ makes code more readable, easier for new developers coming into these projects. There is now progress [on making packages of the headers for Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=974678) and that may mean the headers end up on a lot more systems. If the headers are installed on a lot of systems then it is better to have them under a directory with an obvious name like `/usr/include/openh264`

Is it a blocker for Debian to use the pre-existing include path that has worked fine for nearly a decade? Or is it just that you don’t like the name?

Nevcairiel:
> That is not actually true: in most cases, the build system has probably added the higher level directory (e.g. /usr/include) to the include path too so code can use `#include <openh264/codec_api.h>` and it will work
> 

Not everyone is on Linux or uses system-installed headers in default locations. That argumentation is rather weak.

mstorsjo:
> > The thing is, with how the patch is right now, you're expected to change from `#include <wels/codec_api.h>` to just `#include <codec_api.h>`, since include path given in the new pkg-config file already includes the openh264 part.
> 
> That is not actually true: in most cases, the build system has probably added the higher level directory (e.g. /usr/include) to the include path too so code can use `#include <openh264/codec_api.h>` and it will work
> 
> Another enhancement may involve renaming codec_api.h to openh264_api.h

But why change at all? `<wels/codec_api.h>` does not conflict with anything, and has been in use for nearly a decade. Any API change like this is disruptive. The interface is settled since ages ago and there’s no practical issue, other than that your dislike the name and assume breaking all users is fine.

dpocock:
> > > The point is that there’s no need to change things in the first place. It’s just extra churn for countless of projects using the library, for no benefit.
> > 
> > 
> > Most people would be using this through ffmpeg or GStreamer and patches are available for both of those. It would be interesting to identify other users.
> 
> So you basically assume that there’s only 2 projects out there using the API?

No, but those are some of the more well known

> 
> > Things like this do change in libraries from time to time and people just deal with it as part of their release process.
> 
> Yes, projects have breaking API changes occasionally when it’s justified and needed. Here I still don’t see any justification.
> 
> > The rename from wels/ to openh264/ makes code more readable, easier for new developers coming into these projects. There is now progress [on making packages of the headers for Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=974678) and that may mean the headers end up on a lot more systems. If the headers are installed on a lot of systems then it is better to have them under a directory with an obvious name like `/usr/include/openh264`
> 
> Is it a blocker for Debian to use the pre-existing include path that has worked fine for nearly a decade? Or is it just that you don’t like the name?

Different system administrators and package maintainers would have different attitudes to that.

The way I think about it personally: if every project had an obscure name for their header directory then it would be unreasonably tedious for people to identify things in /usr/include or in the `#include` statements.

xclaesse:
@GuangweiWang I think this PR needs to be reverted immediately, it breaks everyone for no benefit. Note that it is also inconsistent with the Meson port that still install header files in the old location and generate a pkg-config file as before.

xclaesse:
The other solution is to drop all Makefile and rely on Meson only, because it already copy all headers into `wels/` subdir so they can be used uninstalled. :D

dpocock:
> The other solution is to drop all Makefile and rely on Meson only, because it already copy all headers into `wels/` subdir so they can be used uninstalled. :D

The existence of constructive solutions like this demonstrates that the "reverted immediately" reaction was never optimal, it is far better to allow a little bit of time for the issue to percolate and then move forward

xclaesse:
> it is far better to allow a little bit of time for the issue to percolate and then move forward

Leaving time for more people to wonder why their build fail? What do we gain from that?

dpocock:
> > it is far better to allow a little bit of time for the issue to percolate and then move forward
> 
> Leaving time for more people to wonder why their build fail? What do we gain from that?

Many people may have already updated their build by now and would have to make extra effort if it is reverted.  The effort of changing the name is so trivial that some people may have already done it without feeling it was worth the effort to post a comment here.  Sadly those people don't have a voice in discussions like this.

Libraries do make changes like this from time to time, this is a relatively trivial case.

Which part of it concerns you more personally, the directory name `openh264` or the (non)inclusion of the suffix in the pkgconfig file?  I'm more sympathetic to people who argue for a partial revert that only impacts pkgconfig.

BtbN:
I highly doubt that.
This isn't even released yet, so majority of people haven't even noticed.
It's just pointless breakage for no advantage, forcing downstream projects to either add hacky ifdef hell, or require super bleeding edge versions.

Please just revert it. Or remove the entire buildsystem if you prefer... having two build systems that produce incompatible results is ridiculous anyway.

xclaesse:
> Most people would be using this through ffmpeg or GStreamer and patches are available for both of those. It would be interesting to identify other users.

Do you have link to the proposed patches? Did they merge it yet?

mstorsjo:
> > Most people would be using this through ffmpeg or GStreamer and patches are available for both of those. It would be interesting to identify other users.
> 
> Do you have link to the proposed patches? Did they merge it yet?

Not merged in ffmpeg at least, because we saw that this had only been merged for a couple of days and the change was entirely pointless, so I requested a revert https://github.com/cisco/openh264/pull/3488 immediately afterwards.

BtbN:
The proposed patch to ffmpeg was also breaking backwards compatibility, for a version without a release, and was wrong due to the wrong patch in the pkg-config file.

eli-schwartz:
> Many people may have already updated their build by now and would have to make extra effort if it is reverted.

This is a logically inconsistent statement on several levels.

- you formerly justified this PR with the statement:
  > Most people would be using this through ffmpeg or GStreamer and patches are available for both of those. It would be interesting to identify other users.

  i.e. you believe that very few people "actually" use openh264, and that no one needs to update their builds other than two projects, one of which responded to this PR by flatly rejecting your patch, refusing to fix their build, and demanding this PR be reverted
- anyone who has updated their build but (unlike you, Mr. Pocock) cares about API stability and backwards compatibility, has to have two codepaths, one for before this change, one for after this change. So, if they support both then they are also compatible with having this change reverted.

> The effort of changing the name is so trivial that some people may have already done it without feeling it was worth the effort to post a comment here.

Everyone who did post a comment here insists that the change is so non-trivial that they refuse to do it because the change is too hard and breaks everything, so "please instead revert the change here".

> Sadly those people don't have a voice in discussions like this.

??? What does that mean, they don't have a voice?

Of course they have a voice. They may decline to exercise the voice, but they certainly have it. If they feel that reverting this change is bad, they are empowered to comment here.

Let's turn the argument around. You made a change that a bunch of people feel was bad. Why did you submit this PR? What about the people who felt the issue was so obvious that it must clearly be rejected, because it breaks API compatibility?

Sadly, the ffmpeg developers don't have a voice in discussions like this, because they thought it was obviously wrong so they didn't feel it was worth posting a comment.

> Libraries do make changes like this from time to time, this is a relatively trivial case.

Libraries do NOT make changes like this, and this is NOT a relatively trivial case.

Or so say a bunch of people who find that this change was NOT trivial, and causes them a lot of problems. :)

xclaesse:
> * one of which responded to this PR by flatly rejecting your patch, refusing to fix their build, and demanding this PR be reverted

and the other one uses the Meson port that has not been changed here, so they cannot accept such a patch anyway. I also failed to see a MR in GStreamer, did you really submitted one? Do you have a link?

eli-schwartz:
> Is it a blocker for Debian to use the pre-existing include path that has worked fine for nearly a decade? Or is it just that you don’t like the name?

I took a look at this.

It turns out that this PR was merged on February 23, and on March 1 @dpocock mentioned "There is now progress [on making packages of the headers for Debian](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=974678)".

However, Debian actually uploaded the package of the headers a month before that, on January 18 the headers package was uploaded to Debian unstable.

The Debian package uses the wels/ directory. The Debian package would be broken by this PR.

@dpocock was not involved in the Debian bug at all, although due to https://www.debian.org/News/2021/20211117 he may have tried but lacked the opportunity. Either way, before proposing that this PR is a Debian requirement, we should probably ask Debian whether they want this patch.

With that in mind, @bgermann seems to be the Debian maintainer of libopenh264-dev, perhaps we can get some enlightenment here on what Debian's preferences are?

mstorsjo:
> Sadly, the ffmpeg developers don't have a voice in discussions like this, because they thought it was obviously wrong so they didn't feel it was worth posting a comment.

Actually, 3 of us commenting here are ffmpeg developers - we just didn’t spell out our affiliation.

I made the revert PR #3488 as soon as I had understood full picture here.

eli-schwartz:
Ah, to clarify: "Let's turn the argument around."

That was my attempt at humorously inverting the scenario, and theorizing that in between Aug 30, 2021 and February 23, 2022, your failure to comment on an unmerged PR constitutes "not having a voice".

According to the logic above, if people "don't have a voice" we should be defensive of their voice by not doing anything. This logic is currently being used to say that the PR should not be reverted, because these theoretical voiceless people can't say anything in defense of the PR.

So my line of thinking is that a couple months ago, the same logic should have been used to reject merging this PR to begin with.

(Alternatively, people who want to have a voice, will exercise that voice. ffmpeg has done so after the merge, but before a stable release. Anyone who actually likes this PR is welcome to volunteer their perspective at any time, but hopefully also before a stable release.)

dpocock:
Back to practical actions we can take:

I personally feel the idea from @xclaesse dropping the old build script completely and focusing on the Meson build script is the best.  @BtbN appears to be saying the same thing too.  Coincidentally, that would also drop the changes introduced by this PR and it makes #3488 redundant

@xclaesse do you want to create a pull request for your proposal?  If that is accepted by the project owner then we can restart this discussion after that.

The Gstreamer patch is on this branch:
https://gitlab.freedesktop.org/pocock/gst-plugins-bad/-/tree/pocock/openh264-include-wels




> Libraries do NOT make changes like this, and this is NOT a relatively trivial case.

The OpenSSL 1.0 to 1.1 change involved code changes too, this only involves a path change, so it is trivial relative to that.

Ultimately, it is for the project owner to decide and everybody else will have to live with it.  I'm happy to update my patches for both ffmpeg and GStreamer to go with the final decision of the project owner and the Meson build if that is what comes about.

mstorsjo:
> I personally feel the idea from @xclaesse dropping the old build script completely and focusing on the Meson build script is the best. @BtbN appears to be saying the same thing too. Coincidentally, that would also drop the changes introduced by this PR and it makes #3488 redundant

So if we agree that we should restart this discussion from a clean slate, let's revert this commit (which is a much smaller decision to make than to remove the build system that, as far as I know, Cisco themselves actually use) and restart it from there. As far as I undestood @xclaesse 's suggestion, it was a tongue-in-cheek comment implying that removing one build system would have the same effect as reverting the patch.

I don't see why that would need to be linked with removing one out of two build systems (which is an entirely orthogonal issue).

> Ultimately, it is for the project owner to decide and everybody else will have to live with it. I'm happy to update my patches for both ffmpeg and GStreamer to go with the final decision of the project owner and the Meson build if that is what comes about.

If the final decision for openh264 is to change its include paths, then one requirement is patches for ffmpeg that transparently work for users both with the existing releases and upcoming ones - without requiring users to jump through any hoops.

dpocock:
> I don't see why that would need to be linked with removing one out of two build systems (which is an entirely orthogonal issue).

If one of the build systems is not removed then I should get to work creating an equivalent patch for the Meson build

But I think it is better to just have one system or the other.  Otherwise we end up with changes like this where somebody forgets to update Meson.

> If the final decision for openh264 is to change its include paths, then one requirement is patches for ffmpeg that transparently work for users both with the existing releases and upcoming ones - without requiring users to jump through any hoops.

This could be streamlined somewhat by only partially reverting the commit, so the path name will definitely be `openh264/*` but the pkgconfig file will be the same as before without any suffix.

dpocock:
@eli-schwartz if you are going to share what is obviously a defamation then please also share the Red Hat case

January, Matthew Miller declares a web site doesn't obey his Code of Conduct (CoC)

March, an independent legal panel looked at evidence presented by both sides and decided not only that Miller was wrong, but went as far as declaring that Miller's attempt to enforce a CoC is harassing a volunteer and abuse of an administrative process:
https://www.adrforum.com/DomainDecisions/1980642.htm

News sites were so quick to ridicule the case that they didn't even spell the trademark correctly in one case
https://www.theregister.com/2022/03/16/red_hat_fedotra/

We can thank Miller for submitting his claim to an independent review.  The rogue elements of Debian have never submitted their gossip to any independent review because they know they are always right, just as Scientologists know they are right

bgermann:
I do not have a strong opinion on this. Please release a version when you have decided on this. It would be great to make that happen this year because the freeze for Debian's first stable release including the openh264 pkg will be next January.

I just want to let you know that I apply a [related patch](https://sources.debian.org/src/openh264/2.2.0%2Bdfsg-2/debian/patches/Correct-.pc-s-libdir.patch/).

eli-schwartz:
@dpocock 
> @eli-schwartz if you are going to share what is obviously a defamation then please also share the Red Hat case

I am not "sharing a defamation" -- who is right and who is wrong isn't topical here.

I'm positing that the result of it is, you are not currently coordinated with Debian in order to authoritatively rule on what Debian considers a libopenh264-dev packaging blocker.

@bgermann 
> I just want to let you know that I apply a [related patch](https://sources.debian.org/src/openh264/2.2.0%2Bdfsg-2/debian/patches/Correct-.pc-s-libdir.patch/).

This patch is entirely unrelated. The .pc currently hardcodes:
- includedir
- libdir

and your patch fixes the latter, this PR fixes neither one but swaps the former out for a different hardcoded location.

This PR also makes a second change, to remove the wels/ directory (the includedir specifies the location of the wels/ directory) entirely.

ystreet:
> That is not actually true: in most cases, the build system has probably added the higher level directory (e.g. /usr/include) to the include path too so code can use `#include <openh264/codec_api.h>` and it will work

That is not guaranteed at all and requiring that shows that this solution is all the more broken.

> The Gstreamer patch is on this branch:
> https://gitlab.freedesktop.org/pocock/gst-plugins-bad/-/tree/pocock/openh264-include-wels

A patch on an unmerged branch is not anywhere close to actually being merged in the upstream project.  Speaking with my GStreamer maintainer hat on, this branch would not be merged as-is because it doesn't support both the old openh264 include layout and the new openh264 include layout introduced by this PR.  This requirement would be the same as ffmpeg.

I would very much vote in favour of reverting this PR entirely as it doesn't actually seem to provide any real benefit.

huili2:
After thoroughly thinking and discussion, we decide to revert this PR to keep the code backward-compatibility. (PR #3488 would be merged then.)
@dpocock actually we first thought it was really good to make code easier-to-understand, we forgot that this would bring much more effort to do for people who adopt it. We wish we had a better name at the very beginning!
Again, thank all of you for your precious voice! We hope that we can hear from you again in the future!

PR #3431, opened 2021-10-02
meson: Let meson decide what kind of lib we are building

thiblahute:
This was done to behave the same way as the `Makefile` but with meson
the user should decide what to build and use `--default-library=both`
to build both static and dynamic libs.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -191,17 +191,12 @@ all_objects = [
   libdecoder.extract_all_objects()
 ]
 
-libopenh264_shared = shared_library('openh264',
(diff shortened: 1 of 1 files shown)


MathieuDuponchelle:
lgtm :+1:

xclaesse:
@thiblahute ahah, I've been a year ahead of you: https://github.com/cisco/openh264/pull/3351

xclaesse:
Let's close this as dup if you don't mind, my PR does a few more cleanups.

PR #3458, opened 2022-01-09
Update meson build files

nirbheek:
```
build: Update version numbers to 2.1.1

This was forgotten when doing the release.
```

```
meson: Fix warning about extract_all_objects usage

We use this because of a meson bug that was fixed in 0.52:
```
https://mesonbuild.com/Release-notes-for-0-52-0.html#improved-support-for-static-libraries
```
Bump the requirement and remove the extract_all_objects workaround.
This gets rid of a meson warning:

WARNING: extract_all_objects called without setting recursive
keyword argument. Meson currently defaults to
non-recursive to maintain backward compatibility but
the default will be changed in the future.
```

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - run: sudo apt-get update -qq
       - run: sudo apt-get install -qq nasm g++-multilib gcc-multilib libc6-dev-i386 python3-pip python3-setuptools
-      - run: sudo python3 -m pip install meson==0.50.1 ninja
+      - run: sudo python3 -m pip install meson==0.52.1 ninja
       - run: nasm -v 
       - run: c++ --version
       - run: make gmp-bootstrap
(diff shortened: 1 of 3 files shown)


eli-schwartz:
Commit 2 is once again an issue which would be solved by #3351

nirbheek:
Maybe this one will be simpler to review and merge for the maintainers! :)

joakim-tjernlund:
Should not master be updated to whatever the next version will be called? Probably as a separate commit though.
I guess next version also changes ABI so .so version for shared libs should be bumped?

eli-schwartz:
That is traditionally something done during the release process itself.

Some projects will as a matter of personal preference do a post-release version bump to a `*-dev` or `*.dev1` suffixed version number, but that's very much again, a matter of personal preference. Does this project happen to have such a policy?

joakim-tjernlund:
Version number has been updated in master now but not the external_all_objects meson rule

nirbheek:
> can you help resolve the conflicts?

Done!

PR #3465, opened 2022-01-20
support Windows arm64 in build script

GuangweiWang:
as the title.

change:
1. support arm64 compilation on Windows platform.
2. build script enhancement.

### DIFF
--- a/build/AutoBuildForWindows.bat
+++ b/build/AutoBuildForWindows.bat
@@ -1,45 +1,43 @@
 @echo off
 rem *************************************************************************************************
 rem   usage:
-rem      AutoBuildForWPAndWindows.bat % Configuration %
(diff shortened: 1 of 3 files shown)


huili2:
LGTM

mstorsjo:
This commit contained LOTS of unrelated changes:
- adding `-cetcompat` flag which is an x86 only flag - totally unrelated to this PR
- adding `-WX` which makes the build error out on any warning, which is a problem if end users compile it with a different compiler version
- Adding lots of unnecessary and confused `-DWIN32` and similar defines. They have not been needed so far (the C/C++ code only uses the compiler built-in define `_WIN32` which doesn't need to be defined manually

I made PR #3470 which cleans up all these aspects, please have a look at it.

PR #3470, opened 2022-01-23
Clean up MSVC build after c256ae3b2774030966913e47a9088ec47d053535

mstorsjo:
c256ae3b2774030966913e47a9088ec47d053535 contained lots of unnecessary changes to the MSVC build configuration. This removes the unnecessary/redundant changes that were part of that commit.

When merging, please use "create merge commit" or "rebase and merge". Please DO NOT use "squash and merge" for this PR. The changes are split up into individual changes that each explain exactly what's changed and why, which is important information to keep in the commit log for posteriority, while "squash and merge" would lose all that information. (If it's possible to merge in any other way than "squash and merge", I can make one separate PR for each change, but that's a bit more overhead for everybody.)

### DIFF
--- a/build/msvc-common.mk
+++ b/build/msvc-common.mk
@@ -24,9 +24,6 @@ CXX = clang-cl
 CCAS = clang-cl
 CFLAGS += -DHAVE_NEON_AARCH64 --target=arm64-windows
 CCASFLAGS = -nologo -DHAVE_NEON_AARCH64 --target=arm64-windows
-LDFLAGS += -link
-else
-LDFLAGS += -link -cetcompat
 endif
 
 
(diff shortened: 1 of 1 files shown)


GuangweiWang:
quote from commit message:

* msvc: Remove unnecessary defines of WIN32/_WIN64

_WIN32 and _WIN64 are defined automatically by the compiler and
never has to be specified manually.

WIN32 and WIN64, without leading underscores, are never defined
automatically by the compiler, but some Visual Studio project
templates contain such a manually specified define. These aren't
used by the code base, it only relies on the automatically defined
_WIN32.

The only exception is for the x86 assembly, built by the external
nasm/yasm tools, where we manually pass -DWIN64.

These defines were unnecessarily added in
c256ae3b2774030966913e47a9088ec47d053535.

Also get rid of one unnecessarily added -nologo option; this is
added unconditionally later in CFLAGS.

* msvc: Stop unconditionally building with -WX (erroring out on any warning)

This was added without a reason in
c256ae3b2774030966913e47a9088ec47d053535.

Options like -Werror or -WX may make sense when built as part of CI,
to catch warnings in those known warning-free configurations. But
the build system itself shouldn't enable such errors by default,
as that can break building for users with different compiler versions.

* msvc: Clean up setting of the -cetcompat linker flag

Pass this flag specifically if the architecture is an x86 architecture.
(The previous code form would erroneously add this flag when building
for 32 bit arm too.)

This flag was added unrelatedly without any specific argument for
adding it, in c256ae3b2774030966913e47a9088ec47d053535.

* msvc: Remove a redundant define of -DHAVE_NEON_AARCH64 for CFLAGS

This flag is already set in arch.mk.

* msvc: Build the C/C++ code with MSVC, not clang-cl, for ARM64

c256ae3b2774030966913e47a9088ec47d053535 switched CCAS to clang-cl.
Previously it was built with gas-preprocessor.pl, which also works,
but is a bit more cumbersome to set up, and clang-cl is distributed
with MSVC these days.

c256ae3b2774030966913e47a9088ec47d053535 also switched CC and CXX
to clang-cl, but that's not strictly needed.

We could of course want to compile all of the code using clang-cl
instead of MSVC's cl.exe itself, but that should probably be a
separate option then, not tied to the architecture.

GuangweiWang:
thanks @mstorsjo for so detailed fix.

mstorsjo:
> thanks @mstorsjo for so detailed fix.

Thanks for the quick merge! But I’m disappointed that you couldn’t do the merge in either of the ways I requested, as now all the extra explanations are lost from the commit log. Luckily they’re available here in the comment thread though.

PR #3471, opened 2022-01-24
msvc: Fix parallel building when creating PDB debug info

mstorsjo:
The current way of creating debug info, using the -Zi option,
writes all debug info to one central file. This doesn't work
if building with multiple compiler invocations in parallel. By adding
the -FS option, writing of the PDB file is passed via mspdbsrv.exe,
allowing multiple concurrent compiler processes.

Alternatively, the debug info could be created using the -Z7 flag
instead, which stores the debug info within the object files until
the linking stage, which also works for parallel compilation.

### DIFF
--- a/build/msvc-common.mk
+++ b/build/msvc-common.mk
@@ -32,8 +32,8 @@ endif
 CFLAGS += -nologo -W3 -EHsc -fp:precise -Zc:wchar_t -Zc:forScope -D_VARIADIC_MAX=10
 CXX_LINK_O=-nologo -Fe$@
 AR_OPTS=-nologo -out:$@
-CFLAGS_OPT=-O2 -Ob1 -Oy- -Zi -GF -GS -Gy -DNDEBUG
-CFLAGS_DEBUG=-Od -Oy- -Zi -RTC1 -D_DEBUG
+CFLAGS_OPT=-O2 -Ob1 -Oy- -Zi -FS -GF -GS -Gy -DNDEBUG
+CFLAGS_DEBUG=-Od -Oy- -Zi -FS -RTC1 -D_DEBUG
 CFLAGS_M32=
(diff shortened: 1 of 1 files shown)


GuangweiWang:
I will create another PR for changing '-Zi' to '-Z7'

GuangweiWang:
does `-Zi -FS` is better?

GuangweiWang:
both way is ok for me.  I am just not sure , using `/Zi /FS` for parallel building, if the `/Gm` flag should be remove or not.
if use `/Z7` flag instead, I think the `/Gm` flag should be removed. 
if use `/Zi /FS` flags, does the `/Gm` flag should be removed or kept?

mstorsjo:
> both way is ok for me. I am just not sure , using `/Zi /FS` for parallel building, the `/Gm` flag should be remove or not.
> if use `/Z7` flag instead, I think the `/Gm` flag should be removed.

Yes, I see that that flag should be removed if using `/Z7`.

> if use `/Zi /FS` flags, does the `/Gm` flag should be removed or kept?

I think that flag is meant to speed up incremental rebuilds, so maybe it's worthwhile to keep it, if we can.


Overall for `/Zi /FS` vs `/Z7`, in MSVC environments, `/Zi /FS` is the modern default way of doing things. (Clang-cl internally only implements the equivalent of `/Z7` though, but it doesn't warn about it iirc.) Speedwise, both seem to run pretty much exactly as fast (both for parallel and serial builds), but a build with `/Z7` takes 30 MB more space for the object files.

The only upside with `/Z7` is that it's more robust when building with MSVC in Wine on linux, but that's probably not a usecase worth worrying about here. So I think it's best to just keep the current form for now.

GuangweiWang:
thank you for educating me. let's keep the current way.

PR #3488, opened 2022-03-01
Revert "Makefile, pkg-config: Add INCLUDE_PREFIX variable, use includ…

mstorsjo:
…e/openh264 as default, include correct default for pkg-config (#3415)"

This reverts commit ccf65bc7b1eb257980b02528133b3a5a65e1497c.

This change broke the existing official way of including the headers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,6 @@ CFLAGS_DEBUG=-g
 BUILDTYPE=Release
 V=Yes
 PREFIX=/usr/local
-INCLUDE_PREFIX=$(PREFIX)/include/openh264
 SHARED=-shared
 OBJ=o
 DESTDIR=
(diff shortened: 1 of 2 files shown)


dpocock:
Reverting feels like a bit of an overreaction, I agree the contribution in #3415 could be made more convenient for people who want both new and old versions.  That could be done with symlinks or macros.

For any new projects using OpenH264 it would be a good idea to use the openh264/*.h path

quink-black:
An old story, [What did Ken Thompson mean when he said, "I'd spell creat with an 'e'."?](https://unix.stackexchange.com/questions/10893/what-did-ken-thompson-mean-when-he-said-id-spell-creat-with-an-e)

You can't fix a misspelling `creat()` which break all down stream projects.

dpocock:
I suspect these headers will never be known as widely as something like the `creat` syscall

While the idea of reverting the change is a valid opinion, one reason why I felt the pull request is an overreaction is that we now have three separate discussions going on: here in #3488 and in the original #3415 and also in the [ffmpeg mailing list thread](http://ffmpeg.org/pipermail/ffmpeg-devel/2022-March/293591.html)

I submitted the original pull request a long time ago, August 2021.  There was one comment in December and then it was merged.  When other concerns appeared this week, I immediately came to the ffmpeg-devel community, I'm always happy to support my contributions.  Maybe the timing could be changed or maybe the patch can be improved but before any decision to revert or adapt the patch, I would encourage people to share their thoughts in the original pull request because that is where most of the voices have been this week, #3415

PR #3514, opened 2022-04-26
Rate control fix

YCSun-Meta:
1. Fix rate control to not overshoot target. Bitrate deficit wasn't carried over to next VGOP, so we lose track of it and hence could end up overshooting.
2. Carry over RC budget on key frame insertion. Add higher QP for extremely low bitrates.
3. Fix overshoot with frame skipping inside codec. CheckFrameSkipBasedMaxbr() is only used inside  WelsRcCheckFrameStatus() which is called from PrepareEncodeFrame() to check if frame must be skipped. If it is then UpdateBufferWhenFrameSkipped() will be called in PrepareEncodeFrame() and will update same bitrate and buffer counters again. This results in overshoot as rate control thinks it has more bits than it actually has.
4. Add control on bit allocation ratio of key frames.


YCSun-Meta:
We tested this change on the video calling apps from Meta with public A/B test. The results show significant video freeze improvement and vbv delay reduction.

codec/encoder/core/src/ratectl.cpp
```diff
@@ -548,11 +571,17 @@ void RcDecideTargetBits (sWelsEncCtx* pEncCtx) {
   SRCTemporal* pTOverRc         = &pWelsSvcRc->pTemporalOverRc[pEncCtx->uiTemporalId];
 
   pWelsSvcRc->iCurrentBitsLevel = BITS_NORMAL;
+  const bool fix_rc_overshoot = pEncCtx->pSvcParam->bFixRCOverShoot;
   //allocate bits
   if (pEncCtx->eSliceType == I_SLICE) {
-    pWelsSvcRc->iTargetBits = pWelsSvcRc->iBitsPerFrame * IDR_BITRATE_RATIO;
+    if( pWelsSvcRc->iIdrNum != 0 ){
+      pWelsSvcRc->iTargetBits = pWelsSvcRc->iBitsPerFrame * pEncCtx->pSvcParam->iIdrBitrateRatio / 100;
```
huili2, line 578: By default "sParam.iIdrBitrateRatio = IDR_BITRATE_RATIO;"
then here it is to be divided by 100? Is it wrong or the original value should be 100 * it? Otherwise, the non-first IDR will be allocated with too few bits, compared with that of the first IDR.

YCSun-Meta: Hi @huili2,
Thank you for pointing this out. You are right. The default value of iIdrBitrateRatio should be 100* IDR_BITRATE_RATIO. I have fixed this.

huili2 (approved):

YCSun-Meta:
Hi @huili2,

Thank you for the stamp. 
I have a question: since the pull request changes the encoding algorithm, we have to fix test references as https://github.com/cisco/openh264/pull/3272/commits

However, I cannot find any script to fix test references. Could you advise me on how I could do this efficiently? 
Thank you.

Hi @mstorsjo,
Could you advise me on how you fixed this last time? Thank you.

mstorsjo:
> Hi @mstorsjo, Could you advise me on how you fixed this last time? Thank you.

I just manually updated hashes by copypasting and rerunning the tests - as far as I know, there's no script for it.

The testcases that can have more than one test reference are tricky though (https://github.com/cisco/openh264/blob/fd92219def875f8a7c74f1007d3a6f6e1bf67671/test/api/decode_api_test.cpp#L782-L786), as you need to produce new hashes for both cases. I don't remember exactly, but I think you can achieve that by disabling the x86 specific routines in `CDownsampling::InitDownsampleFuncs`.

YCSun-Meta:
I attempted to fix test references error. Could you please approve for running the workflows. @huili2 Thanks.

YCSun-Meta:
I see that the check has passed, so probably need the final approval and someone with right access to merge. @huili2 Thanks for you help.

PR #3521, opened 2022-05-25
Add a Windows build to Github Actions.

jrmuizel:
This is just a basic build and run of the unit tests that can be expanded on in a followup.


kanzhu-cisco:
After the patch, both linux and windows will build for test?
image

GuangweiWang:
@kanzhu-cisco
checks are triggered now.

PR #3524, opened 2022-05-30
remove warnings

GuangweiWang:
1. remove several warnings
2. astyle
3. pass flag "-Werror" to compiler on Linux and MacOS


mstorsjo:
This is wrong. We should **not** be adding `-Werror` in the makefiles itself!

By adding `-Werror` in the build files, you make sure that when someone checks out an older version of openh264 and try to build it with a new compiler, it fails, just because newer/different compilers have more diagnostics than before. Or when building with an older compiler, or a different one than the ones that have been tested.

As long as the code itself is standard C++, we shouldn't be creating extra hurdles for users with other, standard C++ compliant compilers to build the coe.

If we want the CI environment to enforce that there are no warnings, we should be adding `-Werror` in the CI environment only!

Also, looking at the build files, in `run_Test.sh` there already is such a case, so I don't really see what this gains with respect to the CI environment. If you don't want warnings in macOS builds, then add a macOS configuration in CI, but don't enforce `-Werror` on builds outside of CI.

mstorsjo:
I can try to make a PR to fix up the handling of `-Werror` in CI, and add a few more CI configurations.

PR #3525, opened 2022-05-31
enable option class-memaccess to gcc8 above only

GuangweiWang:
hi @mstorsjo 
if possible, can you help to review this PR? which is related to #3524 . thanks.


build/platform-linux.mk
```diff
@@ -3,13 +3,20 @@ SHAREDLIBSUFFIX = so
...
 LDFLAGS += -lpthread
 STATIC_LDFLAGS += -lpthread -lm
 AR_OPTS = crD $@
+
+ifeq ($(USE_STACK_PROTECTOR), Yes)
+CFLAGS += -fstack-protector-all
+endif
+
+GCCVER_GTEQ8 = $(shell echo $$((`gcc -dumpversion | awk -F "." '{print $$1}'` >= 8)))
+ifeq ($(GCCVER_GTEQ8), 1)
```
mstorsjo, line 16: This isn't really reliable. What if you're building with a case with `CC=` set to something else than `gcc` (see e.g. the clang++ case at the bottom of the file)?

PR #3529, opened 2022-06-02
update openh264 version to v2.3.0

GuangweiWang:
as the title.


eli-schwartz:
Is there going to be a corresponding Github Releases tag created?

GuangweiWang:
> Is there going to be a corresponding Github Releases tag created?

yes.

eli-schwartz:
So two weeks later, I'm kind of wondering what that's even supposed to mean.

Is that "yes" as in "oh, we forgot to do that while bumping the version and deploying new binaries", or "yes" as in "no", or "yes" as in "probably someday but don't hold your breath"? Something else I haven't thought of?

PR #3559, opened 2022-08-30
fix oss-fuzz issue-45607. checking if reference pic count < maximum r…

xiaotiansf:
…eference count allowed.


jrmuizel:
Is it practical to add a test case?

xiaotiansf:
@jrmuizel Might be possible to add a test case. But the checking for reference pic count is based on H.264 specifications. So it may not be necessary to add test case,

PR #3565, opened 2022-09-06
update openh264 version

GuangweiWang:


mcatanzaro:
Can you merge this into master as well, please? You don't want to leave the master branch with the older API version.

Looking forward to the release.

xclaesse:
@GuangweiWang this PR is still missing in master branch I think.

hmaarrfk:
I also hit the fact that this was missing in the master branch today.

mcatanzaro:
> I also hit the fact that this was missing in the master branch today.

It's not though. It landed via 0c19a5c266275693484ed84edf6ff8cb7cd89d4e on the same day that Xavier complained in the comment above.

PR #3566, opened 2022-09-06
remove waring: maybe-uninitialized

GuangweiWang:
as the title.

### DIFF
--- a/test/decoder/DecUT_PredMv.cpp
+++ b/test/decoder/DecUT_PredMv.cpp
@@ -175,6 +175,9 @@ TEST (PredMvTest, PredMv) {
   const int32_t kiRandTime = 100;
   bool bOK = true;
 
+  memset(&sWelsMvPred, 0, sizeof(SWelsMvPred));
+  memset(&sAncMvPred, 0, sizeof(SAnchorMvPred));
+
   //test specific input: 16x16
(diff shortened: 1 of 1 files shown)


test/decoder/DecUT_PredMv.cpp
```diff
@@ -249,6 +255,9 @@ TEST (PredMvTest, PredInter8x16Mv) {
   const int32_t kiRandTime = 100;
   bool bOK = true;
 
+  memset(&sWelsMvPred, 0, sizeof(SWelsMvPred));
+  memset(&sAncMvPred, 0, sizeof(SAnchorMvPred));
+
```
jrmuizel, line 260: How about the more type safe `SWelsMvPred sWelsMvPred = {}` instead?

huili2: Can I know why this is more type safe, given that we know the definition of the structure?

jrmuizel: Because you can accidentally mix up the sizes e.g.:
```c
  memset(&sWelsMvPred, 0, sizeof(SAnchorMvPred));
  memset(&sAncMvPred, 0, sizeof(SWelsMvPred));
  ```

PR #3583, opened 2022-10-08
fix unsigned integer(uint32_t) overflow

skzxcc:


jrmuizel:
Do you have a test case that triggers this?

skzxcc:
It's easily triggered when working with HD video sources with complex textures on BitrateMode ratecontrol
you can trigger it,  run ".\encConsole.exe test-welsenc.cfg"















At 2022-10-08 22:19:10, "Jeff Muizelaar" ***@***.***> wrote:

Do you have a test case that triggers this?

huili2:
still suggest have a detailed case for debugging, as this will be assigned to other variables.

skzxcc:
> 仍然建议有一个详细的案例进行调试，因为这将被分配给其他变量。

Can you provide an email address? I will send you the video source file and configuration file,  you can easily trigger it.

Issue #3585, opened 2022-10-21
DyadicBilinearQuarterDownsampler_sse fails to preserve xmm7, can cause undefined behavior


randomascii:
The calling convention for 64-bit processes on Windows is documented here:

https://learn.microsoft.com/en-us/cpp/build/x64-calling-convention?view=msvc-170

and it says:

"The x64 ABI considers registers RBX, RBP, RDI, RSI, RSP, R12, R13, R14, R15, and XMM6-XMM15 nonvolatile. They must be saved and restored by a function that uses them."

However DyadicBilinearQuarterDownsampler_sse uses xmm7 but does not preserve and restore it. This can cause arbitrary bugs in calling code that assumes that xmm7 will be preserved. This is a possible cause of crashes and memory corruption in Chrome. See crbug.com/1218384 for a long-running discussion, especially the more recent comments.

This appears to be a serious bug, but the fix should be fairly straightforward.

Issue #3563, opened 2022-08-31
mac CI sometimes fails


jrmuizel:
We see intermittent failures on mac CI. e.g:
https://github.com/cisco/openh264/runs/8100566320?check_suite_focus=true

This could be related to #3562

huili2:
@xiaotiansf We also meet with such kinds of UT failures occasionally. Is it caused by multi-thread decoding?

xiaotiansf:
I'll look into the issue.

PR #3586, opened 2022-10-21
Save xmm7 in DyadicBilinearQuarterDownsampler_sse

randomascii:
Static analysis of chrome.dll showed that xmm7 was being used but not preserved in this function. The Windows calling convention requires that xmm7 be preserved so this change adds the necessary "PUSH_XMM 8" and POP_XMM directives to fix this.

This fixes issue #3585. This may fix a bug in Chrome but that is unknown.

### DIFF
--- a/codec/processing/src/x86/downsample_bilinear.asm
+++ b/codec/processing/src/x86/downsample_bilinear.asm
@@ -1948,6 +1948,7 @@ WELS_EXTERN DyadicBilinearQuarterDownsampler_sse
     %assign push_num 0
 %endif
     LOAD_6_PARA
+    PUSH_XMM 8
     SIGN_EXTENSION r1, r1d
     SIGN_EXTENSION r3, r3d
     SIGN_EXTENSION r4, r4d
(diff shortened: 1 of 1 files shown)


randomascii:
Can somebody take a look? Perhaps @huili2 ? This issue may be causing random crashes, and certainly will at some point.

randomascii:
Uggh. I just noticed the failure. Unfortunately I'm not clear on how to analyze it and whether it is possible that my change could have caused it (as opposed to fixing an existing error). Thoughts?

randomascii:
Thinking about this some more - it seems odd that the test failure is on OSX, but the changes are protected by "%ifdef WIN64" (in the PUSH_XMM and POP_XMM macros).

Therefore I believe that the failure is simply a flaky failure, or otherwise unrelated to my change.

jrmuizel:
The macos tests are known to be flaky: #3563

randomascii:
That's good to know (about the test being flaky). Given that, is there a way to force the checks to run again? Or, more to the point, should I keep waiting or is there something else I should do to get this reviewed and landed?

PR #3594, opened 2022-11-28
remove warnings

GuangweiWang:
1. remove warning in clang 14 and fix potential bug.
2. update corresponding SHA1 Table.


mstorsjo:
+1, please go ahead with this patch (but the hash mismatches do need to be fixed - how have they appeared?)

GuangweiWang:
> +1, please go ahead with this patch (but the hash mismatches do need to be fixed - how have they appeared?)
yes. xiaotiaon might check this issue.

test/utils/HashFunctions.h
```diff
@@ -8,7 +8,7 @@
 
 static void ToHashStr (char* dst, const unsigned char* src, size_t src_len) {
   for (size_t i = 0; i < src_len; ++i) {
-    sprintf (&dst[i * 2], "%.2x", src[i]);
+    snprintf (&dst[i * 2], 3, "%.2x", src[i]);
```
mstorsjo, line 11: This change from `sprintf` to `snprintf` seems a bit unrelated here - or what warning did it fix (and if this change was to fix a warning, there's plenty of more `sprintf` in the codebase).

The reason why I'm asking is that I'm running a couple continuous test builds with older MSVC versions (2010 and 2013) which have run fine so far, but fail now. MSVC versions before 2015 didn't have `snprintf`; within the main codec code we have `WelsSnprintf` as a wrapper, and within the test code, we have mostly used `sprintf` so far.

There's probably not much practical reason to keep using so old MSVC versions any more, but they used to work and I've kept testing them - and now they no longer work.

If this is irrelevant, I can of course bump those test configurations to only test with newer MSVC versions instead.

PR #3598, opened 2022-12-08
Improve checking for GCC >= 8 and Clang for passing -Wno-class-memaccess

mstorsjo:
Don't pass -Wno-class-memaccess to Clang, which doesn't support that option name. (When inspecting $(CXX) with a pattern, "clang++" also matches the pattern "%g++".)

Simplify the pattern checking; use $(filter %g++,$(CXX)) instead of $(patsubst %g++,,$(CXX)) making it less obscure. (This requires changing the ifeq ($(),) into ifneq.)

Check for %clang++ in addition to %g++.

Also use a similar pattern for checking for clang++ instead of plain exact matches for "clang++", for passing the option -Wc++11-compat-reserved-user-defined-literal - this allows applying the option when CXX contains a path, or if "clang++" is prefixed by a cross triple.

### DIFF
--- a/build/platform-gnu-chain.mk
+++ b/build/platform-gnu-chain.mk
@@ -28,13 +28,15 @@ CFLAGS += -march=armv8-a
 ASMFLAGS += -march=armv8-a
 endif
 
-ifeq ($(CXX), clang++)
+ifneq ($(filter %clang++,$(CXX)),)
 CXXFLAGS += -Wc++11-compat-reserved-user-defined-literal
 endif
 
-ifeq ($(patsubst %g++,,$(CXX)),)
+ifneq ($(filter %g++,$(CXX)),)
+ifeq ($(filter %clang++,$(CXX)),)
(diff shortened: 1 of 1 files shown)


GuangweiWang:
LGTM

mstorsjo:
All the failures in CI are in `ThreadDecodeFile/ThreadDecoderOutputTest.CompareOutput/*`, which fails occasionally independent of these changes.

I'm considering if we should entirely ignore the `ThreadDecodFile` tests in the CI setup, since they're not passing deterministically. That would lose test coverage of them (instead of being mostly-working but failing occasionally, they could break entirely), but currently they're causing a lot of noise.

(Also, why can't the threaded decoding be fixed so that it deterministically always produces the correct reference output, instead of randomly producing different results?)

huili2:
@xiaotiansf now this series of UT greatly breaks our CI. Do you think it's easy to fix or we'd disable them first?

xiaotiansf:
I think it is safer to disable it in UT for now until I find time to fix it.

mstorsjo:
Now after rebasing past 8535525f4e0f423de054a4afba844db5961d24ed, CI is passing cleanly.

Issue #3416, opened 2021-09-01
Dangling commits on v2.1.0 branch


ade:
There are dangling commits on branch: **openh264v2.1.0** that are not included in neither of 2.1.0 or 2.1.1 releases.

For example, the code in https://github.com/cisco/openh264/pull/3261

Please merge this code properly to releases

kanzhu-cisco:
From history log, the #3261 already merged to cisco:openh264v2.1.0

GuangweiWang:
it should be on master now. 
The master contains all the commits

act262:
@GuangweiWang this commit not merge into master, is any idea?

GuangweiWang:
PR #3599 will merge PR #3261 to master.

GuangweiWang:
has been meerged now.

PR #3261, opened 2020-04-05
easy build with Android SDK circa 2020 and NDK r20 (also r21)

alexcohn:
switch from ant to gradle
look at the commonly defined ANDROID_HOME and ANDROID_NDK_ROOT env variables
don't need $(ANDORID_HOME)/tools to be on PATH

This should answer https://github.com/cisco/openh264/issues/3051 https://github.com/cisco/openh264/issues/3069 https://github.com/cisco/openh264/issues/3107 https://github.com/cisco/openh264/issues/3234 https://github.com/cisco/openh264/issues/3250 https://github.com/cisco/openh264/issues/3251

PR #3599, opened 2022-12-09
pick up PR#3261: switch from ant to gradle

GuangweiWang:
for issue #3416
as the title. pick up PR #3261. 
changes:
1) switch from ant to gradle
2) use clang as the default toolchain


GuangweiWang:
hi @mstorsjo 
I changed the defalut compiler from GCC to clang for Android platform compiling. do you think it will break your build?

mstorsjo:
> hi @mstorsjo
> I changed the defalut compiler from GCC to clang for Android platform compiling. do you think it will break your build?

Thanks for the headsup! Maybe it will, but it’s probably fine anyway, I can update my builds to a newer NDK if necessary.

Issue #3317, opened 2020-07-26
SEncoderStatistics.uiAverageFrameQP always return 0 when get the encoder stats


mypopydev:
SEncoderStatistics.uiAverageFrameQP  always return 0 when get the encoder stats with option ENCODER_OPTION_GET_STATISTICS

huili2:
any detailed case for this issue? Actually If the encoder is called properly, and encoding is just in the process, the return value should be true.

mypopydev:
I think uiAverageFrameQP means  Average Frame QP base on Frame, but openh264enc just return 0 now

huili2:
your guess is right. But it is strange that it returns 0. Any more details on how it occurs?

Allen1207:
that because the uiAverageFrameQP is not assigned when GetOption ENCODER_OPTION_GET_STATISTICS. You can add the following bold line to solve it:
`case ENCODER_OPTION_GET_STATISTICS: {
    SEncoderStatistics* pStatistics = (static_cast<SEncoderStatistics*> (pOption));
    SEncoderStatistics* pEncStatistics = &m_pEncContext->sEncoderStatistics[m_pEncContext->pSvcParam->iSpatialLayerNum - 1];
    pStatistics->uiWidth = pEncStatistics->uiWidth;
    pStatistics->uiHeight = pEncStatistics->uiHeight;
    pStatistics->fAverageFrameSpeedInMs = pEncStatistics->fAverageFrameSpeedInMs;
    pStatistics->sEncMv = pEncStatistics->sEncMv;

    // rate control related
    pStatistics->fAverageFrameRate = pEncStatistics->fAverageFrameRate;
    pStatistics->fLatestFrameRate = pEncStatistics->fLatestFrameRate;
    pStatistics->uiBitRate = pEncStatistics->uiBitRate;
    **pStatistics->uiAverageFrameQP = pEncStatistics->uiAverageFrameQP;**

    pStatistics->uiInputFrameCount = pEncStatistics->uiInputFrameCount;
    pStatistics->uiSkippedFrameCount = pEncStatistics->uiSkippedFrameCount;

    pStatistics->uiResolutionChangeTimes = pEncStatistics->uiResolutionChangeTimes;
    pStatistics->uiIDRReqNum = pEncStatistics->uiIDRReqNum;
    pStatistics->uiIDRSentNum = pEncStatistics->uiIDRSentNum;
[...]

PR #3603, opened 2022-12-16
update enc metrics QP

huili2:
to fix https://github.com/cisco/openh264/issues/3317


jrmuizel:
Can this have a test?

huili2:
@jrmuizel do you find any wrong info about this PR?

jrmuizel:
Nope, just wondering if it makes sense for there to be a test that checks if the metrics are properly filled out.

huili2:
I thought it would be correct, but not sure. Let me double check.

huili2:
Checked the code, it should be OK (stands for the last frame's average qp).

PR #3608, opened 2022-12-22
Use WelsSnprintf instead of direct snprintf in tests

mstorsjo:
Versions of MSVC before 2015 lack snprintf; within the codec codebase itself, this is wrapped into WelsSnprintf. Within the testcases, there were no uses of snprintf previously, only sprintf.

We could of course decide to stop caring about older versions of MSVC, but so far, those versions have still worked just fine. However, that requires changes to the build files for Windows Phone ("make OS=msvc-wp"), which currently expect to be built with MSVC 2013.

Define NOMINMAX before including windows.h, to avoid windows.h breaking std::min in test code.

### DIFF
--- a/codec/common/inc/crt_util_safe_x.h
+++ b/codec/common/inc/crt_util_safe_x.h
@@ -48,6 +48,7 @@
 #include <time.h>
 
 #if defined(_WIN32)
+#define NOMINMAX
 #include <windows.h>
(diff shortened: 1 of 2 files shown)


GuangweiWang:
LGTM

mstorsjo:
Thanks!

PR #3612, opened 2022-12-30
Switch building of the unittest from Ant to Gradle

mstorsjo:
This was missed in b16f0c23213237c7957f47d605df4e2f4c4357a1 where the other targets were switched from Ant to Gradle.

This fixes building after switching to a newer version of the Android SDK (suppporting Gradle) where Ant no longer is supported.

CC @alexcohn


GuangweiWang:
thanks!

PR #3618, opened 2023-02-08
Avoid providing a warning if the requested number of threads is the current number of threads

hmaarrfk:
It seems that this thread number parameter is always defined.

When starting multiple contexts, it is seem hard to not specify this.

Therefore, you end up with many warnings.

Thank you for considering

### DIFF
--- a/codec/encoder/core/src/wels_task_management.cpp
+++ b/codec/encoder/core/src/wels_task_management.cpp
@@ -102,8 +102,12 @@ WelsErrorType CWelsTaskManageBase::Init (sWelsEncCtx* pEncCtx) {
   iReturn = CWelsThreadPool::SetThreadNum (m_iThreadNum);
   m_pThreadPool = (CWelsThreadPool::AddReference());
   if ((iReturn != ENC_RETURN_SUCCESS) && pEncCtx) {
-    WelsLog (& (pEncCtx->sLogCtx), WELS_LOG_WARNING, "Set Thread Num to %d did not succeed, current thread num in use: %d",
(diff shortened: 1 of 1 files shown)


huili2:
@hmaarrfk This thread initialization should be done only once, and from the code, it seems OK if the return code is NULL (meaning thread pool initialization failed). So In what case would this happen?

hmaarrfk:
I can see a few reasons:

1. In live applications (i.e. where you can't spawn a new process and start a totally new global context), you may want to tune the number of threads as the desired resolution changes. This can be when a user is changing the resolution of the output video due to encoding a totally different stream, or when changing the resolution of the broadcasting resolution in a streaming sense.
2. In applications where you are writing multiple videos to disk, you may also want to tune the number of threads to a particular need. Do you have many other steps in your pipeline happening in parallel? Choose "1" thread? If you have fewer expected background processing, you can increase this to "4".

I think FFMPEG will always set this:
https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/libopenh264enc.c#L176

So even if you serially encode, multiple different 1080 p streams in a single application (using libavcodec directly in your program instead of the ffmpeg executable), you will start to see this popup many times.

hmaarrfk:
I was somewhat surprised that the threadpool is global to all different encoding contexts, instead of being local to each encoding context.

I would like for example, to start 5 encoding contexts, and limit each to 2 threads. I couldn't find a good way to do this.

hmaarrfk:
Thank you@!

PR #3630, opened 2023-03-02
Use environment for mips feature detection

matoro:
The -march= option is perfectly happy to emit code to run on a processor different than the one on which it is being compiled.  This results in misdetection of mips features because the test compiles specify that a given extension should be emitted, but this does not check whether or not this corresponds to the subarchitecture targeted in CFLAGS by the rest of the build.

```
$ echo "void main(void){ __asm__ volatile(\"punpcklhw \$f0, \$f0, \$f0\"); }" > test.c $ CFLAGS="-march=loongson3a" make test
cc -march=loongson3a    test.c   -o test
$ ./test
Illegal instruction
$ CFLAGS="-march=native" make -B test
cc -march=native    test.c   -o test
/tmp/ccLbeyM1.s: Assembler messages:
/tmp/ccLbeyM1.s:25: Error: opcode not supported on this processor: octeon2 (mips64r2) `punpcklhw $f0,$f0,$f0'
make: *** [<builtin>: test] Error 1
```

This leads to -march=loongson3a getting appended to CFLAGS, which may conflict with previously specified -march= levels for the build, or other options.  Calling make in the test will use whatever CC/CFLAGS are specified in the environment to determine whether the actual compile command line to be used in the build supports these features.

### DIFF
--- a/build/arch.mk
+++ b/build/arch.mk
@@ -39,14 +39,14 @@ ASM_ARCH = mips
 ASMFLAGS += -I$(SRC_PATH)codec/common/mips/
 #mmi
 ifeq ($(ENABLE_MMI), Yes)
-ENABLE_MMI = $(shell $(SRC_PATH)build/mips-simd-check.sh $(CC) mmi)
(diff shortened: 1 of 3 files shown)


LodewijkIVX:
;-)

Issue #3350, opened 2020-10-19
Seeking in stream works with IDR frames only


stransky:
When OpenH264 is used for youtube playback in Firefox and the stream position is moved, decoding fails if it starts on non IDR I-frame. When decoding starts on IDR frame the stream is decoded correctly.

Mozilla bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1670333

huili2:
Sorry for too late response. Openh264 does not support I-frame decoding as IDR.

Conan-Kudo:
Could support for this be added to OpenH264?

PR #3647, opened 2023-04-12
Add support for exporting the decoder timestamp in the GMP interface.

aosmond:
This resolves frame jank in Firefox by using the correct timestamps for playback.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1827703 for the changes on the Gecko side.

### DIFF
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -625,6 +625,7 @@ class OpenH264VideoDecoder : public GMPVideoDecoder, public RefCounted {
     callback_ (nullptr),
     decoder_ (nullptr),
     stats_ ("Decoder"),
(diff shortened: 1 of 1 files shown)


aosmond:
I found a bug with this, as seeking appears to be broken with Firefox. Likely this is on the Gecko side of changes.

Darkspirit:
#3350 = https://bugzilla.mozilla.org/show_bug.cgi?id=1670333 are about broken seeking.

aosmond:
Actually, turns out we have all of the information we need already taken from our media pipeline, and using the timestamp from OpenH264 is counter productive since it continues going up after a seek (which we don't expect but I can see an application desiring that).

PR #3648, opened 2023-04-14
Bump the API version to add support for new requirements from Firefox.

aosmond:
Export the updated presentation timestamp from the decoder. This resolves frame jank in Firefox with B frames by displaying frames in order, without being dropped due to being too old for display.

Implement Reset to flush pending frames from the decoder. This resolves seeking in Firefox with the updated presentation timestamps, where old frames would get processed and interfere with the seek.


aosmond:
It also lets us access multi-threaded decoding and enabling logging.

aosmond:
We updated the gmp-api repository which should resolve the build issues:
https://github.com/mozilla/gmp-api/releases/tag/Firefox114

aosmond:
Made a small change to the build system to add "-std=c++17" for OSX builds to make it compatible with our GMP API

aosmond:
When you merge this, could you please cherry pick this onto 2.3.1 to release 2.3.2?

aosmond:
Or whatever version bump this change would merit.

PR #3707, opened 2023-12-06
Fix glitches that sometimes happen near the scene changes.

tyan0:
Since the first time I used the openh264 decoder, I have encountered glitches that sometimes happen near the scene changes. I looked into this problem and found two causes.

1. Lack of necessary code which sets m_sReoderingStatus.iLastWrittenPOC.
2. Overwrite frame buffer occasionally happens at GOP change.

This issue happens frequently enough to frustrate me. I wonder why the other users have not pointed this problem out till today.

This pull request provides two commits corresponding to each problem.

### DIFF
--- a/codec/decoder/core/inc/decoder_context.h
+++ b/codec/decoder/core/inc/decoder_context.h
@@ -285,14 +285,15 @@ typedef struct tagPictInfo {
   int32_t                 iPOC;
   int32_t                 iPicBuffIdx;
   uint32_t                uiDecodingTimeStamp;
-  bool                    bLastGOP;
+  int32_t                 iSeqNum;
 } SPictInfo, *PPictInfo;
 
 typedef struct tagPictReoderingStatus {
   int32_t iPictInfoIndex;
+  int32_t iMinSeqNum;
   int32_t iMinPOC;
   int32_t iNumOfPicts;
-  int32_t iLastGOPRemainPicts;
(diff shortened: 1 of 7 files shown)


tyan0:
If it is better to provide sample video file with the glitch, I am ready to do that.

BenzhengZhang:
@tyan0 CI failed.

tyan0:
Thanks for checking. I have fixed that problem. Could you please rerun CI?

tyan0:
I revised the pull request several times. Now I am almost sure that this is the final change. Please review. Thanks.

tyan0:
I revised the pull request several times. Now I am almost sure that this is the final change. Please review. Thanks.

tyan0:
I revised the pull request several times. Now I am almost sure that this is the final change. Please review. Thanks.

tyan0:
I found another problem of the current openh264 code regarding scene change.

3. When more than 2 frame sequences exist in the reordering buffer, the reordering does not work correctly.

This causes when multiple scene canges happen in a very short time. I would like to add new commit for that problem.

tyan0:
Done. Please review.

tyan0:
Fix typo in commit message.

BenzhengZhang:
Thanks for the fix @tyan0

tyan0:
Thanks!

PR #3716, opened 2023-12-20
fix cross compile from darwin arm64 to x86_64

### LINKED ISSUE #3715: Cross compiling from an macOS ARM64 machine to x86_64 fails
I just tried updating from 1.8.0 to 2.4.0 and noticed that the makefile is set up in a way that makes it impossible to cross-compile from an ARM64 machine (M2 Ultra in this case) to Intel x86_64.

Make invocation is like this:
`make OS=darwin ARCH=x86_64`

But `-arch x86_64` is not passed to `CFLAGS` and `LDFLAGS` so the build fails.
I assume that the reverse (cross-compiling from an Intel machine to ARM64) will work but haven't tried it.

zhuzeitou:
fix #3715

### DIFF
--- a/build/platform-darwin.mk
+++ b/build/platform-darwin.mk
@@ -23,8 +23,12 @@ endif
 ifeq ($(ASM_ARCH), x86)
 ASMFLAGS += -DPREFIX
 ifeq ($(ARCH), x86_64)
+CFLAGS += -arch x86_64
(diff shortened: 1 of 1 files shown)


lemonmojo:
Why hasn't this been merged yet? In 2.4.1 it's still impossible to cross-compile from ARM64 to x86_64.

PR #3721, opened 2024-01-19
Fix frame-decode errors.

tyan0:
This pull request is for both https://github.com/cisco/openh264/issues/3710 and https://github.com/cisco/openh264/issues/3713.
I am not very sure that these are the correct fixes even though the frame decode errors are indeed resolved, so please review them. Especially for https://github.com/cisco/openh264/issues/3713.

Issue #3713, opened 2023-12-10
Another frame decode error.


tyan0:
Openh264 fails to decode mp4 file:
https://tyan0.yr32.net/openh264/decode_frame_error2.mp4
with error.

I am not sure this is the right thing, however, the following patch solves the issue.

```
diff --git a/codec/decoder/core/src/decoder.cpp b/codec/decoder/core/src/decoder.cpp
index dba6711b..c6b6bddc 100644
--- a/codec/decoder/core/src/decoder.cpp
+++ b/codec/decoder/core/src/decoder.cpp
@@ -796,8 +796,9 @@ int32_t WelsDecodeBs (PWelsDecoderContext pCtx, const uint8_t* kpBsBuf, const in
           bNalStartBytes = true;
         } else if (pSrcNal[2 + iSrcIdx] == 0x03) {
           if ((3 + iSrcConsumed < iSrcLength) && pSrcNal[3 + iSrcIdx] > 0x03) {
-            pCtx->iErrorCode |= dsBitstreamError;
-            return pCtx->iErrorCode;
+            /* Just skip */
+            iSrcIdx      += 3;
+            iSrcConsumed += 3;
           } else {
             ST16 (pDstNal + iDstIdx, 0);
             iDstIdx      += 2;
```

Could maintainer please have a look?

tyan0:
This errror handling was added by the commit https://github.com/cisco/openh264/commit/e90068c234a8f93df01b30b0d3de9ca8d8915c1d.
What situation is this necessary? And does the above code change break this fix?

Issue #3710, opened 2023-12-08
Frame decode error.


tyan0:
Frame decode error occurs if playing the following video file.
[https://tyan0.yr32.net/openh264/decode_frame_error.mp4](https://tyan0.yr32.net/openh264/decode_frame_error.mp4)

I am not sure this is the right thing, but the following patch solves the issue.
```
diff --git a/codec/decoder/core/src/manage_dec_ref.cpp b/codec/decoder/core/src/manage_dec_ref.cpp
index c3c98312..76c7ed43 100644
--- a/codec/decoder/core/src/manage_dec_ref.cpp
+++ b/codec/decoder/core/src/manage_dec_ref.cpp
@@ -460,7 +460,7 @@ int32_t WelsReorderRefList (PWelsDecoderContext pCtx) {
           }
         }
         if (i < 0) {
-          return ERR_INFO_REFERENCE_PIC_LOST;
+          continue;
         }
         pPic = ppRefList[i];
         if (i > iReorderingIndex) {
```

Could maintainer please have a look?

tyan0:
Sorry, I noticed above patch causes test failure. I'd like to propose alternative fix below.
New patch enables the function WelsReorderRefList2() which was introduced by the commit 1b8caef6ccc7, and fixes trivial issue of that function. The two patches are completely different, but, both solve the issue.

```
diff --git a/codec/decoder/core/src/decoder_core.cpp b/codec/decoder/core/src/decoder_core.cpp
index 106d0d41..48e03a0d 100644
--- a/codec/decoder/core/src/decoder_core.cpp
+++ b/codec/decoder/core/src/decoder_core.cpp
@@ -2432,7 +2432,7 @@ int32_t InitRefPicList (PWelsDecoderContext pCtx, const uint8_t kuiNRi, int32_t
   } else
     iRet = WelsInitRefList (pCtx, iPoc);
   if ((pCtx->eSliceType != I_SLICE && pCtx->eSliceType != SI_SLICE)) {
-#if 0
+#if 1
     if (pCtx->pSps->uiProfileIdc != 66 && pCtx->pPps->bEntropyCodingModeFlag)
       iRet = WelsReorderRefList2 (pCtx);
     else
diff --git a/codec/decoder/core/src/manage_dec_ref.cpp b/codec/decoder/core/src/manage_dec_ref.cpp
index 410b9c47..fe3e033a 100644
--- a/codec/decoder/core/src/manage_dec_ref.cpp
[...]

PR #3723, opened 2024-01-24
Fix frame-decode errors.

tyan0:
** This PR is re-submission of PR (#3721) to fix rebase issue. **

This pull request is for both https://github.com/cisco/openh264/issues/3710 and https://github.com/cisco/openh264/issues/3713.
I am not very sure that these are the correct fixes even though the frame decode errors are indeed resolved, so please review them. Especially for https://github.com/cisco/openh264/issues/3713.


joakim-tjernlund:
@BenzhengZhang ping?

BenzhengZhang:
@tyan0 Is there a risk of regression?

tyan0:
I cannot say that there is never a risk of regression, however, as far as I tested (with many of my video archives and the test cases of which recently reported) I have not found any regression so far.

joakim-tjernlund:
> @tyan0 Is there a risk of regression?

At this point I guess it would be best to merge it to master to find out.

tyan0:
Fixed a bug reported in https://github.com/cisco/openh264/pull/3772

PR #3724, opened 2024-01-29
update openh264 version to v2.4.1

BenzhengZhang:
as the title.


joakim-tjernlund:
There is no Tag/Release for 2.4.1

PR #3752, opened 2024-05-10
Change picOrderCnt to 2

fippo:
changes the picOrderCnt field to 2 (which means the output order
is the same as the decode order) which is more suited to realtime
usage and allows the decoder to immediately decode a frame without
waiting for out-of-order frames. This reduces the decoding latency
on Qualcomm hardware decoders which do not honor the VUI parameter
  max_num_reorder_frames


fippo:
Looks like the ASAN test is running some tests that are not run by the normal test suite, see the commit titled "bump asan hashes"? And fixing the hashes for ASAN breaks them for the normal tests locally (on Linux)?! Reverting and `make clean` fixed this so this seems to be a minor issue with the test build deps; as long as the CI catches it all is well.

BenzhengZhang:
@fippo Could you pls submit a new PR and I will start CI.

fippo:
@BenzhengZhang now that I have had a change merged github considers me to be a friendly person so starts the CI automatically! Builds passed :tada:, would you like me to squash all the hash updates?

PR #3773, opened 2024-08-11
build: fix build with meson on riscv64

ziyao233:
RISC-V is a rising open-standarded RISC ISA. This commit makes mesonbuild fallbacks to C implementation correctly on riscv64 platform instead of throwing a "FIXME: unhandled CPU family" error.


ziyao233:
Another option is using C fallbacks for all unrecognized architectures. I'm not sure which is better, thus waiting for your comment.

PR #3782, opened 2024-08-30
Fix broken decode result in multi-thread decoding.

tyan0:
If the following test video is decoded with multi-thread by openh264, the decode result will be broken.

https://github.com/user-attachments/assets/b3fbfc09-f3a3-468c-bbc7-d761ae7fdd8c

The following file is the re-encoded result of the broken decoded result.

https://github.com/user-attachments/assets/ed5de14d-c78c-4a92-a0fc-82c5f32f119f

This patch just removes
```
          if (iThreadCount > 1) isNewFrame = false;
```
which was introduced by the commit 54f6c4ddbd771e3af64d763f1551ed22655102f2, but it is not clear why this was necessary.


joakim-tjernlund:
Finding out why that line was added may be hard.
Just merge to master and see if any issues are reported ?

joakim-tjernlund:
FYI, I have added this patch to our openh264(on top of master) and not seen/heard any problems.

PR #3790, opened 2024-09-24
Motion Vector Extraction from H.264 Encoded Bitstream (CABAC Only)

mohammedzakikochargi:
This PR introduces a feature that allows the extraction of motion vectors from H.264 encoded bitstreams, specifically for streams using CABAC (Context-Adaptive Binary Arithmetic Coding), without performing full frame reconstruction. By focusing only on CABAC, this optimization reduces the computational overhead for decoding when motion vectors are the primary data of interest.

Key benefits:

CABAC-specific optimization: Reduces decoding time by bypassing frame reconstruction in CABAC-encoded streams, which typically consume the most computational resources.
Efficient for motion analysis: Ideal for use cases requiring only motion vector data, such as motion tracking and real-time analysis.
Streamlined decoding: Maintains the integrity of the bitstream while providing an efficient way to access motion vectors.
This enhancement is especially valuable for performance-critical applications where CABAC encoding is used, and full frame reconstruction is unnecessary.

### DIFF
--- a/codec/api/wels/codec_api.h
+++ b/codec/api/wels/codec_api.h
@@ -45,6 +45,7 @@ typedef unsigned char bool;
 
 #include "codec_app_def.h"
 #include "codec_def.h"
+#include <stdint.h>
 
 #if defined(_WIN32) || defined(__cdecl)
 #define EXTAPI __cdecl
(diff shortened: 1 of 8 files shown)


joakim-tjernlund:
Since current master break ABI anyway, I guess it would be good to have this merged as well.

PR #3793, opened 2024-10-18
Remove missingFrames arg from GMP API

Pehrsons:
It is being removed from the interface in https://github.com/mozilla/gmp-api/pull/24.


Pehrsons:
Not planning to do this anymore.

PR #3824, opened 2024-12-13
Allow calculating PSNR for Y/U/V components

fippo:
by requesting this either as a option for all frames in
  SEncParamExt.bPsnrY (and U/V)
or per-frame using
  SSourcePicture.bPsnrY (and U/V)

The resulting data goes into
  SLayerBSInfo.rPsnr
with is a three-element array for the Y/U/V components of the PSNR.

This is disabled by default. Also removes the ENABLE_PSNR_CALC preprocessor define in favor of the API.

See
  https://www.researchgate.net/publication/383545049_Low-Complexity_Video_PSNR_Measurement_in_Real-Time_Communication_Products
by @YCSun-Meta for a research paper explaining the background.

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -2186,7 +2186,7 @@ void StatOverallEncodingExt (sWelsEncCtx* pCtx) {
 
   }
 }
-#endif
+#endif //#if defined(STAT_OUTPUT)
 
 
 int32_t GetMultipleThreadIdc (SLogContext* pLogCtx, SWelsSvcCodingParam* pCodingParam, int16_t& iSliceNum,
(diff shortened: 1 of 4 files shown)


BenzhengZhang:
@fippo Have you done any performance testing?

YCSun-Meta:
Hi @BenzhengZhang ,

We did performance test on a mid-range device (Pixel 4 phone), and the psnr calculation takes ~350us. 
I attached the research paper, which contains the detailed numbers. 
[ICME2024_Low-ComplexityVideoQualityMeasurementinReal-timeCommunicationProducts_v13_0526.pdf](https://github.com/user-attachments/files/18254540/ICME2024_Low-ComplexityVideoQualityMeasurementinReal-timeCommunicationProducts_v13_0526.pdf)

In addition, the patch also provides frame-level control.
If a app has concern on the additional ~350us computation, it can do the temporal subsampling of video frames for PSNR calculation or disable the feature at all.  

Meta have tested and enabled the feature on the RTC apps on all devices. Its reliability as a metric at scale has been
validated. 
We also had discussion with Google, they think that the numbers are acceptable and will enable this feature in Google WebRTC implementation as well.

joakim-tjernlund:
if this is an API change, should you not increase .so version as well?

BenzhengZhang:
> if this is an API change, should you not increase .so version as well?

Should be increased in next release.

codec/encoder/core/src/encoder_ext.cpp
```diff
@@ -3956,6 +3956,19 @@ int32_t WelsEncoderEncodeExt (sWelsEncCtx* pCtx, SFrameBSInfo* pFbi, const SSour
              (iLayerSize << 3));
 #endif//LAYER_INFO_OUTPUT
 
+    pLayerBsInfo->rPsnr[0] = NAN;
```
mstorsjo, line 3959: This change broke compiling with MSVC 2010 and 2012; they lack the `NAN` constant in `math.h`.

Not sure if MSVC 2012 really is within the scope of targets we need to support - but it did compile fine before this.

mstorsjo: Or put another way - does this strictly need to be `NAN`, or could we initialize it to some other trivial value, like 0 or -1 or something like that?

BenzhengZhang: @fippo  As @mstorsjo mentioned, this patch broke the build for older versions of MSVC. Could you look into making it compatible with those versions?

fippo: I think setting it to 0.0 should work, will do a PR

PR #3836, opened 2025-01-24
psnr: do not use NAN, initialize to 0

fippo:
which is not supported in some older compilers as pointed out in #3824 Applications should always check whether they requested calculating PSNR before evaluating that field in the result.


fippo:
@BenzhengZhang PTAL

BenzhengZhang (approved):

PR #3894, opened 2025-09-06
Enable access to reconstructed picture output in the encoder

marecki315:
Access to the reconstructed frame from the H.264 encoder is useful in scenarios where the application layer needs to perform further processing or analysis on the encoded video. By exposing the reconstructed picture, the encoder facilitates integration with advanced video processing workflows and enables more flexible application-level control.


freitaschega2368-rgb:
https://go.wps.com/a8M4ofQQ8WW2

PR #3905, opened 2025-10-14
guard aarch64 assembly with __aarch64__

akallabeth:
The symbol HAVE_NEON_AARCH64 is defined by the build system and not a full compile time check. When building multiarch binaries (apple universal) then also check for the compiler defined symbol __aarch64__ to ensure we are in the arm64 compiler block


BenzhengZhang:
@akallabeth  could you merge master to include commit:a0175163e430c2e6f9f36c7f89e57817faa47e12 to your branch in order to pass windows CI.

akallabeth:
@BenzhengZhang done.

PR #3924, opened 2026-01-14
Fix NEON argument in ARM configuration

thomasdangl:
This fixes a copy-paste error for ARM builds where NEON is enabled twice for C instead of C and C++


philipbry:
LGTM

PR #3928, opened 2026-02-11
Improve support for CYGWIN

carlo-bramini:
I tried to build openh264 on CYGWIN but I got few errors that can be fixed quite easily, hopefully.

1)
The platform "cygwin" is not recognized by the script for Meson and this message is printed on the console.
```
meson.build:143:2: ERROR: Problem encountered: FIXME: Unhandled system cygwin

A full log can be found at meson-logs/meson-log.txt
```
However, CYGWIN can be handled by adding its name in the same path where MinGW is also handled.

2)
Later, this error appears on the console:
```
openh264/codec/decoder/plus/src/welsDecoderExt.cpp: In function ‘DWORD WelsDec::pThrProcInit(LPVOID)’:
openh264/codec/decoder/plus/src/welsDecoderExt.cpp:94:3: error: ‘_alloca’ was not declared in this scope; did you mean ‘alloca’?
   94 |   _alloca (WELS_DEC_MAX_THREAD_STACK_SIZE * (sThreadInfo->uiThrNum + 1));
      |   ^~~~~~~
      |   alloca
```
Similar to other POSIX platforms, CYGWIN supports `alloca()` rather than `_alloca()`, so this function needs to be used for reserving stack space.

3)
Finally, this last error happens:
```
openh264/codec/decoder/core/src/wels_decoder_thread.cpp: In function ‘void SemRelease(SWelsDecSemphore*, long int*)’:
openh264/codec/decoder/core/src/wels_decoder_thread.cpp:118:30: error: cannot convert ‘long int*’ to ‘LPLONG’ {aka ‘int*’}
  118 |   ReleaseSemaphore (s->h, 1, prevcount);
      |                              ^~~~~~~~~
      |                              |
      |                              long int*
```
On x86_64 and aarch64 ports, `long` has a size of 8 byte, while `LONG` is still 4 bytes even when `_WIN64` is defined with Windows APIs.
This can be fixed by using a local parameter declared as `LONG` and, in my opinion, it would be worth to do it also for traditional `_WIN32` platforms for a clear type matching.

I hope that you will find these tiny patches useful.

### DIFF
--- a/codec/decoder/core/src/wels_decoder_thread.cpp
+++ b/codec/decoder/core/src/wels_decoder_thread.cpp
@@ -115,7 +115,10 @@ int SemWait (SWelsDecSemphore* s, int32_t timeout) {
 }
 
 void SemRelease (SWelsDecSemphore* s, long* prevcount) {
-  ReleaseSemaphore (s->h, 1, prevcount);
+  LONG _prevcount;
+
+  ReleaseSemaphore (s->h, 1, &_prevcount);
+  *prevcount = _prevcount;
 }
 
 void SemDestroy (SWelsDecSemphore* s) {
--- a/codec/decoder/plus/src/welsDecoderExt.cpp
+++ b/codec/decoder/plus/src/welsDecoderExt.cpp
@@ -90,7 +90,7 @@ namespace WelsDec {
 ***************************************************************************/
 DECLARE_PROCTHREAD (pThrProcInit, p) {
   SWelsDecThreadInfo* sThreadInfo = (SWelsDecThreadInfo*)p;
-#if defined(WIN32)
+#if defined(WIN32) && !defined(__CYGWIN__)
   _alloca (WELS_DEC_MAX_THREAD_STACK_SIZE * (sThreadInfo->uiThrNum + 1));
 #endif
   return sThreadInfo->pThrProcMain (p);
--- a/codec/decoder/plus/src/welsDecoderExt.cpp
+++ b/codec/decoder/plus/src/welsDecoderExt.cpp
@@ -90,7 +90,7 @@ namespace WelsDec {
 ***************************************************************************/
 DECLARE_PROCTHREAD (pThrProcInit, p) {
   SWelsDecThreadInfo* sThreadInfo = (SWelsDecThreadInfo*)p;
-#if defined(WIN32)
+#if defined(_WIN32)
   _alloca (WELS_DEC_MAX_THREAD_STACK_SIZE * (sThreadInfo->uiThrNum + 1));
 #endif
   return sThreadInfo->pThrProcMain (p);
--- a/meson.build
+++ b/meson.build
@@ -112,7 +112,7 @@ if ['linux', 'freebsd', 'android', 'ios', 'darwin'].contains(system)
   if ['ios', 'darwin', 'android'].contains(system)
     cpp_lib = '-lc++'
   endif
-elif system == 'windows'
+elif ['windows', 'cygwin'].contains(system)
   if cpu_family == 'x86'
     asm_format = 'win32'
     asm_args += ['-DPREFIX', '-DX86_32']


tyan0:
> The platform "cygwin" is not recognized by the script for Meson and this message is printed on the console.

The patch https://github.com/cisco/openh264/pull/3928/commits/244c84e4c2ce7a0458a2ff0fd17a53e1f9c878df looks good to me.

> Later, this error appears on the console:
> 
> ```
> openh264/codec/decoder/plus/src/welsDecoderExt.cpp: In function ‘DWORD WelsDec::pThrProcInit(LPVOID)’:
> openh264/codec/decoder/plus/src/welsDecoderExt.cpp:94:3: error: ‘_alloca’ was not declared in this scope; did you mean ‘alloca’?
>    94 |   _alloca (WELS_DEC_MAX_THREAD_STACK_SIZE * (sThreadInfo->uiThrNum + 1));
>       |   ^~~~~~~
>       |   alloca
> ```
> 
> Similar to other POSIX platforms, CYGWIN supports `alloca()` rather than `_alloca()`, so this function needs to be used for reserving stack space.

`_alloca()` causes exception (SEH) if stack is not enough. So the intent of the code is check the stack space for multi-thread operation and if not, raise an exception. I'm not sure why the exception here is the right thing for WIN32 platform,  however, the behavirour of `alloca()` when the stack is not enough is undefined unlike `_alloca()`.
So, I think the following patch would be better:
``` patch
--- b/codec/decoder/plus/src/welsDecoderExt.cpp
+++ a/codec/decoder/plus/src/welsDecoderExt.cpp
@@ -90,7 +90,7 @@ namespace WelsDec {
 ***************************************************************************/
 DECLARE_PROCTHREAD (pThrProcInit, p) {
   SWelsDecThreadInfo* sThreadInfo = (SWelsDecThreadInfo*)p;
-#if defined(WIN32)
+#if defined(WIN32) && !defined(__CYGWIN__)
   _alloca (WELS_DEC_MAX_THREAD_STACK_SIZE * (sThreadInfo->uiThrNum + 1));
 #endif
   return sThreadInfo->pThrProcMain (p);
```
> Finally, this last error happens:
> 
> ```
> openh264/codec/decoder/core/src/wels_decoder_thread.cpp: In function ‘void SemRelease(SWelsDecSemphore*, long int*)’:
> openh264/codec/decoder/core/src/wels_decoder_thread.cpp:118:30: error: cannot convert ‘long int*’ to ‘LPLONG’ {aka ‘int*’}
>   118 |   ReleaseSemaphore (s->h, 1, prevcount);
>       |                              ^~~~~~~~~
>       |                              |
>       |                              long int*
> ```
> 
> On x86_64 and aarch64 ports, `long` has a size of 8 byte, while `LONG` is still 4 bytes even when `_WIN64` is defined with Windows APIs. This can be fixed by using a local parameter declared as `LONG` and, in my opinion, it would be worth to do it also for traditional `_WIN32` platforms for a clear type matching.

You are right. This (https://github.com/cisco/openh264/pull/3928/commits/a1ae5908e5cab60579fb3c163575c06d46ef1d80) should be the right thing here.

carlo-bramini:
Thank you very much.
I updated the patch to `codec\decoder\plus\src\welsDecoderExt.cpp` according to your request.

tyan0:
BTW, I think using POSIX-like API, rather than WIN32 API, is more correct approach in cygwin, as in the following patch. @carlo-bramini , what do you think?
``` patch
diff --git a/codec/common/inc/WelsThreadLib.h b/codec/common/inc/WelsThreadLib.h
index cd26dbdd..f1693af6 100644
--- a/codec/common/inc/WelsThreadLib.h
+++ b/codec/common/inc/WelsThreadLib.h
@@ -47,7 +47,7 @@
 extern "C" {
 #endif
 
-#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__)
 
 #include <windows.h>
 
diff --git a/codec/common/src/WelsThreadLib.cpp b/codec/common/src/WelsThreadLib.cpp
index 4e44f248..c1e9bac8 100644
--- a/codec/common/src/WelsThreadLib.cpp
+++ b/codec/common/src/WelsThreadLib.cpp
@@ -73,7 +73,7 @@
 #include <stdlib.h>
 
 
-#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__)
 
 WELS_THREAD_ERROR_CODE    WelsMutexInit (WELS_MUTEX*    mutex) {
   InitializeCriticalSection (mutex);
@@ -119,7 +119,7 @@ WELS_THREAD_ERROR_CODE    WelsMutexDestroy (WELS_MUTEX* mutex) {
 
 #endif /* !_WIN32 */
 
-#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__)
 
 WELS_THREAD_ERROR_CODE    WelsEventOpen (WELS_EVENT* event, const char* event_name) {
   WELS_EVENT   h = CreateEvent (NULL, FALSE, FALSE, NULL);
@@ -236,7 +236,7 @@ WELS_THREAD_ERROR_CODE    WelsThreadCreate (WELS_THREAD_HANDLE* thread,  LPWELS_
   err = pthread_attr_init (&at);
   if (err)
     return err;
-#if !defined(__ANDROID__) && !defined(__Fuchsia__)
+#if !defined(__ANDROID__) && !defined(__Fuchsia__) && !defined(__CYGWIN__)
   err = pthread_attr_setscope (&at, PTHREAD_SCOPE_SYSTEM);
   if (err)
     return err;
@@ -515,7 +515,7 @@ WELS_THREAD_ERROR_CODE    WelsQueryLogicalProcessInfo (WelsLogicalProcessInfo* p
   pInfo->ProcessorCount = emscripten_num_logical_cores();
   return WELS_THREAD_ERROR_OK;
 
-#elif defined(__Fuchsia__)
+#elif defined(__Fuchsia__) || defined(__CYGWIN__)
 
   pInfo->ProcessorCount = sysconf(_SC_NPROCESSORS_ONLN);
   return WELS_THREAD_ERROR_OK;
diff --git a/codec/decoder/core/inc/wels_decoder_thread.h b/codec/decoder/core/inc/wels_decoder_thread.h
index c133b59c..6289a764 100644
--- a/codec/decoder/core/inc/wels_decoder_thread.h
+++ b/codec/decoder/core/inc/wels_decoder_thread.h
@@ -52,7 +52,7 @@ extern "C" {
 #define WELS_DEC_THREAD_COMMAND_RUN 0
 #define WELS_DEC_THREAD_COMMAND_ABORT 1
 
-#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__)
 typedef struct tagWelsDecSemphore {
   WELS_THREAD_HANDLE h;
 } SWelsDecSemphore;
diff --git a/codec/decoder/core/src/wels_decoder_thread.cpp b/codec/decoder/core/src/wels_decoder_thread.cpp
index 0b6b7721..b6c897f4 100644
--- a/codec/decoder/core/src/wels_decoder_thread.cpp
+++ b/codec/decoder/core/src/wels_decoder_thread.cpp
@@ -59,7 +59,7 @@ int ThreadWait (SWelsDecThread* t) {
   return WelsThreadJoin (t->h);
 }
 
-#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__)
 
 int EventCreate (SWelsDecEvent* e, int manualReset, int initialState) {
   e->h = CreateEvent (NULL, manualReset, initialState, NULL);
diff --git a/meson.build b/meson.build
index a0dfd219..6c79f184 100644
--- a/meson.build
+++ b/meson.build
@@ -57,7 +57,7 @@ cpp_lib = '-lstdc++'
 libm_dep = cpp.find_library('m', required : false)
 deps += [libm_dep]
 
-if ['linux', 'freebsd', 'android', 'ios', 'darwin'].contains(system)
+if ['linux', 'freebsd', 'android', 'ios', 'darwin', 'cygwin'].contains(system)
   asm_format32 = 'elf'
   asm_format64 = 'elf64'
   if ['ios', 'darwin'].contains(system)
```

carlo-bramini:
Hello,

> BTW, I think using POSIX-like API, rather than WIN32 API, is more correct approach in cygwin, as in the following patch. @carlo-bramini , what do you think?
> 

well, if you prefer to use POSIX APIs instead, there is no need to add a test on `__CYGWIN__` keyword.
On Cygwin, the `_WIN32` macro is not defined, so you can just do this:
```
 -#if defined(_WIN32) || defined(__CYGWIN__)
+#if defined(_WIN32)
```
I don't know the history behind the existing code... probably, the reason of using WIN32 APIs also on Cygwin comes from supporting also the ancient `-no-cygwin` option for building a native Win32 executable with GCC for Cygwin, although nowadays it has been deprecated in favour of the MinGW-w64 cross compilers.
If it's not a problem, you could leave as it is now (it works...), or if you prefer using POSIX APIs you can just remove `__CYGWIN__` from the preprocessor test, I think that's not a big trouble.

Instead, this is wrong:
```
--- a/meson.build
+++ b/meson.build
@@ -57,7 +57,7 @@ cpp_lib = '-lstdc++'
 libm_dep = cpp.find_library('m', required : false)
 deps += [libm_dep]
 
-if ['linux', 'freebsd', 'android', 'ios', 'darwin'].contains(system)
+if ['linux', 'freebsd', 'android', 'ios', 'darwin', 'cygwin'].contains(system)
   asm_format32 = 'elf'
   asm_format64 = 'elf64'
   if ['ios', 'darwin'].contains(system)
```
Cygwin uses COFF/PE, so it doesn't undestand ELF formats.

EDIT:
Into ‎`codec/decoder/plus/src/welsDecoderExt.cpp`, I made this patch:
```
-#if defined(WIN32)
+#if defined(WIN32) && !defined(__CYGWIN__)
```
Because:
- `WIN32` is defined when you include Windows.h. So, it can exists on both Win32 and Cygwin.
- `_WIN32` is intrinsic and it's defined by the compiler. It exists on Win32 but not on Cygwin.

So, here is also another way for fixing it and it can be this one:
```
-#if defined(WIN32)
+#if defined(_WIN32)
```
which looks a bit cleaner.
I updated the patch for it.

PR #3958, opened 2026-07-03
Fix SumOfFrame for extremely wide frames

sprangerik:
When using very large strides (e.g. with wide resolutions like 21824x16), the value of `3 * kiRefStride` exceeds 65535 and sets the 17th bit of `r6`. 

Inside the loop, `r6w` is used to load a 16-bit value from memory. However, the 16-bit move does not clear the upper bits of `r6`, leaving the stale high bits from the stride calculation intact. This leads to incorrect index calculations when `r6` is subsequently used.

This PR resolves the issue by using `movzx` to zero-extend the 16-bit load into `r6d`, ensuring the upper bits of the register are cleared.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,7 @@ endif
 
 ifeq ($(USE_ASAN), Yes)
 CFLAGS += -fsanitize=address
+CXXFLAGS += -fsanitize=address
 LDFLAGS += -fsanitize=address
 endif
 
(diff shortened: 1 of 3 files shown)


sprangerik:
A workaround had to be added to meson.build because otherwise compilation of the gtest library failed. The correct fix there is probably just to uprev the gtest dependency.

Other than that this now passes the tests on all platforms and is ready for review.

BenzhengZhang:
> A workaround had to be added to meson.build because otherwise compilation of the gtest library failed. The correct fix there is probably just to uprev the gtest dependency.
> 
> Other than that this now passes the tests on all platforms and is ready for review.

The gtest dependency has been upgraded on master to fix the compiler failure, so the -Wno-uninitialized-const-pointer workaround is no longer needed. Could you please merge the latest master into your branch and drop that workaround? Thanks! @sprangerik

PR #3961, opened 2026-07-06
Fix asymmetric chroma plane stride handling in encoder preprocessing

sprangerik:
When preprocessing incoming I420 source frames, WelsMoveMemoryWrapper
previously read only the U chroma plane stride (iStride[1]) and used it as a
shared stride for copying both the U and V chroma planes. If an incoming
frame has distinct, asymmetric strides for the U and V chroma planes (for
example, when the U plane stride is larger than the V plane stride), copying
the V plane using the U plane's stride advances the source pointer incorrectly.

This change ensures correct memory advancement for each chroma plane by:
- Modifying WelsMoveMemory_c to accept separate destination and source
  strides for the U and V chroma planes (iDstStrideU, iDstStrideV,
  iSrcStrideU, iSrcStrideV).
- Updating WelsMoveMemoryWrapper to read and pass iStride[1] and iStride[2]
  separately from the input SSourcePicture.
- Updating internal callers (DownsamplePadding and Padding) to pass their
  respective U and V line sizes cleanly.
- Adding a unit test (CustomChromaPlaneStrides) to verify that distinct,
  asymmetric chroma plane strides are handled correctly.

### DIFF
--- a/codec/encoder/core/src/encoder_ext.cpp
+++ b/codec/encoder/core/src/encoder_ext.cpp
@@ -3486,10 +3486,10 @@ int32_t WelsEncoderEncodeExt (sWelsEncCtx* pCtx, SFrameBSInfo* pFbi, const SSour
     pFbi->sLayerInfo[iNalIdx].iNalCount  = 0;
   }
   // perform csc/denoise/downsample/padding, generate spatial layers
-  iSpatialNum = pCtx->pVpp->BuildSpatialPicList (pCtx, pSrcPic);
-  if (iSpatialNum == -1) {
-    WelsLog (& (pCtx->sLogCtx), WELS_LOG_ERROR, "Failed in allocating memory in BuildSpatialPicList");
-    return ENC_RETURN_MEMALLOCERR;
+  int32_t iRet = pCtx->pVpp->BuildSpatialPicList (pCtx, pSrcPic, &iSpatialNum);
+  if (iRet != ENC_RETURN_SUCCESS) {
+    WelsLog (& (pCtx->sLogCtx), WELS_LOG_ERROR, "Failed in BuildSpatialPicList, error=%d", iRet);
+    return iRet;
   }
 
   if (pCtx->pFuncList->pfRc.pfWelsUpdateMaxBrWindowStatus) {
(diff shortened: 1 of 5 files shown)


BenzhengZhang (approved):

BenzhengZhang:
@sprangerik **Two suggestions:**

**1. Add chroma stride validation.** The input guard currently only validates the Y plane (`iSrcWidth > kiSrcStrideY`). Since this PR is specifically about hardening chroma stride handling, it would be natural to also reject inputs whose chroma stride is too small — otherwise a caller passing a chroma stride smaller than `width/2` would still `WelsMemcpy` past the end of each U/V row:

```cpp
if ((iSrcWidth >> 1) > kiSrcStrideU || (iSrcWidth >> 1) > kiSrcStrideV)
  return;
```

The new test happens to use `strideV = 32 == width/2`, which just fits, so this gap isn't currently exercised.

**2. Strengthen the test to verify correctness, not just memory safety.** `CustomChromaPlaneStrides` only asserts `rv == 0`, so it relies on ASAN to catch the OOB read. A stride-handling bug that stays in-bounds could still pass silently. Consider filling the U and V planes with distinguishable patterns and asserting the copied/encoded chroma content, so the test guards actual correctness in addition to memory safety.

codec/encoder/core/src/encoder_ext.cpp
```diff
@@ -3487,9 +3487,9 @@ int32_t WelsEncoderEncodeExt (sWelsEncCtx* pCtx, SFrameBSInfo* pFbi, const SSour
   }
   // perform csc/denoise/downsample/padding, generate spatial layers
   iSpatialNum = pCtx->pVpp->BuildSpatialPicList (pCtx, pSrcPic);
-  if (iSpatialNum == -1) {
-    WelsLog (& (pCtx->sLogCtx), WELS_LOG_ERROR, "Failed in allocating memory in BuildSpatialPicList");
-    return ENC_RETURN_MEMALLOCERR;
+  if (iSpatialNum < 0) {
+    WelsLog (& (pCtx->sLogCtx), WELS_LOG_ERROR, "Failed in BuildSpatialPicList, error=%d", -iSpatialNum);
+    return -iSpatialNum;
```
BenzhengZhang: `-iSpatialNum` only maps to the right enum for `-1` (→ `ENC_RETURN_MEMALLOCERR`),
and only because that code happens to be 1. Other negatives are wrong:
`-2 → 0x02 (UNSUPPORTED_PARA)`, `-3 → 3` (not even a valid flag). Since this path
only ever returns `-1` today anyway, the generalized `< 0` / `-iSpatialNum`
handling is unneeded now and unsafe the moment another negative is introduced.

Suggest returning real `-ENC_RETURN_*` codes from the sentinels, or returning the
count via an out-param so the return value stays a plain positive `ENC_RETURN_*`.

sprangerik: Good point. I've changing it back to return ENC_RETURN_MEMALLOCERR with an output param for iSpatialNum.

PR #3966, opened 2026-07-15
Fix slice buffer size validation and allocation during dynamic load balancing

sprangerik:
This change resolves a potential crash that can occur during multi-threaded encoding when dynamic load balancing is enabled on frames with extreme aspect ratios.

Under dynamic load balancing, slices can expand significantly beyond their initial estimated sizes. A mismatch between the tracked buffer size and the actual allocated slice buffer size led to incorrect bounds checks during NAL writing.

This is resolved by adding a dedicated `uiBsSize` field to `TagWelsSliceBs` to track the actual allocated size of the slice bitstream buffer (`pBs`) and using it in `WriteSliceBs` for boundary verification, while also increasing the default slice buffer allocation size when load balancing is enabled.


BenzhengZhang:
@sprangerik another conflict

BenzhengZhang (approved):

PR #3987, opened 2026-08-27
encoder: use memset instead of mmx for fixed-size buffer zero-ing

fippo:
The sizes are compile-time constants, so memset expands to two inline SSE stores instead of a call to WelsSetMemZeroSize8_mmx, which zeroes 8 bytes per loop iteration and ends in emms, serialising against the SSE code around it.

Bit-exact, no changes to the bitstream. Increases performance by 1.5-2%


fippo:
(tl;dr is: memset got smarter, now sees the constant and there is less pointer indirection which apparently helps with branch prediction)

PR #4004, opened 2026-09-04
decoder: fix bi-predicted 16x8/8x16 macroblocks in B slices

### LINKED ISSUE #3712: Decoding quality is very poor in some situation.
For example, decoding [original.mp4](https://github.com/cisco/openh264/assets/32232575/b6a457c9-6498-4aa2-b854-85c546389c04) by libopenh264 and encoding again by libx264 results in:

https://github.com/cisco/openh264/assets/32232575/8b2f36db-d92a-4df9-a927-210338b9e076

while decoding the same video by h264 decoder in ffmpeg and encoding again by libx264 results in:

https://github.com/cisco/openh264/assets/32232575/9c9de0d2-568e-4ae2-ac94-909d9e8af46a

Is any important feature of h264 missing from openh264?

tyan0:
# Fix bi-predicted 16x8/8x16 macroblocks in B slices

Fixes #3712.

## The bug

`GetInterBPred()` (`codec/decoder/core/src/rec_mb.cpp`) declares `pMCRefMem` outside
the partition loop, but advances it to the second partition from **inside** the
reference-list loop, and never resets it between partitions:

```c
for (int32_t i = 0; i < 2; ++i) {              // i = partition (0 = left/top, 1 = right/bottom)
  for (int32_t listIdx = LIST_0; listIdx < LIST_A; ++listIdx) {
    if (IS_DIR (iMBType, i, listIdx)) {
      ...
      if (i) { pMCRefMem.pDstY += 8; ... }     // once per LIST, never reset
      BaseMC (pCtx, &pMCRefMem, listIdx, ...); // LIST_1 overwrites LIST_0's result
```

For a bi-predicted partition this fails twice over:

1. **`i == 0`** — the `BaseMC()` for `LIST_1` overwrites the `LIST_0` prediction already
   sitting in `pMCRefMem`. The `listCount == 2` branch then redoes the `LIST_1` motion
   compensation into `pTempMCRefMem` and calls `BiPrediction()`, which therefore averages
   **LIST_1 with LIST_1**. The partition ends up as a plain LIST_1 prediction.

2. **`i == 1`** — the `if (i)` offset is applied once per list instead of once per
   partition, so the destination is advanced **twice** (+16 luma). The `LIST_1` motion
   compensation and the combine step write **outside the macroblock**, into the
   neighbour, and the partition keeps a plain LIST_0 prediction.

Both the `IS_INTER_16x8` and `IS_INTER_8x16` paths have this shape.

A single wrong macroblock is enough to ruin the whole picture, because Intra_16x16
DC and Vertical prediction copy from the left and top neighbours: the error travels
along the macroblock row via `I16_PRED_DC_L` and then straight down every column via
Intra_16x16 Vertical. Reference B pictures carry it into later frames.

## Worked example

Display frame 32 of the `original.mp4` attached to #3712 — MB(1,0), `mbtype = 0xf020`
(`MB_TYPE_8x16 | MB_TYPE_L0 | MB_TYPE_L1`, i.e. B_Bi_8x16), both partitions bi-predicted
with `refidx = 0/0`:

| partition | L0-only pred | L1-only pred | correct average | openh264 produced |
|---|---|---|---|---|
| left 8x16  | 191 | 174 | **183** | 174 (= L1 only) |
| right 8x16 | 191 | 174 | **183** | 191 (= L0 only) |

The residual is identical in openh264 and in the reference decoders, so the
discrepancy is purely in the prediction. Of the 1530 macroblocks in that frame only
**2** were genuinely mis-predicted; the other 754 broken macroblocks were intra
macroblocks reproducing the error through DC/Vertical prediction.

## The fix

Derive each partition's destination from a saved macroblock base, send the first list
of a partition to `sRef` and the second to `sTempRef` so the second `BaseMC()` cannot
clobber the first, and combine once after the list loop.

## Verification

The sample attached to #3712 (720x540, High profile, CABAC, B frames,
`weighted_bipred_idc = 2`), 60 frames, compared against two independent decoders:

| | frames differing |
|---|---|
| Microsoft Media Foundation vs Intel Quick Sync | **0 / 60** (references agree) |
| openh264 master | 30 / 60 (worst 34.9 dB) |
| **openh264 + this PR** | **2 / 60** (92.6 dB, 102.8 dB) |

The two remaining frames differ by ±1 on block edges only and are a separate
B-slice deblocking boundary-strength issue, unrelated to this change.

### In-tree test vectors

All 54 decodable vectors under `res/`, patched vs unpatched:

* **46 unchanged.**
* **`Cisco_Men_whisper_640x320_{CABAC,CAVLC}_Bframe_9.264` — fixed.** Before: 4 of 9
  frames differ from the reference decoders (worst 60.8 dB / 59.8 dB). After:
  **bit-exact with both**, which agree with each other. These two vectors are a
  self-contained, in-tree demonstration of the bug.
* **6 `VID_*_temporal_direct` vectors — output changes but they are still not
  conformant.** Both before and after this change they differ from the reference
  decoders on most frames, so they are affected by a **separate, larger bug in the
  temporal direct path** that this PR does not address. Worst-case PSNR against Intel
  Quick Sync moves as follows:

  | vector | before | after |
  |---|---|---|
  | VID_1280x544_cabac_temporal_direct | 36.21 dB | 62.89 dB |
  | VID_1280x544_cavlc_temporal_direct | 23.94 dB | 24.30 dB |
  | VID_1280x720_cabac_temporal_direct | 48.27 dB | 52.95 dB |
  | VID_1280x720_cavlc_temporal_direct | 39.02 dB | 39.07 dB |
  | VID_1920x1080_cabac_temporal_direct | 41.49 dB | 80.90 dB |
  | VID_1920x1080_cavlc_temporal_direct | 30.27 dB | 30.68 dB |

  No vector gets worse. I am happy to open a separate issue for the temporal direct
  problem.

## Expected-hash updates

The hashes currently in `test/api/decoder_test.cpp` match the **unpatched** output for
all 8 affected vectors, i.e. they were recorded from the current decoder rather than
from a conformant reference. They are updated here to the new output.

`test/api/thread_decoder_test.cpp` keeps a separate table for the same vectors. It is
**deliberately not touched here**, because its expectations cannot currently be made
meaningful: the multi-threaded decoder is non-deterministic. Six runs of the same binary
on `res/VID_1280x544_cabac_temporal_direct.264` with a fixed thread count produced six
different digests, while `threads=1` is perfectly reproducible. Since `SetUp()` also picks
`(rand() % 2) + 2` threads per run, no single expected hash can be stable for the affected
vectors. This is a pre-existing problem unrelated to this PR, and I will report it
separately.

`test/api/decode_api_test.cpp` compares two decode passes against each other rather than
against fixed hashes, so it needs no change.

## Notes

* Style follows `build/astyle.cfg`: 2-space indent, no tabs, no trailing whitespace,
  LF line endings, longest added line 115 columns.
* No new warnings with `-Wall`.
* This PR was created with assistance from Claude Opus 5.


BenzhengZhang:
@tyan0 Only affect B Frame decoding? Is this a independent fix? If so I will merge it.

tyan0:
Yes, B slices only. `GetInterBPred()` is reached only from the `else` of
`if (pCtx->eSliceType == P_SLICE)` in `WelsMbInterConstruction()` and `WelsMbInterPrediction()`,
so P and I slices never enter it. Inside it the change is confined to the `IS_INTER_16x8` and
`IS_INTER_8x16` paths, and within those to partitions that are actually bi-predicted — a 16x8 or
8x16 partition using a single list takes the same route as before, and so do the 16x16 and 8x8
paths.

Independent, yes: it merges cleanly into master as it stands, and it does not depend on anything
else I have open.

One thing worth knowing rather than finding out later: **#4006 contains this same change.** I
branched the threading work on top of it, so all 104 changed lines of `rec_mb.cpp` here are
present there as well. That does not get in the way of merging this first — I have checked that
master + this + #4006 and master + #4006 + this both merge cleanly, so #4006 will simply absorb
it.

Issue #4008, opened 2026-09-08
Crashes on macOS at WelsDec::DecodeFrameConstruction() as OpenH264 plugin is shutting down


steven-michaud:
I have a fix for these crashes, and am about to open a pull request. But first I think it'd be useful to open a detailed bug report.

These crashes happen [here](https://github.com/cisco/openh264/blob/652bdb7719f30b52b08e506645a7322ff1b2cc6f/codec/decoder/core/src/decoder_core.cpp#L210), on the first reference in `WelsDec::DecodeFrameConstruction()` to its second parameter: `uint8_t** ppDst`. The "crash reason" is `KERN_INVALID_ADDRESS`: the (virtual) address in `ppDst` is no longer mapped into physical memory. These crashes happen on a secondary thread, created by the OpenH264 plugin when it's running in multi-threaded mode.

The object `ppDst` points to is [here](https://github.com/cisco/openh264/blob/652bdb7719f30b52b08e506645a7322ff1b2cc6f/module/gmp-openh264.cpp#L1032). By the time `WelsDec::DecodeFrameConstruction()` is called on its secondary thread, it has gone out of scope.

These crashes happen only on macOS, and then only with non-default settings. In Firefox's `about:config`, the following two settings need to be changed:

`media.rdd-applemedia.enabled`, from `true` to `false`
`media.gmp.decoder.multithreaded`, from `false` to `true`

Then the crashes can be triggered by playing any H.264 video, for example [this BBB variant](https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/1080/Big_Buck_Bunny_1080_10s_1MB.mp4) or [this trailer for the 2007 Simpsons movie](https://download4.dvdloc8.com/trailers/divxdigest/simpsons_movie_1080p_hddvd_trailer.zip). They happen as the video ends, and the plugin shuts down.

steven-michaud:
[bugzilla2065570-macOS-crash.txt](https://github.com/user-attachments/files/31966930/bugzilla2065570-macOS-crash.txt)

Here's an Apple crash report for one of these crashes. It's preferable to a Mozilla crash report because those don't symbolicate any of the addresses in the OpenH264 plugin. You can make Firefox generate an Apple crash report for any crash by running it from Terminal as follows. You won't see a crash report dialog for this bug's crashes. You'll need to look for the crash report in Apple's Console app.

```
MOZ_CRASHREPORTER_DISABLE=1 open /Applications/Firefox\ Nightly.app
```

[...]

PR #4009, opened 2026-09-08
Fix crashes at WelsDec::DecodeFrameConstruction() as reported in Issue #4008

steven-michaud:
This fixes the crashes I reported in Issue #4008.

I had no problems with it in any of my tests, on macOS, Linux or Windows (Windows 11). In all the decoding sessions I observed, there is only ever a single `OpenH264VideoDecoder` object, and it lives as long as the session itself (after which the Gecko Media Plugins process terminates). And all calls to `OpenH264VideoDecoder::Decode_w()`  happen on the `GMPThread`.

### DIFF
--- a/module/gmp-openh264.cpp
+++ b/module/gmp-openh264.cpp
@@ -1040,11 +1040,11 @@ class OpenH264VideoDecoder : public GMPVideoDecoder, public RefCounted {
     if (gmp_api_version_ >= kGMPVersion34) {
       decoded.uiInBsTimeStamp = inputFrame->TimeStamp();
     }
-    unsigned char* data[3] = {nullptr, nullptr, nullptr};
+    memset(data_, 0, sizeof(data_));
 
     dState = decoder_->DecodeFrameNoDelay (inputFrame->Buffer(),
                                      inputFrame->Size(),
-                                     data,
+                                     data_,
                                      &decoded);
 
     if (dState) {
@@ -1058,7 +1058,7 @@ class OpenH264VideoDecoder : public GMPVideoDecoder, public RefCounted {
                                  &OpenH264VideoDecoder::Decode_m,
                                  inputFrame,
                                  &decoded,
-                                 data,
+                                 data_,
                                  renderTimeMs,
                                  valid));
   }
@@ -1177,6 +1177,8 @@ class OpenH264VideoDecoder : public GMPVideoDecoder, public RefCounted {
   FrameStats stats_;
   uint32_t gmp_api_version_;
   bool shutting_down;
+  // Make data_, used in Decode_w(), live as long as the class object.
+  unsigned char* data_[3];
 };
 
 extern "C" {


steven-michaud:
My patch failed the `ThreadDecoderPreviousPicRaceTest.ThreadedOutputIsConsistent` test on your servers. But I suspect this failure is spurious. I ran the tests locally (`meson test -C builddir -v`) on macOS 15.7.9 (on a 2019 Intel MacPro, `MacPro7,1`) and Pop!_OS 24.04 LTS Linux (on a System76 Meerkat), and had no failures, including with this test. For good measure I also ran the BinaryCompare tests, using `./test/encoder_binary_comparison/run_Main.sh LocalTest`. No failures there, either. Both machines have 16 cores. The MacPro has 48GB of RAM. The Meerkat has 64GB.

What's my next step? Please let me know. Should I test on less powerful machines, possibly running in VMs?

steven-michaud:
I finally figured out how to run the failed test individually, and on an infinite loop:

```
builddir/test/api/test_api --gtest_filter=ThreadDecoderPreviousPicRaceTest.ThreadedOutputIsConsistent --gtest_repeat=-1
```

Now I'm seeing the reported failures with my patch, once for every 500 or so passes. I don't see them in a build without my patch.

I'll be looking for a fix.

steven-michaud:
Oops, spoke too soon. My first tests were on macOS. On Linux I see the failures more often, and *both with and without my patch*. These failures are definitely spurious.

steven-michaud:
@BenzhengZhang, you should be aware of this, since you wrote the test.

steven-michaud:
Actually, the test results *aren't* spurious. As best I can tell they show a real problem. But that problem has nothing to do with my patch.

steven-michaud:
The continuing failures in the `ThreadDecoderPreviousPicRaceTest.ThreadedOutputIsConsistent` test are also discussed [here](https://github.com/cisco/openh264/pull/4006).

BenzhengZhang:
@steven-michaud Intermittent issue; the cause has been identified.

steven-michaud:
I tested with the current `master` branch, and found the `ThreadDecoderPreviousPicRaceTest.ThreadedOutputIsConsistent` test no longer fails, on either macOS or Linux, even when run hundreds of times. But now my test video (the Simpsons movie trailer) no longer works in multi-threaded mode on macOS, Windows or Linux. On Linux it even crashes.

I'll open a new bug on this, once I have a better idea of what's going on.

Edit: Backing out my patch doesn't fix the problem.

BenzhengZhang:
> I tested with the current `master` branch, and found the `ThreadDecoderPreviousPicRaceTest.ThreadedOutputIsConsistent` test no longer fails, on either macOS or Linux, even when run hundreds of times. But now my test video (the Simpsons movie trailer) no longer works in multi-threaded mode on macOS, Windows or Linux. On Linux it even crashes.
> 
> I'll open a new bug on this, once I have a better idea of what's going on.
> 
> Edit: Backing out my patch doesn't fix the problem.

@steven-michaud What about version 2.6.0? Master already has a number of multi-thread fixes, so it's worth confirming whether this regression is still present there. I'd like to reproduce it locally — could you share the test file and the steps/process you used, so we can dig into the root cause together?

steven-michaud:
I've opened Issue #4014.

steven-michaud:
@BenzhengZhang, when do you think this will get into a release version?

BenzhengZhang:
A release is expected as early as October.

PR #4010, opened 2026-09-09
decoder: clear stale hand-off signals when the new-sequence chain is cut

tyan0:
One-commit follow-up to the new-sequence handling in `ParseAccessUnit()`: the barrier
introduced there clears the state that would point into DPB storage about to be replaced, and
this adds the one remaining piece of per-worker state that the same cut leaves behind.

## What is missing

`sSliceDecodeStart` is a manual-reset event, one per worker, and the worker that *consumes* it
is the one that resets it — `ConstructAccessUnit()` waits on the previous worker's event and
calls `RESET_EVENT()` on it afterwards.

When the new-sequence path cuts the hand-off chain, that consumer never runs, so a signal
raised just before the boundary stays raised. The next frame that waits on that worker passes
the wait immediately — before the frame it is about to mark has been decoded — and carries out
the reference marking against whatever picture and marking state that context still holds.

The fix resets the events in the same place the rest of that state is cleared. Past the
barrier nothing is decoding and the chain is cut for every context, so any hand-off signal
still raised has lost its consumer.

## How it was found

`res/BA_MW_D.264` carries an IDR every 30 frames. Every access unit at which threaded output
first diverged from single-threaded output was congruent to an IDR's modulo the thread count —
the effect follows the worker that decoded the IDR rather than a fixed frame offset, which is
what pointed at per-worker state rather than at the picture data. Instrumenting the marking
then shows the consumer arriving before the producer has recorded anything for the frame, and
the following frame's reference list missing a picture:

    threads=1   frame_num=3   L0 = [2 1 0]
    threads=3   frame_num=3   L0 = [1 0]

`ThreadDecoderPreviousPicRaceTest` is what made this findable at all: it fails intermittently
on the parent commit, and it is deterministic afterwards.

## Measurements

Cygwin/x86_64, gcc 14, 8 cores. `ThreadDecoderPreviousPicRaceTest`, 60 attempts:

| | passed | failed |
|---|---|---|
| parent commit | 58 | 2 |
| with this change | 60 | 0 |

The whole `ThreadDecoder*` set passes. `res/BA_MW_D.264`, `res/MR2_MW_A.264`,
`res/MR2_TANDBERG_E.264` and the 1280x544 and 1280x720 `temporal_direct` vectors each give a
single digest over fifteen in-process passes at two and at three threads, equal to their
`threads=1` digest.

Two failures in sixty is a thin margin to measure a change against, so the same change was
measured on a build carrying diagnostic instrumentation, which widens the window and raises
the rate on `res/BA_MW_D.264` at three threads from roughly one pass in seventy to better than
one in two. Decoding it repeatedly in one process — the effect only shows up that way, one
decode per process looks clean — on the same build with the resets compiled in or out:

| | without | with |
|---|---|---|
| output differs from `threads=1` | 54 of 99 | 0 of 100 |
| reference list differs from `threads=1` | 54 | 0 |
| `ThreadDecoderPreviousPicRaceTest` | 50 of 60 fail | 0 of 60 fail |

## One caveat

I could not observe the stale signal being consumed directly. The proportion of hand-off waits
that find the event already raised is around 26% even in runs whose output is correct, so that
count cannot separate a legitimate early completion from a leftover. The mechanism above
matches the instrumented log, but the causal claim rests on the effect of the change rather
than on a direct observation of the stale wait. If someone sees a cheaper way to pin that
down, I would be glad to add it.

## Relationship to #4006

#4006 is based on an older master and does not contain this code path, so the two are
independent: this applies to master with no fuzz and does not apply to #4006's head at all.
Merging them in either order is clean, and the combined tree passes
`ThreadDecoderPreviousPicRaceTest` 60 of 60 and the whole `ThreadDecoder*` set. The
intermittent failures of that test seen in #4006's CI come from this path, not from #4006 —
CI tests the pull request merged with master.

### DIFF
--- a/codec/decoder/plus/src/welsDecoderExt.cpp
+++ b/codec/decoder/plus/src/welsDecoderExt.cpp
@@ -1386,6 +1386,14 @@ DECODING_STATE CWelsDecoder::ParseAccessUnit (SWelsDecoderThreadCTX& sThreadCtx)
           RELEASE_SEMAPHORE (&m_pDecThrCtxActive[i]->sThreadInfo.sIsIdle);
         }
       }
+      for (int32_t i = 0; i < m_iCtxCount; ++i) {
+        //Past the barrier nothing is decoding and the chain is cut for every context, so a
+        //hand-off signal still raised has lost the consumer that would have reset it. Left
+        //raised, the next frame that waits on that worker passes the wait at once and marks
+        //a picture that has not been decoded.
+        if (m_pDecThrCtx[i].pCtx != NULL)
+          RESET_EVENT (&m_pDecThrCtx[i].sSliceDecodeStart);
+      }
       sThreadCtx.pCtx->pLastThreadCtx = NULL;
     }
     iErr = AllocPicBuffOnNewSeqBegin (sThreadCtx.pCtx);


tyan0:
The `linux-cross-loongarch64` failure is not from this change — it is a pre-existing defect in
the unit test.

`WelsIDctFourT4Rec_lasx` is an encoder-side kernel; this pull request is one hunk in
`codec/decoder/plus/src/welsDecoderExt.cpp`. `DecodeMbAuxTest` also finishes before any
`ThreadDecoder*` test starts (link order: decoder, **encoder**, processing, api), and this pull
request adds no test, so it does not move the shared `rand()` sequence the test draws its inputs
from. Run #644 on `0bab7fe8`, this pull request's base, passed all nine jobs, and the same tree
passes the same job on my fork.

The test itself is the problem. `TestIDctFourT4Rec` compares a 16-bit SIMD kernel against
`WelsIDctT4Anchor<int16_t>`, which adds the rounding term inside the 16-bit width it emulates:

```c
p_dst[i] = WelsClip1 (uiDst + (clip_t (tmp[i] + ... + 32) >> 6));
```

A rounding-shift instruction — `xvsrari.h` here — adds it in a wider intermediate and does not
wrap, so the two differ exactly when `S + 32` leaves the int16 range, giving 0 instead of 255.
The test draws coefficients from the whole int16 range, so it walks into that band on its own.
Modelling both forms with the same input generation, over 500000 sets:

| | |
|---|---|
| sets with at least one mismatching pixel | 15456 (**3.09%**) |
| mismatching pixels | 15659 |
| times `S + 32` left the int16 range | 15659 |

The two counts are equal and the per-pixel rate matches 32/65536, so the mechanism is not in
doubt. This is already known in the same file — the SSE2 tests limit the range for it:

```c
iRefDct[i][j] = iDct[i * 4 + j] = (rand() & ((1 << 15) - 1)) - (1 <<
                                  14); //2^14 limit, (2^15+32) will cause overflow for SSE2.
```

`TestIDctFourT4Rec`, used by `_lasx` and `_mmi`, does not. I will report that separately (#4011) since
it is independent of this pull request.

One caveat: the failure is input-dependent, not timing-dependent, so a re-run in an identical
environment should reproduce it. What moved the input position between two runs of the same tree
I have not established — the compiler and runner image versions are identical, and `qemu-user-static`
is installed unpinned with `-qq`, which does not print its version.

Nothing to change here. Recording it so the red CI is not read as a defect in this change.

BenzhengZhang:
---

**1. Why is only `sSliceDecodeStart` reset, not `sSliceDecodeFinish`?** *(most important — I'd treat this as blocking)*

On the new loop:
```cpp
if (m_pDecThrCtx[i].pCtx != NULL)
  RESET_EVENT (&m_pDecThrCtx[i].sSliceDecodeStart);
```
`sSliceDecodeStart` and `sSliceDecodeFinish` are adjacent manual-reset events in the same struct (`decoder_context.h`), created/closed as a pair, and reset together on the normal consumption path elsewhere in this file (`RESET_EVENT(sSliceDecodeStart)` immediately followed by `RESET_EVENT(sSliceDecodeFinish)`). If `sSliceDecodeFinish` is subject to the same "chain cut leaves it stale" hazard, this fix only closes half the hole — a stale `sSliceDecodeFinish` could cause a symmetric bug where some later wait on "this frame finished decoding" is satisfied prematurely. Can you confirm whether `sSliceDecodeFinish` was analyzed and ruled out, or whether it should be reset here too?

**2. Does `sImageReady` have the same exposure?**

Third event in the same struct, also created/closed as part of the same set. Lower confidence than #1, but worth asking: what's the consumption semantics for `sImageReady`, and could it go stale the same way when the hand-off chain is cut?

**3. The "past the barrier nothing is decoding for every context" invariant needs to be substantiated**

The new loop iterates all `m_iCtxCount` slots, but the preceding drain loop only iterates `m_pDecThrCtxActive` (the subset currently tracked as active, sized `m_DecCtxActiveCount`). These two sets aren't necessarily equal — e.g. a slot that was never dispatched yet, or one that was recently removed from the active list but whose worker thread hasn't fully reached idle at this exact instant. Can you confirm there's an earlier guarantee that *every* `m_iCtxCount` context — not just the previously-active ones — is genuinely quiescent by this point? If so, it'd be worth a comment or debug assert making that invariant explicit so a future refactor doesn't silently break it.

**4. Validation is fairly weak for a ~1% baseline flake rate**

The PR description reports "combined tree passes `ThreadDecoderPreviousPicRaceTest` 60 of 60." Given the original failure rate was reportedly around 1%, 60 clean runs isn't strong statistical evidence on its own (even an unfixed build has a non-trivial chance of passing 60/60 by luck). Suggestions:
- Bump `--gtest_repeat` to several hundred/thousand and report the before/after failure rate as evidence in the PR description.
- Ideally add a targeted deterministic unit test that directly constructs the "new-sequence cuts the chain while `sSliceDecodeStart` is signaled" scenario and asserts the event is cleared — rather than relying solely on a pre-existing, inherently probabilistic integration test.

**5. Consider adding instrumentation, since the root cause is inferred rather than directly observed**

You mention "I have not identified the writer, and I would rather say so than guess." Given that, it'd help to add a debug-only log/counter in this reset loop that records whenever it actually finds an event in the signaled state before resetting it. That would give direct forensic evidence if this class of bug resurfaces, instead of relying on hash-mismatch failures alone.

**6. Please state the merge-order dependency with #4006 explicitly in the description**

`#4006`'s branch predates this fix, so its CI intermittently hits this same stale-event bug through no fault of its own. Rather than combining the two PRs, please note in the description that `#4010` should land first (or that `#4006` should be rebased on top of it) — that way reviewers don't mistake #4006's flakiness for a regression it introduced.

BenzhengZhang:
well done!  @tyan0 
Your fix same as copilot suggestion https://github.com/cisco/openh264/pull/4013 , so I would merge your PR as first priority.

PR #4016, opened 2026-09-18
Fix regression when encoding very large frames

aosmond:
WelsMoveMemoryWrapper rejects any source picture whose area exceeds MAX_MBS_PER_FRAME. That limit bounds a coded frame, which ParamValidationExt already enforces on the spatial layers, but this copy runs before downsampling, so the input picture may legitimately be larger. WelsInitScaledPic allocates pScaledInputPicture for SUsedPicRect, so the guard rejects a copy the allocator sized to fit.

The result is that EncodeFrame returns cmUnsupportedData with videoFrameTypeSkip for every frame, producing no bitstream, for any caller encoding above 36864 macroblocks (~4096x2304) into a smaller layer. Introduced by the validation guards added in: https://github.com/cisco/openh264/pull/3961

Bound the copy by the destination's own iWidthInPixel/iHeightInPixel instead.

This was caught when attempting to update OpenH264 for Firefox. There is a web-platform test which encodes large frames: https://searchfox.org/firefox-main/rev/572cbc53633d3ed8a0b1390828611752f0317938/testing/web-platform/tests/webcodecs/video-encoder-h264.https.any.js#29


aosmond:
@sprangerik, @BenzhengZhang
Let me know if this does not maintain the spirit of what the original PR was going for

BenzhengZhang (approved):

BenzhengZhang:
@aosmond Is there any other GMP API update?  If not, I would start make a new release version nowadays.

Issue #4014, opened 2026-09-15
PR #4002 breaks OpenH264 plugin's H.264 video decoding


steven-michaud:
As of PR #4002 landing on the master branch, the OpenH264 plugin doesn't decode video correctly when run in multi-threaded mode. After 10-15 seconds, video output gets stuck at a single frame. Audio output continues as normal. Sometimes, on Linux, the plugin also crashes (which of course stops both video and audio).

I've been testing on macOS and Linux with today's Firefox nightly, with [this trailer for the 2007 Simpsons movie](https://download4.dvdloc8.com/trailers/divxdigest/simpsons_movie_1080p_hddvd_trailer.zip).

Firefox requires settings changes to make it use the OpenH264 plugin in multi-threaded mode. These need to be done in `about:config`.

**MacOS**

`media.rdd-applemedia.enabled`, from `true` to `false`
`media.gmp.decoder.multithreaded`, from `false` to `true`

**Linux**

`media.rdd-ffmpeg.enabled`, from `true` to `false`
`media.gmp.decoder.multithreaded`, from `false` to `true`

You can make Firefox use a newly compiled plugin by copying it over the "original" plugin in its profile's `gmp-gmpopenh264/2.6.0` directory.

PR #4002 is the first commit with which this happens. When I reset the master branch to just before this commit, things return to normal. Though this also reopens Issue #4008, since my patch for it also gets reverted.

```
git reset 9d974972d9f1ef572b87d8d774af8f8c304ad3e1
git restore .
```

steven-michaud:
When testing on Linux, you need to use [today's nightly](https://archive.mozilla.org/pub/firefox/nightly/2026/09/2026-09-15-09-16-05-mozilla-central/) or a more recent one. Today's nightly is the first to contain my fix for https://bugzilla.mozilla.org/show_bug.cgi?id=2071378. Without this you can't make Firefox use the OpenH264 plugin in multi-threaded mode on Linux.

steven-michaud:
I've now gotten exactly the same results on Windows 11. Video output gets stuck unless I revert the master branch to just before PR #4002 landed.

Firefox requires the following settings changes on Windows to use the OpenH264 plugin in multi-threaded mode.

`media.rdd-wmf.enabled`, from `true` to `false`
`media.gmp.decoder.multithreaded`, from `false` to `true`

steven-michaud:
[...]

PR #4019, opened 2026-09-20
Re-land #4002 and #4010 with a fix for the stall they caused

tyan0:
Requested in #4014: re-land #4002 and #4010 together with the fix for the stall that took them
out.

## The two commits

**`Reapply "Revert PR4002 && PR4010 (#4015)"`** restores exactly what #4015 removed, including
the threaded regression tests that went with it. Against the commit immediately before the
revert, every decoder file is byte-identical; the only differences are the three files #4016
and #4017 have touched since.

**`decoder: release references dropped by a reference-list copy`** is the fix. It is a separate
commit deliberately: if the verification below does not come out clean, the re-land can be
backed out on its own without losing the analysis.

## Why #4002 stalled

Threaded decoding moves reference-picture state between worker contexts by assigning the whole
`SRefPic` in `DecodeCurrentAccessUnit()`:

    pLastThreadCtx->pCtx->sTmpRefPic = pLastThreadCtx->pCtx->sRefPic;
    WelsMarkAsRef (pLastThreadCtx->pCtx, pLastThreadCtx->pDec);
    pCtx->sRefPic = pLastThreadCtx->pCtx->sTmpRefPic;

The destination lists are replaced wholesale. A picture the destination was holding that the
source does not hold leaves every reference list at that instant with `bUsedAsRef` still set,
and since `SetUnRef()` was never called on it there is no deferred unref armed either, so
nothing can release it. Its buffer never returns to the pool, `PrefetchPic()` eventually fails,
and the worker returns before signalling the hand-off the next frame is waiting on — which is
what #4014 saw as video frozen on one frame while audio continued.

The assignment is older than #4002. What #4002 changed is that the two contexts' lists began to
diverge in a way that makes it bite. Full measurements are in
https://github.com/cisco/openh264/issues/4014#issuecomment-5746054667.

## Measurements

Cygwin/x86_64, gcc 14, 8 cores, on this branch.

A 1280x720 High profile stream (`pic_order_cnt_type = 2`, `log2_max_frame_num_minus4 = 0` so
MaxFrameNum is 16, `num_ref_frames = 6`, P slices only), decoded through a harness that calls
`DecodeFrameNoDelay()` and takes each frame immediately — this is what stalled at about six
seconds on three threads and sixteen on two before the fix:

| excerpt | threads | result |
|---|---|---|
| 480 frames | 1, 2, 3 | completes, no `PrefetchPic` error |
| 1440 frames | 1, 2, 3 | completes, no `PrefetchPic` error |

| test | |
|---|---|
| `codec_unittest` | 629 of 629 |
| `ThreadDecoderPreviousPicRaceTest.ThreadedOutputIsConsistent`, 60 runs | 60 pass, 0 fail |
| `ThreadDecoder*`, 20 runs | 20 pass, 0 fail |

The second row is the one #4010 is for: that race only exists once #4002 is in, so it is the
check that matters for bringing both back together.

## What this does not fix

**Threaded output still does not match single-threaded output.** On the stream above the three
thread counts give three different digests — the same values master produces with this patch
applied on its own, so the re-land does not make it worse, but it does not make it better
either. That divergence is what #4006 is for. This pull request only stops the decode from
stalling.

**`PrefetchPic()` failing still hangs rather than returning an error.** Whatever exhausts the
pool, `DecodeCurrentAccessUnit()` returns `ERR_INFO_REF_COUNT_OVERFLOW` and the hand-off that
worker would have signalled never happens, so the caller waits forever with no error it could
act on. This branch removes the cause that #4014 hit; it does not make that path safe in
general. Worth its own change.

## Verifying

@steven-michaud — the Firefox reproduction from #4014 is the one that should decide this. The
test you used repeats cleanly here:

    ./codec_unittest --gtest_filter=ThreadDecoderPreviousPicRaceTest.ThreadedOutputIsConsistent --gtest_repeat=-1

but a harness is not a player, and yours is the reproduction that found the problem.


steven-michaud:
I tested with [this patch](https://github.com/cisco/openh264/pull/4019.patch) applied to the `master` branch, and had no problems. I used [the 2007 Simpsons movie trailer](https://download4.dvdloc8.com/trailers/divxdigest/simpsons_movie_1080p_hddvd_trailer.zip), and tested on macOS, Linux and Windows. I also reran the `ThreadDecoderPreviousPicRaceTest.ThreadedOutputIsConsistent` test on an infinite loop, for hundreds of iterations, on macOS and Linux. No failures.

BenzhengZhang:
Thanks @steven-michaud @tyan0

tyan0:
Please hold off merging this — it is not ready.

The asan job on my fork caught a frame loss caused by my own commit:

    ThreadDecoderReorderQueueRaceTest.BufferedPictureQueueDrainsAllFrames
    Expected equality of these values:
      iDecodedFrames_
        Which is: 49
      50

`WelsReleaseDroppedRefs()` calls `SetUnRef()`, which decides a picture is unheld from
`iRefCount <= 0 && iPinCount <= 0`. In threaded mode a picture waiting in the output queue
does not satisfy "held" by that test: `BufferingReadyPicture()` takes a reference only when
the decoder is single-threaded, while both dequeue paths release one unconditionally. So my
release can reset a picture the queue still points into — `m_sPictInfoList` holds plane
pointers, not copies — and `PrefetchPic()` then hands that buffer to another frame.

The obvious repair, pairing the increment with the decrements, is not right either: it makes
`ThreadDecoderInitTest.ThreadedSameResolutionRefQueueResizeNoUseAfterFree` hang, so the DPB
resize paths do not expect queued pictures to hold references. I have backed that out.

Neither of these shows up on x86 without asan — I had 100 of 100 clean runs locally on the
build that loses a frame — so I am treating my own local results as unable to establish much
here, and will be verifying on CI rather than reasoning about it.

The analysis of *why* #4002 stalls, in #4014, is unaffected by any of this: that part was
measured and does not depend on the fix being right. It is the release itself that needs a
safe home.

Still working on it. I will report either a fix that CI accepts, or the reason I cannot find
one, rather than push another attempt on top.

@steven-michaud — nothing needed from you for now; I will say when there is something worth
retesting.

tyan0:
Correction to my previous comment: **the frame loss is not caused by my commit.** I said it
was, and that was wrong.

I instrumented the three events by picture-buffer index — `BufferingReadyPicture()` publishing
a buffer, either dequeue path releasing it, and `WelsReleaseDroppedRefs()` calling
`SetUnRef()` — and ran the failing test twenty times. On its stream
(`res/Adobe_PDF_sample_a_1024x768_50Frms.264`):

    enqueues 50, dequeues 50, releases 0

`WelsReleaseDroppedRefs()` never fires there at all, so it cannot be what drops a frame. The
counts confirm the logging itself was reaching, which is what I had not checked before
reasoning from it.

I also could not reproduce the failure anywhere. Three branches on my fork — plain master
`6e99da76`, master plus the re-land `6e06f235`, and this pull request's head — each running
`./codec_unittest --gtest_filter='ThreadDecoder*' --gtest_repeat=40` and two full
`./codec_unittest` runs under asan. All clean, with no difference between the three. My first
attempt at this used a shell loop over a filtered run, which was useless: the thread count in
`BaseThreadDecoderTest::SetUp()` comes from `rand()`, so filtering and restarting the process
tested one and the same configuration forty times.

So the 49-of-50 remains a single unexplained sample, and there is no evidence connecting it to
this pull request.

What still stands from the previous comment: the `iRefCount` pairing I described as "the
obvious repair" is wrong on its own terms — it hangs
`ThreadDecoderInitTest.ThreadedSameResolutionRefQueueResizeNoUseAfterFree`, so the DPB resize
paths do not expect queued pictures to hold references. It is backed out and there is nothing
to re-land there. I had built it on the diagnosis I am now retracting.

`decoder: do not touch reference-list entries the buffer no longer owns` is unaffected by any
of this: CI found a real use-after-free on `linux-cross-arm`, and that commit fixes it.

As far as I can establish, this is ready. One asan sample on my fork that I could not reproduce
on plain master either — your call on whether that is enough to proceed.

BenzhengZhang:
@tyan0 Is this the final version?

tyan0:
> @tyan0 Is this the final version?

Yes. I’m sorry for having caused confusion.

